PR #1, opened 2022-11-28
Adds the first architecture decision record

rossjones:
The first architecture decision record, and its index page, are added in this commit.  As always the first is a record of our decision to use decision records.

### DIFF
--- a/architecture/decision/README.md
+++ b/architecture/decision/README.md
@@ -1,0 +1,20 @@
+# Data Plaftorm - Architectural Decisions
+
+This repository is used to record architectural decisions made by the data-platform service.
+
+To understand why we are recording decisions and how we are doing it, please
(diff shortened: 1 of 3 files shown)


architecture/decision/README.md
```diff
@@ -0,0 +1,20 @@
+# Data Plaftorm - Architectural Decisions
+
+This repository is used to record architectural decisions made by the data-platform service.
+
+To understand why we are recording decisions and how we are doing it, please
+see [ADR-000](decisions/000-record-architecture-decisions.md).
+
+## Table of contents
+
+* ✅ [1. Record architecture decisions](records/000-record-architecture-decisions.md)
```
bagg3rs: This should be 0 or 000

bagg3rs (approved):

PR #5, opened 2022-12-02
Add a new ADR describing our proposed hosting environment

rossjones:
A proposal suggesting that the Modernisation Platform is likely the best place for the Data Platform to live, rather than within its own collection of AWS accounts.

### DIFF
--- a/architecture/decision/records/001-use-modernisation-platform.md
+++ b/architecture/decision/records/001-use-modernisation-platform.md
@@ -32,6 +32,14 @@ the data platform itself falls outside the type of system that the [Cloud Platfo
 
 As a result of this decision we will benefit from all of the features documented in the [Modernisation User Guide](https://user-guide.modernisation-platform.service.justice.gov.uk/user-guide/our-offer-to-you.html).
 
-In addition to the security baseline,
-and the automatic environment provisioning,
-we will likely make use of many other features that we would otherwise have to build ourselves when we could be enhancing the data platform itself.
+In addition to the drawing on wider Platforms & Architecture expertise,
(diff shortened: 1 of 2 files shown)


architecture/decision/records/001-use-modernisation-platform.md
```diff
@@ -0,0 +1,37 @@
...
+We have had 5 years experience of self-managing a collection of AWS accounts on the Analytical Platform,
+and it has not been a positive experience.
+
+While the Cloud Platform is a great environment for us to encourage our customers to host their services,
+the data platform itself falls outside the type of system that the [Cloud Platform is best at hosting](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/concepts/what-is-the-cloud-platform.html).
+
+## Consequences
+
+As a result of this decision we will benefit from all of the features documented in the [Modernisation User Guide](https://user-guide.modernisation-platform.service.justice.gov.uk/user-guide/our-offer-to-you.html).
+
```
bagg3rs, line 34: Should we add a consequence around being bound by MP account/env policies e.g. limited administrative rights (although a good thing and can be altered) and Dev environment being nuked weekly (again good thing and _can_ be altered)

bagg3rs (approved):
Looks good see my comment

architecture/decision/records/001-use-modernisation-platform.md
jemnery: @bagg3rs I think it's only sandbox / experimental envs which get nuked. It's fine to have a stable dev environment to allow a dev -> preprod -> prod workflow

PR #8, opened 2022-12-05
Adds a sample service along with a sample workflow

rossjones:
Adds a sample Python service which does very little, but does show how we can configure github actions for a specific service, without having to run them for all services.

In addition to showing the sample workflow for a specific service, it also includes the steps necessary to generate an SBOM for a service, in our bid to secure our supply chain. An example sbom for our example service is attached as a build artifact and can be found in actions (e.g. https://github.com/ministryofjustice/data-platform/actions/runs/3697555453 ).  During a release, this will be attached to the release.

CodeQL is also enabled as part of the github actions workflow to ensure common errors are caught early

This should be documented under: How do I add a new service to the monorepo?


jemnery:
Nice. Probably enough of a stub example for now until we have a think about architecture (e.g. affecting `services/` folder structure & documentation). We'll have core / infrastructure related services deployed to MP via ECR, but will we also have user-facing services deployed to CP from this repo? 

We should probably require a README for each service as they'll effectively all have their own mini runbook. 

A README in the root services folder would be good also - what languages do we accept? What are the minimum requirements for a service (docs / tests ...)?

rossjones:
Mostly I just wanted to test out whether it was feasible to have different sets of CI based on where the changes were committed, but I think I agree about documentation.  I'll add skeleton docs until we've agreed details around tech stack etc, and we can track that work in a github issue.

.github/workflows/svc_example_test.yml
```diff
@@ -0,0 +1,73 @@
...
+        python-version: ["3.9", "3.10"]
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: python -mpip install -r requirements.txt
```
tom-webber:
```suggestion
        run: python -m pip install -r requirements.txt
```
:pick: (Nitpick) I'm surprised that this doesn't cause issues, but the action appears to run fine!

julialawrence (approved):

PR #13, opened 2023-01-25
Moving github terraform to the new mono-repo and adding a format-code workflow.

julialawrence:
Moving Github terraform to new monorepo.


jacobwoffenden (approved):
LGTM

PR #35, opened 2023-02-01
Workflow file for github terraform.

julialawrence:
Adding workflow file for `terraform/github` which plans and applies the terraform using OIDC.


jacobwoffenden:
@julialawrence do we want to use the fancy Terraform PR comment from https://github.com/hashicorp/setup-terraform?

julialawrence:
> @julialawrence do we want to use the fancy Terraform PR comment from https://github.com/hashicorp/setup-terraform?

I'm having issues getting this working. Propose merging as-is and I'll return to it as I'm looking at a similar issue in another repo.

rossjones (approved):
Some missing trailing newlines, but otherwise LGTM

PR #40, opened 2023-02-02
 testing token permissions

Emterry:
fixing token permission issues from code scanner

### DIFF
--- a/.github/workflows/add-issues-to-project.yml
+++ b/.github/workflows/add-issues-to-project.yml
@@ -1,10 +1,14 @@
 name: Add issues to project
 on:
+  push:
+    branches:
+      - token-permission-fixes
   issues:
     types:
       - opened
 
-permissions: read-all
+permissions:
+  repository-projects: write
 
 jobs:
   add-to-project:
@@ -14,7 +18,7 @@ jobs:
       - uses: actions/add-to-project@960fbad431afda394cfcf8743445e741acd19e85 # v0.4.0
         with:
           project-url: https://github.com/orgs/ministryofjustice/projects/27
-          github-token: ${{ secrets.ORG_PROJECT_TOKEN }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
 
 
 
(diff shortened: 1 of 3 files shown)


.github/workflows/add-issues-to-project.yml
```diff
@@ -4,8 +4,12 @@ on:
     types:
       - opened
```
julialawrence: If you add
```push:```
to the `on:` block you can test these workflows before the PR is merged.

File: .github/workflows/add-issues-to-project.yml [hunk 2]
```diff
@@ -4,8 +4,12 @@ on:
...
 
+permissions: {}
+
 jobs:
   add-to-project:
```
julialawrence: Same as above.

julialawrence (changes requested):

.github/workflows/add-issues-to-project.yml
Emterry: https://github.com/ministryofjustice/data-platform/actions/runs/4083720472/jobs/7039578006
ive had to generate a PAT and give it the project scope because I cant change the GITHUB_TOKEN as the project is not at the repo level. This action run although failed proves that it has relevant permissions. It fails because on push it doesnt have the required data thats part of an issue event

.github/workflows/add-issues-to-project.yml
```diff
@@ -4,6 +4,8 @@ on:
     types:
       - opened
 
+permissions: read-all
+
 jobs:
   add-to-project:
```
julialawrence: Can you try adding this here:
```
         permissions:
            repository-projects: write

Emterry:
ive had to generate a PAT and give it the project scope because I cant change the GITHUB_TOKEN as the project is not at the repo level. This action run although failed proves that it has relevant permissions. It fails because on push it doesnt have the required data thats part of an issue event

julialawrence (approved):

PR #57, opened 2023-02-06
👷 Add tech docs paths for gh action

bagg3rs:
This PR adds a must include only `source` and `config` folders to workflow, this will stop GitHub pages from publishing all merges to main.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,6 +5,9 @@ on:
   push:
     branches:
       - "main"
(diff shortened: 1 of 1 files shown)


julialawrence:
Thank you!!! Looks good, but I'd recommend adding the workflow file itself to the paths.

julialawrence (dismissed):
Approved with a comment.

julialawrence (approved):
👍

PR #85, opened 2023-02-09
👨‍🔧 Fixed incorrect version references in feature installers

### LINKED ISSUE #84: 🐞 Devcontainer feature versions are only using latest
Features that have multiple tools within them are using the following

```bash
#!/usr/bin/env bash

set -e

source /usr/local/bin/devcontainer-utils

VERSION=${VERSION:-"latest"}
```

instead they should use

```bash
#!/usr/bin/env bash

set -e

source /usr/local/bin/devcontainer-utils

VERSION=${EXAMPLETOOLVERSION:-"latest"}
```

jacobwoffenden:
Resolves #84


julialawrence (approved):
LGTM

PR #102, opened 2023-02-14
:memo: Add adr-000 and adr index to tech-docs

bagg3rs:
This PR adds an adr index and adr000 to tech-docs


jacobwoffenden (approved):
Looks great to me!

PR #170, opened 2023-03-03
:green_heart: Update documentation workflow

jacobwoffenden:
This pull request updates the skipped image URL and also moves URL Check to after `deploy` as it checks links against the artifact but references the live site, which might have not been deployed yet as it was in parallel


julialawrence (approved):
LGTM

bagg3rs (approved):

PR #174, opened 2023-03-07
✨ Add Simulated Data Producer migration

jacobwoffenden:
Creates Kubernetes manifests for running migration against Cloud Platform RDS

I have validated that the Kubernetes manifests work, so merging this will run it again via GitHub Actions, duplicating the RDS data, but that is OK.


georges1996 (dismissed):

scripts/simulated-data-producer/configmap.yml
```diff
@@ -0,0 +1,52 @@
...
+
+    # push to postgres
+    DB_ENDPOINT=os.environ.get('DB_ENDPOINT')
+    DB_USERNAME=os.environ.get('DB_USERNAME')
+    DB_PASSWORD=os.environ.get('DB_PASSWORD')
+    DB_NAME=os.environ.get('DB_NAME')
+
+    engine = create_engine(f'postgresql://{DB_USERNAME}:{DB_PASSWORD}@{DB_ENDPOINT}:5432/{DB_NAME}')
+    df = pd.read_csv(os.path.join("data", adjudications_file))
+    df.to_sql('adjudications', engine)
```
murdo-moj: I'm doing it through sqlalchemy to avoid having to write a schema for the table

PriyaBasker23: Can main,py and requirements.txt be as seperate file within scripts rather than yml file itself ?

georges1996 (dismissed):

scripts/simulated-data-producer/configmap.yml
jacobwoffenden: Yes, with the following

```
kubectl --namespace ${NAMESPACE} create configmap \
  ${CONFIGMAP_NAME} \
  --from-file=main.py \
  --from-file=requirements.txt
```

However we'd have to replace it everytime, see https://stackoverflow.com/questions/38216278/update-k8s-configmap-or-secret-without-deleting-the-existing-one

jacobwoffenden:
Update:

* Containers are not allowed to run as `root` on Cloud Platform, meaning we cannot use `docker.io/python:3.10-alpine3.17`

* I created an image based on Alpine 3.17, built using Apko (part of a wider piece of work to offer secure baseline images to Data Platform Products), however `pandas`, and by extension `numpy` do not install/build well on Alpine (see [link](https://stackoverflow.com/questions/49037742/why-does-it-take-ages-to-install-pandas-on-alpine-linux)). A _workaround_ for this is to use the Alpine packages for the required softwares
  * `requests==2.28.1` ➡️ https://pkgs.alpinelinux.org/package/v3.17/main/x86_64/py3-requests
  * `pandas==1.5.3` ➡️ https://pkgs.alpinelinux.org/package/v3.17/community/x86_64/py3-pandas
  * `SQLAlchemy==2.0.5.post1` ➡️ https://pkgs.alpinelinux.org/package/v3.17/community/x86_64/py3-sqlalchemy
  * `psycopg2-binary==2.9.5` ➡️ https://pkgs.alpinelinux.org/package/v3.17/community/x86_64/py3-psycopg2

Alternatively, I have just written up a document for connecting to RDS from our local machines, however I need to update access to the Cloud Platform namespace

julialawrence (approved):

PR #181, opened 2023-03-07
✨ Add Alpine container images

jacobwoffenden:
Add Alpine based images built with [apko](https://github.com/chainguard-dev/apko). Unable to use vanilla images as they run as root ☹️

This is directly related to #174, currently we cannot execute jobs on Cloud Platform.


.github/workflows/container-images.yml
```diff
@@ -0,0 +1,139 @@
...
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+
+      - name: Install cosign
+        id: install_cosign
+        uses: sigstore/cosign-installer@c3667d99424e7e6047999fb6246c0da843953c65 # v3.0.1
+
+      - name: Prepare build paramerters
+        id: prepare
+        run: |
+          echo "name=$(jq -r '.name' container-images/${{ matrix.images }}/config.json)" >> ${GITHUB_OUTPUT}
```
julialawrence: I think you can use github_env to pass values between steps. Github_outputs is for moving things between jobs.

.github/workflows/container-images.yml
```diff
@@ -0,0 +1,139 @@
...
+            ghcr.io/ministryofjustice/data-platform-${{ steps.prepare.outputs.name }}:latest
+
+      - name: Push image
+        if: github.ref == 'refs/heads/main'
+        id: push_image
+        run: |
+          docker push ghcr.io/ministryofjustice/data-platform-${{ steps.prepare.outputs.name }}:${{ steps.prepare.outputs.version }}
+          docker push ghcr.io/ministryofjustice/data-platform-${{ steps.prepare.outputs.name }}:latest
+
+      - name: Sign image
```
julialawrence, line 130: The use of ghcr feels not so tactical at this point.

jacobwoffenden: 🐳 will move to DockerHub

julialawrence (approved):
LGTM

PR #185, opened 2023-03-08
patch '"custom_branch_policies" and "protected_branches" cannot have …

tom-webber:
…the same value []'

add template_repo var to github_repository resource and specify for data-platform-apps


julialawrence (approved):
LGTM

PR #196, opened 2023-03-09
🩹 temp fix for app repo migration to moj org

tom-webber:
## What has changed?

Added switches to the repo- and branch-protection rules for data-platform-apps dependent on the repository `type = "app"`

## Why is this needed?

We're migrating mojap org (`moj-analytical-services`) app repos into newly created `ministryofjustice` org repos through a [migration script](https://github.com/ministryofjustice/analytical-platform-migration-scripts/blob/ANPL-1422/repository-migration-script/scripts/repo_migration.sh), and the existing protection rules scupper this process.

## What should the reviewer concentrate on?

Ensuring all of the altered cases have the correct default, and check whether there are any missed protection rules that'll likely throw spanners (🔧 ) into the efforts.

### DIFF
--- a/terraform/github/modules/repository/main.tf
+++ b/terraform/github/modules/repository/main.tf
@@ -16,7 +16,7 @@ resource "github_repository" "default" {
   has_wiki               = var.type == "core" ? true : false
   has_downloads          = true
   is_template            = var.type == "template" ? true : false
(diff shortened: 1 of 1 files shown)


tom-webber:
🔧

julialawrence (approved):

PR #207, opened 2023-03-14
update migration apps json to include existing repo name

tom-webber:
## What has changed?
Added the source repo name as well as the new repo name for migration apps in the json for ap-migration-apps

## Why is this needed?
A script is to be added that requires both the existing repo name and new repo name for migration across GitHub orgs.

## What should the reviewer concentrate on?
The glory of consistent formatting that is coming in with the new names


julialawrence (approved):
Thank you for this!

PR #211, opened 2023-03-15
add repo content migration stage

tom-webber:
## What has changed?
add repo migration script
add migration null_resource triggered by all app repos being created 
add core-team to data-platform-app-template repo

## Why is this needed?

We're migrating mojap org (`moj-analytical-services`) app repos into newly created `ministryofjustice` org repos, and this adds that migration stage after the repos are created


terraform/github/migration.tf
```diff
@@ -0,0 +1,21 @@
...
+  provisioner "local-exec" {
+    interpreter = ["bash", "-c"]
+    command     = <<EOF
+  mkdir ${path.module}/tmp_migration \
+   && cd ${path.module}/tmp_migration \
+   && chmod +x "${path.module}/scripts/repo-content-migration/repo-migration.sh" \
+   && "${path.module}/scripts/repo-content-migration/repo-migration.sh" \
+   -s ${each.value.source_repo_name} \
+   -t ${module.data-platform-apps[each.value.name].repository.name} \
+   && rm -rf ${path.module}/tmp_migration
```
jacobwoffenden, line 18: This feels a bit complicated, can this logic be moved into the script?

jacobwoffenden (changes requested):

terraform/github/migration.tf
julialawrence: This is only relevant for the initial migration run and then it's gonna be removed. It's useful to see it inline in case there are issues.

jacobwoffenden (approved):

PR #222, opened 2023-03-16
Adding depends-on block for repository ownership.

julialawrence:
When destroying repos, we are removing ourselves as owners before we have a chance to destroy the repos, thus blocking ourselves. This adds a depends_on block which should theoretically solve this issue.


jacobwoffenden (changes requested):
Should we make `archive_on_delete` match the actual flag `archive_on_destroy`?

julialawrence:
> Should we make `archive_on_delete` match the actual flag `archive_on_destroy`?

Yes.

terraform/github/main.tf
```diff
@@ -165,4 +166,5 @@ module "data_platform_labs_team" {
   maintainers = local.data_platform_labs_maintainers
   members     = local.all_members_data_platform_labs
   ci          = local.ci_users
+  depends_on  = [module.data-platform]
```
tom-webber: Why is this dependency with the `data_platform_labs_team` module? Is it arbitrary? Could it sit within any team module?

jacobwoffenden (dismissed):
~LGTM~ 👀

terraform/github/main.tf
julialawrence: Ugh....

jacobwoffenden (approved):
LGTM

PR #227, opened 2023-03-19
Surgery on existing analytical-platform app role trust policies in production data account

julialawrence:
https://dsdmoj.atlassian.net/browse/ANPL-1449

## Context:

Applications currently hosted in the alpha cluster already have roles that define the required permissions in the Production Data account (including S3 and other resources.) To minimise the work of migration, a decision was made to reuse the roles for the apps going into Cloud Platform, as creating new roles and amending policies on existing resources was considered to be more work. However, for the existing roles to work, the trust policies (also now known as trust relationships) needed to be modified to be assumable by the new CP service accounts via the Cloud Platform OIDC provider. 

## What this PR does:

* Retrieves existing app roles in the data account, as well as role details such as `arn` and `assume-role-policy`
* Uses the agreed naming conventions for migrated apps to generate a trust statement that allows the new service account to assume the existing role from Cloud Platform via IRSA and OIDC.
* Creates a new trust policy that consists of the existing policy (for dual-running) and the new trust statement to create an `updated_trust` policy document (the code was written in such a way as to avoid attempts to attach the same statement to the policy on every rerun)
* Updates the role's trust policy with the new updated trust.

## Additionally

The PR also plants the existing role `arn` as an environment secret in the new application repo


jacobwoffenden (approved):
Looking good!

PR #228, opened 2023-03-20
✨ Analytical Platform container images ✨ 

### LINKED ISSUE #218: ✨ Create Analytical Platform container images
## User Story

As a DevOps/Platform engineer, I want to control the supply chain of our container images 

## User Type(s)

DevOps/Platform engineers

## Value

We have visibility over the entire image and its contents

## Proposal

If we can use Alpine, we should look into that, but if we need to use a Debian image, it should at the minimum be

`Containerfile`:

```Dockerfile
FROM public.ecr.aws/lts/ubuntu:22.04 
```

## Definition of done

- [ ] Images provide identical functionality
- [ ] Images do not run as root
- [ ] Images pass scanning checks
- [ ] Images are signed
- [ ] SBOMs are published

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

---

## Images

| Name | Source |
|:---:|:---:|

jacobwoffenden:
Resolves #218

### Headlines
- Moved Alpine built images to `/container-images/alpine`
- Created Ubuntu base image via `/container-images/ubuntu`
- Created RShiny image and variants. This is a special case image that uses Ubuntu but because of it's complex matrix, it can't be wedged into `/.github/workflows/container-images-ubuntu.yml`


.github/workflows/container-images-alpine.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         id: detect_changes
         uses: dorny/paths-filter@4512585405083f25c027a35db413c2b3b9006d50 # v2.11.1
         with:
-          filters: .github/workflows/path-filter/container-images.yml
+          filters: .github/workflows/path-filter/container-images-alpine.yml
```
julialawrence, line 37: Like this alot

julialawrence (approved):
LGTM

PR #229, opened 2023-03-21
reorder app migration list to reflect app migration batches

tom-webber:
## What changed
The ap_migration_apps json list was converted to yaml to allow for apps to be commented out and the 'whole' list used for the map used in the locals, rather than slices.
Added gold-scorecard-form and the first batch of apps to the repos to create.
(judicial reviews paused until manual repo is removed) linting and formatting

## Why?
Need the repos to exist so next steps can be carried out.
Yaml allows us to finely control the creation of app repos as we process the batches of apps during the migration


terraform/github/ap_migration_apps.yml
julialawrence, line 1: Why are we switching to yaml when JSON works?

julialawrence (changes requested):
?

terraform/github/ap_migration_apps.yml
julialawrence: We have the benefit via the magic of source control to recall the parts of the JSON file we remove that doesn't contain the tranche 1 apps. We have a fair bit of work still outstanding before the migration in less than 2 weeks and I am wondering if maybe we shouldn't focus effort on things that work. Please feel free to disagree. 🙂

tom-webber: As far as I understand, the output map from the json or from the yaml are the same. The change was purely to allow commenting out items in the file.
I do not mind which format we use to store the data.

bogdan-mania-moj (dismissed):

julialawrence (approved):
Looks good.

jacobwoffenden (approved):

PR #247, opened 2023-03-28
🔥 Replace hclq with hcl2json

jacobwoffenden:
`hclq` couldn't handle HCL2 😢

`hcl2json` can 🚀

```bash
$ cat data-platform-teams.tf | hcl2json | jq -r '.locals[].data_platform_core_teams[].members[]'
AlexVilela
bagg3rs
jemnery
bagg3rs
jacobwoffenden
julialawrence
Emterry
andyrogers1973
jhpyke
tamsinforbes
PriyaBasker23
hemeshpatel-moj
murdo-moj
simonsMOJ
```

Caveat is that `hcl2json` doesn't have a version flag to test against 🙃 but `-h` is good enough for validating it executes without issue


julialawrence (approved):
LGTM

PR #267, opened 2023-04-05
✨ Add YAML story template to repo

bagg3rs:
This adds the yml issue template


bagg3rs:
This will break the ZenHub template so unsure if people are using that method of adding issues

jacobwoffenden:
> This will break the ZenHub template so unsure if people are using that method of adding issues

I just use GitHub with the ZenHub extension

bagg3rs:
closing as this is now superseded by #1252

PR #293, opened 2023-04-12
🚑️ Fix PagerDuty scheduled action 🤜 BST 🤛 

jacobwoffenden:
- Fixed scheduling of PagerDuty notification
- Added `-upgrade` to Terraform workflow, we now use the lock file 🎉 
- Switched `.github/workflows/terraform-github.yml` to use `.github/workflows/reusable-terraform.yml`
- Upgraded Terraform version to `1.4.4`
- Pinned providers in `terraform/github/versions.tf` and modules in GitHub component
- Upgraded providers in `terraform/pagerduty/terraform.tf` and modules in PagerDuty component
- Added `scripts/pagerduty/list-vendors.sh` for listing vendors ad-hoc
- Switched GitHub component to get GitHub token from AWS Secrets Manager instead of GitHub Secrets, that way it can use the reusable workflow 🎉


jacobwoffenden:
Failing PagerDuty workflow is tracked in https://github.com/ministryofjustice/data-platform/issues/294

bagg3rs (approved):

PR #311, opened 2023-04-14
Move rshiny dockerfile from old repo

jhackett-ap:


jacobwoffenden (changes requested):
Looks good, but can you update the change log too please 🥺

container-images/rshiny/Dockerfile
```diff
@@ -1,69 +1,80 @@
-# hadolint global ignore=DL3008
+FROM ubuntu:focal
```
jhackett-ap, line 2:
```suggestion
# hadolint global ignore=DL3008
FROM ubuntu:focal
```

PR #389, opened 2023-05-02
🚑 Update API call for getting on-call user

jacobwoffenden:
It was just returning @bagg3rs, I've added the `since` and `until` parameters that were in the original shell script, and I've updated the message


jacobwoffenden:
Successfully ran https://github.com/ministryofjustice/data-platform/actions/runs/4859462330/jobs/8662126931

bagg3rs (approved):
🙏 🤣

PR #474, opened 2023-05-24
Splitting out IAM-related App Migration Code, Migrating OIDC in from IAM repo, Refactoring OIDC providers and workflows

julialawrence:
This PR does the following:

## BAU

* Migrates the OIDC component from the `analytical-platform-iam` repository to keep legacy and new code saparate.
* A slight refactoring of OIDC component's providers to remove the need to role variable to be passed in and aligning the provider declarations with other components in the repo.
* Default provider in OIDC now points at `management` rather than `landing`

## Application Migration

* Moves non-github or github-dependent app migration resources into their own component called `application-migration`. At present there's only one resource there but there will be more
* Refactored how the ECR Access OIDC Role used by new application repos to pull base images manages its trust. It will now generate the list of allowed repositories from the app_migration_apps.json. There is also a local defined to add repositories that are not app ones for https://dsdmoj.atlassian.net/browse/ANPL-1568
* Moved `ap_migration_apps.json` out of `github` and to `configuration` directory so it can be used by multiple components.
* Reconfigure `github` workflow to also trigger when changes to the above json file have been made.


.github/workflows/terraform-application-migration.yml
```diff
@@ -0,0 +1,34 @@
...
+  terraform:
+    name: Terraform
+    permissions:
+      contents: read
+      id-token: write
+    uses: ./.github/workflows/reusable-terraform.yml
+    with:
+      working-directory: terraform/application-migration
+      aws-role-to-assume: arn:aws:iam::042130406152:role/GlobalGitHubActionAccess
+      terraform-version: 1.4.5
```
jacobwoffenden:
```suggestion
      terraform-version: 1.4.6
```

.github/workflows/terraform-oidc.yml
```diff
@@ -0,0 +1,34 @@
...
+  terraform:
+    name: Terraform
+    permissions:
+      contents: read
+      id-token: write
+    uses: ./.github/workflows/reusable-terraform.yml
+    with:
+      working-directory: terraform/oidc
+      aws-role-to-assume: arn:aws:iam::042130406152:role/GlobalGitHubActionAccess
+      terraform-version: 1.4.5
```
jacobwoffenden:
```suggestion
      terraform-version: 1.4.6
```

jacobwoffenden (changes requested):
Just a minor Terraform version bump

jacobwoffenden (approved):

PR #547, opened 2023-06-05
🎨 Rename PagerDuty daily rota functionality

jacobwoffenden:
To remove the existence of doubt as to whether we are "on call", I've renamed this function to "PagerDuty Rota to Slack"

I've also renamed the Slack app to "PagerDuty Rota"


julialawrence (approved):
Thank you kindly!

PR #548, opened 2023-06-05
🚑 Fix Terraform providers

jacobwoffenden:
Dependabot is going wild.

This PR bumps version to latest stable AND removes child module provider requirements as they are inherited from the root component 

- Bumped `terraform/application-migration`
- Bumped `terraform/aws/analytical-platform-data-engineering-production/10ds`
- Bumped `terraform/github`
- Removed `terraform/github` child module provider requirements
- Removed `terraform/pagerduty` child module provider requirements


julialawrence (approved):
👍

PR #570, opened 2023-06-09
temporarily add ecs to the AP dev account for testing

jhackett-ap:
We're testing ECS+fargate as a possible backend for hosting the tools in the future. PoC state, will clean up after we're done (by end of w/c 12th june)


jacobwoffenden:
@jhackett-ap can this be closed?

PR #586, opened 2023-06-13
Change directory layout of github components.

julialawrence:
This moves our current github configuration to `terraform/github/data-platform` in order to allow future splitting out of the application migration github components.

The location of the state file is updated to reflect the directory structure. Migrate-state operation was carried out locally.


jacobwoffenden (approved):
approving with no questions asked

Issue #632, opened 2023-06-20
👥 Create new GitHub Team for editorial team


bagg3rs:
## User Story

As a Technical Writer
I want to be able to create pull requests and push changes 
So that changes can be raised on our documentation / repo / README.

## Value

We will need to add more team members who will need to push and raise PRs and currently we only do not have an appropriate group. 
Least privilege

### Proposal
Create a new GH team `data-platform-tech-writers`? and add Haymon into that team and apply that to our data-platform repositories.

## Definition of done

- [ ] terraform updated with group and assignment
- [ ] another team member has reviewed
- [ ] tests are green

## Reference

[How to write good user stories](https://www.gov.uk/service-manual/agile-delivery/writing-user-stories)

jacobwoffenden:
potential pedantry: can we use non shortened names, e.g. `data-platform-technical-writers`

I know our groups currently are the short versions (`data-platform-core-infra`, `data-platform-tech-archs`), but I'm going to refactor them _soon_

bagg3rs:
I am going to use data-platform-writers, but going to add him to an existing group for now

jacobwoffenden:
covered by @ministryofjustice/data-platform-user-centered-design

PR #642, opened 2023-06-20
Add Haymon to data engineering group for write access

bagg3rs:
Adds Haymon to data-engineering githhub group to give `writer` permission, we will add an additional group as part of #632


julialawrence (approved):
👍

PR #646, opened 2023-06-21
build and push to ecr workflow

LavMatt:
- workflow to build and push dockerfile now works
- added changelog
- changed csv parsing library from default to OpenCsvSerde in a version 0.0.2


jacobwoffenden:
@LavMatt I've refactored the workflows, you can just update your application https://github.com/ministryofjustice/data-platform/tree/main/containers/daap-athena-load

PR #774, opened 2023-07-05
Add derived data proposal B - data producers

SoumayaMauthoorMOJ:
Users who derive data based on data on the Data Platform are classified as Data Producers, even though they may also be data consumers.

See https://github.com/ministryofjustice/data-platform/pull/775 for alternate proposal


SoumayaMauthoorMOJ:
@bagg3rs as discussed

PR #775, opened 2023-07-05
add derived data proposal A - data consumers

SoumayaMauthoorMOJ:
Data consumers can derive new data based on data on the Data Platform.

An alternate diagram is available https://github.com/ministryofjustice/data-platform/pull/774


SoumayaMauthoorMOJ:
@bagg3rs as discussed

PR #781, opened 2023-07-06
❇️ Add version as a build argument

jacobwoffenden:
@LavMatt has asked for a build arg to be passed for `VERSION`, this passes it to all builds based on the version from the config file, up to the container if its used or not


jhpyke (approved):
Looks good to me!

PR #797, opened 2023-07-06
add probation-recruitment-app to app list json

tom-webber:
(also linting)

This adds the teams from the moj-analytical-services repo to the existing ministryofjustice repo


mitchdawson1982 (approved):
LGTM

PR #798, opened 2023-07-06
add derived data proposal C - both users

SoumayaMauthoorMOJ:
Both data producers and data consumers have the ability to derive data products based on data on the Data Platform.


bagg3rs (approved):

AlexVilela (approved):
LGTM

PR #856, opened 2023-07-14
🪄 Terraform QoL improvements

jacobwoffenden:
- Consolidates workflows using path filters
- Adds path filter generator for Terraform
- ~Adds automatic approval if no changes~
- Automatically approves dependabot updates if no changes
- Adds and updates comment with Terraform plan
- ~Only allows for one Terraform component to be updated at a time~

TODO:

- [x] Add PR comment with Terraform plan output
- [ ] ~Delete PR comment when new plan is generated?~


julialawrence (approved):
Looks great. One comment would be that maybe we should look at doing the components in a matrix when running the output plan step.

PR #904, opened 2023-07-20
add authorizer and presigne-url containers

bogdan-mania-moj:
also bumped python version to 3.11 on `get-glue-metadata`


jacobwoffenden (dismissed):
LGTM

bogdan-mania-moj:
FUBAR

PR #908, opened 2023-07-20
🩹 Specify static analysis override label for Terraform

jacobwoffenden:
Automated approval won't work if static analysis fails, but that is out of scope for a dependabot PR

Related https://github.com/ministryofjustice/data-platform/issues/886


julialawrence (approved):
LGTM

PR #974, opened 2023-07-25
update glue crawler and glue database

lalithanagarur:
Creating new database alpha_vcms_data (as hyphens in database name are not permitted downstream in create-a-derived-table).
Creating new crawler.


terraform/aws/analytical-platform-data-production/s3-glue-crawler/glue-catalogs.tf
```diff
@@ -1,3 +1,7 @@
 resource "aws_glue_catalog_database" "alpha_vcms_data" {
   name = "alpha-vcms-data"
 }
+
+resource "aws_glue_catalog_database" "alpha_vcms_data_2" {
+  name = "alpha_vcms_data"
+}
```
jacobwoffenden, line 7:
```suggestion
resource "aws_glue_catalog_database" "alpha_vcms_data" {
  name = "alpha_vcms_data"
}
```

Can the original not be replaced?

lalithanagarur: `alpha-vcms-data` is currently a source database in Create a Derived Table. I want to create `alpha_vcms_data` first, run the crawler, then update the source in CaDT (i.e. remove `alpha-vcms-data` and add `alpha_vcms_data`). And then destroy the `alpha-vcms-data` glue catalog database and crawler.

jacobwoffenden: cool, when you are ready to retire `"alpha-vcms-data"`, get in touch and we can move resources around in the Terraform state

jacobwoffenden (approved):

PR #1088, opened 2023-08-04
adding logging to python for  base image to use in python lambdas

LavMatt:


containers/daap-python-base/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM public.ecr.aws/lambda/python:3.11
+
+ARG BASE_VERSION
+
+ENV BASE_VERSION="${VERSION}"
```
jacobwoffenden:
```suggestion
ARG VERSION

ENV BASE_VERSION="${VERSION}"
```

The argument passed in at build time is `VERSION` https://github.com/ministryofjustice/data-platform/blob/main/.github/workflows/reusable-workflow-containers.yml#L240

containers/daap-python-base/src/var/task/requirements.txt
```diff
@@ -0,0 +1 @@
+boto3==1.28.9
```
jacobwoffenden, line 1: Current version is `1.28.20` but as I understand, `boto3` needs to be pinned at a specific version due to a downstream package, `mojap-metadata` in `athena-load` image?

jacobwoffenden (changes requested):
Hey @LavMatt,

You need to do steps 5 and 6 from https://technical-documentation.data-platform.service.justice.gov.uk/documentation/platform/infrastructure/containers.html#creating-a-new-container

Also, it might be worth adding (a _currently_ undocumented feature) [`.dependabot-pip-ignore`](https://github.com/ministryofjustice/data-platform/blob/2366ca92f5a26b4ac32e785bb6e1e66529430f8f/scripts/dependabot/configuration-generator.sh#L38)

jacobwoffenden (approved):

PR #1162, opened 2023-08-14
Dp 865

murdo-moj:
- Upon requesting an upload URL, check to see if the database has a
  corresponding data product.
- Add 400, 404 responses
- Reinstate md5 hash check

### DIFF
--- a/containers/daap-presigned-url/CHANGELOG.md
+++ b/containers/daap-presigned-url/CHANGELOG.md
@@ -13,7 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Upon requesting an upload url, check to see if the database has a corresponding data product.
+- Upon requesting an upload URL, 
+check to see if the database has a corresponding data product.
 - Add 400, 404 responses
 - Reinstate md5 hash check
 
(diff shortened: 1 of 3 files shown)


containers/daap-presigned-url/CHANGELOG.md
```diff
@@ -8,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.1.0]
+
+### Changed
+
+- Upon requesting an upload url, check to see if the database has a corresponding data product.
```
jacobwoffenden:
```suggestion
- Upon requesting an upload URL, 
check to see if the database has a corresponding data product.
```
I think this should fix the linter

Emterry (approved):

containers/daap-presigned-url/src/var/task/presigned_url.py
```diff
@@ -16,8 +16,22 @@ def handler(event, context):
     database = event["queryStringParameters"]["database"]
     table = event["queryStringParameters"]["table"]
     amz_date = datetime.utcnow().strftime("%Y%m%dT%H%M%SZ")
-    # md5 = str(event["queryStringParameters"]["contentMD5"])
+    md5 = str(event["queryStringParameters"]["contentMD5"])
     uuid_string = str(uuid.uuid4())
+
+    if type(database) != str or type(table) != str:
```
murdo-moj, line 22: since these are passed in via a query string, this code might never fire?

PR #1216, opened 2023-08-18
Creating an inline role with amended trust policy to get around character limitations. This will need a permnanet fix with a change in the oidc module.

julialawrence:
This is mostly a bandaid to get around policy character limits with too many repos using the the same role.

This PR
* Renames the role created via the module to `ecr-oidc-dummy` and adds a dummy repo for assumption
* Uses flat terraform to author a new version of the `ecr-oidc` role with a refactored trust policy with two sets of conditions.

This requires a fix in the near future both to figure out how to limit the number of repos using the role and a PR to the OIDC module to allow more custom options for defining trust policy conditions.

### DIFF
--- a/terraform/aws/analytical-platform/oidc/data.tf
+++ b/terraform/aws/analytical-platform/oidc/data.tf
@@ -6,6 +6,10 @@ data "aws_caller_identity" "session" {
   provider = aws.session
 }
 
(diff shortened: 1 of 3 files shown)


Gary-H9 (approved):
LGTM

PR #1255, opened 2023-08-22
Add and import final existing bucket resources

jhpyke:
Following [the earlier PR](https://github.com/ministryofjustice/data-platform/pull/1251), this looks to implement the remaining bucket-related resources as imported. Please double check the changes planned to see you agree that they're aesthetic only!


tamsinforbes (approved):
plan looks good - 2 changes to be made in place, no creates/destroys
bucket policy matches existing
confirmed kube config file OK on public repo

PR #1261, opened 2023-08-23
DMT-1238/add-airflow-eips

tamsinforbes:
Add the EIPs

|Name|Private IP|EIP Allocation|
|----|----|----|
|airflow-dev-eu-west-1a|10.200.10.49|eipalloc-09e6dcb639d4373a9|
|airflow-dev-eu-west-1b|10.200.11.108|eipalloc-06fdb2c30ce465818|
|airflow-dev-eu-west-1c|10.200.12.174|eipalloc-0774b9a95bbccdfd8|

https://github.com/ministryofjustice/data-platform/issues/1238


terraform/aws/analytical-platform-data-production/airflow/network.tf
```diff
@@ -22,6 +22,16 @@ resource "aws_internet_gateway" "airflow_dev" {
   }
 }
 
+resource "aws_eip" "airflow_dev_eip" {
+  domain                    = "vpc"
+  count                     = length(var.eip_private_ips)
+  associate_with_private_ip = element(var.eip_private_ips, count.index)
```
julialawrence: I would remove this. Unless there's an operational reason to have specific IPs for a gateway (config on the opposite side, firewall rules, explicit usage in R53 record for example or others like it) you shouldn't use specific IP addresses when requesting EIPs. Part of the reason we maintain infra in code is to rebuild in case of DR, and it is entirely possible that this IP will be assigned to another entity by AWS that also lives in this subnet and will be unavailable for eip assignment. If this isn't explicitly referenced elsewhere, please remove this line. if it is specifically used elsewhere, ignore this change but we'll want to review if. we can make the association programmatically instead.

julialawrence (changes requested):
Plan looks good but one comment regarding the hard-coding of private IP addresses to EIPs.

julialawrence (approved):
Looks good. :)

PR #1307, opened 2023-08-29
Dp 998

murdo-moj:
* Added resync data product lambda
* Added recreate data product lambda


PriyaBasker23:
Since Logging module is available now. Can you please add logging module

containers/daap-recreate-data-product/src/var/task/recreate_data_product.py
```diff
@@ -0,0 +1,69 @@
...
+import boto3
+
+
+def handler(event, context):
+    data_product_to_recreate = event.get("data_product", "")
+    raw_data_bucket = os.environ.get("RAW_DATA_BUCKET", "")
+    curated_data_bucket = os.environ.get("CURATED_DATA_BUCKET", "")
+    athena_load_lambda = os.environ.get("ATHENA_LOAD_LAMBDA", "")
+
+    # Check data product has associated data
```
PriyaBasker23: This can made as module and called in by passing the boto3,  so it can be tested easily

containers/daap-recreate-data-product/src/var/task/recreate_data_product.py
```diff
@@ -0,0 +1,69 @@
...
+        s3_client=s3,
+    )
+
+    # Feed all data files through the load process again. Curated files are recreated.
+    # If there are over 1000 files, the lambda will get jammed up
+    aws_lambda = boto3.client("lambda")
+    for file in data_product_registration:
+        key = file["Key"]
+        payload = f'{{"detail":{{"bucket":{{"name":"{raw_data_bucket}"}}, "object":{{"key":"{key}"}}}}}}'
+        aws_lambda.invoke(
```
PriyaBasker23: Can you introduce threading. If we going to recreate a data product there could be files more than 1000 

eg : Pseudo code below 
```
# Number of invocations
num_invocations = 2000

# Create threads and start invoking Lambda
threads = []
for _ in range(num_invocations):
    thread = threading.Thread(target=invoke_lambda)
    threads.append(thread)
    thread.start()

# Wait for all threads to finish
for thread in threads:
    thread.join()
```

containers/daap-recreate-data-product/src/var/task/recreate_data_product.py
```diff
@@ -0,0 +1,69 @@
...
+    s3 = boto3.client("s3")
+    s3_response = s3.list_objects_v2(
+        Bucket=raw_data_bucket, Prefix=f"raw_data/{data_product_to_recreate}"
+    )
+    data_product_registration = s3_response.get("Contents", [])
+    if not any(data_product_registration):
+        raise ValueError(f"No data product found for {data_product_to_recreate}")
+    print(data_product_registration)
+
+    # Drop existing athena tables for that data product
```
PriyaBasker23: This could a module , which will be easy for testing

containers/daap-recreate-data-product/src/var/task/recreate_data_product.py
```diff
@@ -0,0 +1,69 @@
...
+    """Delete all files from a prefix in s3"""
+    paginator = s3_client.get_paginator("list_objects_v2")
+    pages = paginator.paginate(Bucket=bucket, Prefix=prefix)
+
+    delete_us = dict(Objects=[])
+    for item in pages.search("Contents"):
+        delete_us["Objects"].append(dict(Key=item["Key"]))
+
+        # delete once aws limit reached
+        if len(delete_us["Objects"]) >= 1000:
```
PriyaBasker23: Does the curated bucket have version enabled. If so can you list all objects including versions ? .We would be recreating dataset in the scenario  that its corrupted. Do we need to delete the versions as well ? `s3.list_object_versions(Bucket=bucket_name) - API is used to retrieve a list of all object versions (including delete markers) in the specified bucket`

murdo-moj: We do have versioning enabled, but I can't think of any reason we would need to delete the previous versions. Seems safer to just leave them there. What if we deleted uncorrupted historical files we needed?

containers/daap-recreate-data-product/src/var/task/recreate_data_product.py
murdo-moj: the API call here doesn't report on completion but I added a waiter every 1000 calls to it

PriyaBasker23: The lambda is invoking athena lambda ( which is responsible for processing the file . The processing time may vary. i am not sure if  **# Wait 5 mins for other invocations to finish
            time.sleep(360)** will work. It may take longer time finish processing. 

Since we are going to call this **Athena_load_lambda**  few times its better to send response back, so you can use threading 

If this means extra work in another lambda you can very well split it into another ticket.

PriyaBasker23: @AlexVilela

containers/daap-recreate-data-product/src/var/task/recreate_data_product.py
PriyaBasker23: ok

containers/daap-recreate-data-product/src/var/task/recreate_data_product.py
```diff
@@ -0,0 +1,107 @@
...
+        prefix=f"curated_data/database_name={data_product_to_recreate}/",
+    )
+
+    # Feed all data files through the load process again. Curated files are recreated.
+    aws_lambda = boto3.client("lambda")
+    count = 0
+    for item in data_product_pages.search("Contents"):
+        if count == 1000:
+            # Wait 5 mins for other invocations to finish
+            time.sleep(360)
```
murdo-moj:
```suggestion
            time.sleep(900)
```

containers/daap-recreate-data-product/src/var/task/recreate_data_product.py
```diff
@@ -0,0 +1,107 @@
...
+        bucket=curated_data_bucket,
+        prefix=f"curated_data/database_name={data_product_to_recreate}/",
+    )
+
+    # Feed all data files through the load process again. Curated files are recreated.
+    aws_lambda = boto3.client("lambda")
+    count = 0
+    for item in data_product_pages.search("Contents"):
+        if count == 1000:
+            # Wait 5 mins for other invocations to finish
```
murdo-moj:
```suggestion
            # Wait 15 mins (max lambda processing time) for other invocations to finish
```

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+
+    # Find extraction timestamps in the raw area, and not in the curated area
+    paginator = s3.get_paginator("list_objects_v2")
+    curated_pages = paginator.paginate(
+        Bucket=curated_data_bucket, Prefix=curated_prefix
+    )
+
+    # key = "raw_data/data_product/table/extraction_timestamp=timestamp/file.csv"
+    raw_table_timestamps = set(
```
PriyaBasker23: Can be made as function , so you can test if its  emitting expected output

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
```diff
@@ -0,0 +1,99 @@
...
+        "/".join(item["Key"].split("/")[1:4])
+        for item in raw_pages.search("Contents")
+        if item["Size"] > 0
+    )
+
+    curated_table_timestamps = set()
+    for item in curated_pages.search("Contents"):
+        # key = "curated_data/database_name=data_product/table_name=table"
+        #       + "/extraction_timestamp=timestamp/file.parquet"
```
PriyaBasker23, line 51: You are making an assumption here that path will be in the fomat "curated_data/database_name=data_product/table_name=table"
+ "/extraction_timestamp=timestamp/file.parquet".  If the table structure is changed t will error. Can you do pattern matching to check if its correct format  or extract in some other way to get **database_name and table_name**

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
```diff
@@ -0,0 +1,141 @@
...
+        if page["KeyCount"] == 0:
+            error_text = f"No data product found for {data_product_prefix}"
+            logger.error(error_text)
+            logger.write_log_dict_to_s3_json(bucket=log_bucket, **s3_security_opts)
+            raise ValueError(error_text)
+    return pages
+
+
+def get_raw_data_unique_extraction_timestamps(raw_pages: PageIterator) -> set:
+    """
```
murdo-moj, line 111:
```suggestion
    """
    return the unique slugs of data product, table and extraction timestamp
```

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
```diff
@@ -0,0 +1,142 @@
...
+    return the unique slugs of data product, table and extraction timestamp
```
murdo-moj, line 112:
```suggestion
    return the unique slugs of data product, table and extraction timestamp
    designed for use with boto3's pageiterator and list_object_v2
```

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
```diff
@@ -0,0 +1,143 @@
...
+)
+s3_security_opts = {
+    "ACL": "bucket-owner-full-control",
+    "ServerSideEncryption": "AES256",
+}
+raw_data_bucket = os.environ.get("RAW_DATA_BUCKET", "")
+curated_data_bucket = os.environ.get("CURATED_DATA_BUCKET", "")
+log_bucket = os.environ.get("RAW_DATA_BUCKET", "")
+athena_load_lambda = os.environ.get("ATHENA_LOAD_LAMBDA", "")
+s3 = boto3.client("s3")
```
murdo-moj:
```suggestion
```

containers/daap-reload-data-product/src/var/task/reload_data_product.py
```diff
@@ -0,0 +1,107 @@
...
+        bucket=curated_data_bucket,
+        prefix=f"curated_data/database_name={data_product_to_recreate}/",
+    )
+
+    # Feed all data files through the load process again. Curated files are recreated.
+    aws_lambda = boto3.client("lambda")
+    count = 0
+    for item in data_product_pages.search("Contents"):
+        if count == 1000:
+            # Wait 5 mins for other invocations to finish
```
murdo-moj:
```suggestion
            # Wait 15 mins for other invocations to finish (max lambda duration)
```

containers/daap-reload-data-product/src/var/task/reload_data_product.py
```diff
@@ -0,0 +1,107 @@
...
+            time.sleep(360)
```
murdo-moj:
```suggestion
            time.sleep(900)
```

PriyaBasker23 (approved):

PR #1308, opened 2023-08-29
DMT-1238/add-private-route-tables

tamsinforbes:
Add nat gateways and private route tables and route table assocs
- 3 nat gateways one per AZ and public subnet
- 3 private route tables
- 3 route table associations between private subnets and private route tables

https://github.com/ministryofjustice/data-platform/issues/1238


jhpyke (approved):
Looks good to go

PR #1372, opened 2023-09-04
Dpl 1193 create metadata lambda

LavMatt:
added new lambda container image for a create data product metadata endpoint and also added a new module to the python base image. This is because the new `DataProductMetadata` class can be expanded and used by other api endpoints (that are yet to be created)

### DIFF
--- a/containers/daap-create-metadata/src/var/task/create_metadata.py
+++ b/containers/daap-create-metadata/src/var/task/create_metadata.py
@@ -40,7 +40,7 @@ def handler(event, context):
     data_product_metadata = DataProductMetadata(data_product_name, logger)
 
     if not data_product_metadata.metadata_exists:
-        data_product_metadata.validate_metadata(event["metadata"])
+        data_product_metadata.validate(event["metadata"])
     else:
         data_product_metadata.logger.write_log_dict_to_s3_json(
             os.environ["BUCKET_NAME"], **s3_security_opts
@@ -57,7 +57,7 @@ def handler(event, context):
         }
 
     if data_product_metadata.valid_metadata:
-        data_product_metadata.write_metadata_json_to_s3()
+        data_product_metadata.write_json_to_s3()
         data_product_metadata.logger.write_log_dict_to_s3_json(
             os.environ["BUCKET_NAME"], **s3_security_opts
         )
--- a/containers/daap-create-metadata/src/var/task/create_metadata.py
+++ b/containers/daap-create-metadata/src/var/task/create_metadata.py
@@ -29,7 +29,8 @@ def handler(event, context):
             "headers": {"Content-Type": "application/json"},
             "body": json.dumps(
                 {
-                    "message": "The name of the data product is missing, it must be specified in the metadata against the 'name' key"
+                    "message": "Data product name is missing, it must be specified \
+                        in the metadata against the 'name' key"
                 }
             ),
         }
@@ -80,7 +81,8 @@ def handler(event, context):
             "body": json.dumps(
                 {
                     "data_product_name": data_product_name,
-                    "message": f"Your metadata failed validation with this error: {data_product_metadata.error_traceback}",
+                    "message": f"Your metadata failed validation with \
+                        this error: {data_product_metadata.error_traceback}",
                 }
             ),
         }
--- a/containers/daap-create-metadata/src/var/task/create_metadata.py
+++ b/containers/daap-create-metadata/src/var/task/create_metadata.py
@@ -6,7 +6,6 @@ from data_product_metadata import DataProductMetadata
 
 
 def handler(event, context):
-
     logger = DataPlatformLogger(
         extra={
             "image_version": os.getenv("VERSION", "unknown"),
@@ -15,16 +14,24 @@ def handler(event, context):
         }
     )
 
+    s3_security_opts = {
+        "ACL": "bucket-owner-full-control",
+        "ServerSideEncryption": "AES256",
+    }
+
     try:
         data_product_name = event["metadata"]["name"]
     except KeyError:
         logger.error("The name of the data product is missing")
+        logger.write_log_dict_to_s3_json(os.environ["BUCKET_NAME"], **s3_security_opts)
         return {
             "statusCode": 400,
             "headers": {"Content-Type": "application/json"},
-            "body": json.dumps({
-                "message": "The name of the data product is missing, it must be specified in the metadata against the 'name' key"
-            })
+            "body": json.dumps(
+                {
+                    "message": "The name of the data product is missing, it must be specified in the metadata against the 'name' key"
+                }
+            ),
         }
 
     logger.add_extras({"data_product_name": data_product_name})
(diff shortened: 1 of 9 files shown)


containers/daap-create-metadata/src/var/task/create_metadata.py
```diff
@@ -0,0 +1,88 @@
...
+import os
+
+from data_platform_logging import DataPlatformLogger
+from data_product_metadata import DataProductMetadata
+
+
+def handler(event, context):
+    logger = DataPlatformLogger(
+        extra={
+            "image_version": os.getenv("VERSION", "unknown"),
```
murdo-moj, line 11: if we are only using the logger in lambdas, the image version and base image version can be set in the logger object and we don't have to have this code at the start of every lambda

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -0,0 +1,117 @@
...
+
+s3_client = boto3.client("s3")
+
+
+# These can be moved to separate module for all paths when work starts on that
+def get_bucket_name():
+    return os.environ["BUCKET_NAME"]
+
+
+def get_data_product_metadata_path(data_product_name):
```
MatMoore: Question: do we want all functions/classes in these shared modules to have type annotations? I don't have an opinion one way or the other, just wanted to check before I start contributing to this myself.

MatMoore: A small suggestion - we could use the newer [pathlib](https://docs.python.org/3/library/pathlib.html) instead of os.path here https://docs.python.org/3/library/pathlib.html - it does exactly the same things but it allows you to return `Path` objects instead of strings, which might be easier to work with downstream.

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -0,0 +1,117 @@
...
+from jsonschema import validate
+from jsonschema.exceptions import ValidationError
+
+from data_platform_logging import DataPlatformLogger
+from dataengineeringutils3.s3 import get_filepaths_from_s3_folder, read_json_from_s3
+
+s3_client = boto3.client("s3")
+
+
+# These can be moved to separate module for all paths when work starts on that
```
MatMoore, line 15: Sounds good 👍🏻

containers/daap-create-metadata/src/var/task/create_metadata.py
LavMatt: yeah i think that could work, part of the wider logger refactor? You mentioned changing the way it writes the json too which i think would be good. See you've created the issue for it already

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -0,0 +1,117 @@
...
+        "metadata.json"
+    )
+    return metadata_s3_path
+
+
+def get_data_product_metadata_spec_path(version: str = None) -> str:
+    # if version is None we'll get the latest version
+    if version is None:
+        file_paths = get_filepaths_from_s3_folder(
+            os.path.join("s3://", get_bucket_name(), "data_product_metadata_spec")
```
MatMoore, line 36: Were you thinking this `data_product_metadata_spec` path would also be defined in the new paths module as well? I think it would be helpful if all the configuration for what lives where could be in that one place.

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: just missed type for that, and yeah happy to use `pathlib`

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        versions = list(
+            {i for p in file_paths for i in p.split("/") if i.startswith("v")}
+        )
+        versions.sort()
```
MatMoore, line 41: This is something we should have a test for once the unit test PR is merged. We might run into order problems here (depending on the version format) since the default is going to be lexicographic ordering, e.g. 

```
>>> sorted(['v1', 'v10', 'v2'])
['v1', 'v10', 'v2']
```

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -0,0 +1,117 @@
...
+            else:
+                self.logger.error(f"Uknown error - {e}")
+                raise Exception(f"Uknown error - {e}")
+        else:
+            self.logger.info("version 1 of metadata already exists for this data product")
+            self.metadata_exists = True
+
+        return self
+
+    def validate_metadata(
```
MatMoore: [Minor suggestion] Since the class is called `DataProductMetadata` maybe we can drop the `_metadata` component from the method names. Then we would call it like

```python
metadata = DataProductMetadata(...)
metadata.validate(...)
metadata.write_to_s3(...)
```

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: actually types are there. Maybe you were looking at an earlier commit?

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -0,0 +1,117 @@
...
+        metadata_schema = read_json_from_s3(get_data_product_metadata_spec_path(metadata_schema_version))
+        try:
+            validate(instance=data_product_metadata, schema=metadata_schema)
+        except ValidationError:
+            self.error_traceback = traceback.format_exc()
+            self.logger.error(f"metadata has failed validation with error: {self.error_traceback}")
+            self.valid_metadata = False
+        else:
+            self.logger.info("metadata has passed validation")
+            self.valid_metadata = True
```
MatMoore, line 121: Is this object usable by calling code before you've passed in the `data_product_metadata` and `metadata_schema_version` to the validate method? If not, you could move these to the initializer.

I think it might also streamline the calling code if the validate function returns the boolean directly instead of `self`.

I'm thinking something along the lines of this:

```python
metadata = DataProductMetadata(data_product_name, data_product_metadata, metadata_schema_version)
if not metadata.validate():
  # return error response

metadata.write_to_s3()
```

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -0,0 +1,117 @@
...
+            self.data_product_metadata = data_product_metadata
+        return self
+
+    def write_metadata_json_to_s3(self) -> None:
```
MatMoore: I think this method has an assumption that the validate method has already been called first and the validation didn't fail.

If that's the case, I think this method should check that `data_product_metadata` is not None first, and raise some kind of validation error if not.

containers/daap-create-metadata/src/var/task/create_metadata.py
```diff
@@ -0,0 +1,64 @@
...
+            "base_image_version": os.getenv("BASE_VERSION", "unknown"),
+            "lambda_name": context.function_name,
+        }
+    )
+
+    try:
+        data_product_name = event["metadata"]["name"]
+    except KeyError:
+        logger.error("The name of the data product is missing")
+        return {
```
MatMoore, line 27: Not sure yet, but maybe we could have some helper functions to format these kinds of error responses? I think we might end up reusing some of these messages.

Also, from the perspective of someone using the API, they might find it useful to have a machine readable error code that is more specific than the HTTP status code (but we could worry about this later)

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -0,0 +1,117 @@
...
+        json_metadata = json.dumps(self.data_product_metadata)
+        s3_client.put_object(
+            Body=json_metadata,
+            Bucket=self.metadata_bucket,
+            Key=self.metadata_key,
+            **{
+                "ACL": "bucket-owner-full-control",
```
MatMoore: Question - could we move this configuration to the terraform configuration for the bucket, so the encryption etc is applied by default? It seems a bit risky to rely on us always passing this in the `put_object` call, as we could easily forget it, or do something inconsistent.

MatMoore (commented):
Hey Matt, I've done a review and left some suggestions. Bear in mind this is my first time looking at the code though, so feel free to ignore anything that doesn't make sense or will be handled in other tickets.

containers/daap-python-base/src/var/task/data_product_metadata.py
MatMoore: Ah yeah, I went through the earlier commits first, ignore that then 🙂

containers/daap-python-base/src/var/task/data_product_metadata.py
murdo-moj: it's enforced on the bucket policy for all objects which get added - we can have a central write_to_s3 function which sets the write opts so they are always correct? 

there's not an easy layer which could apply those settings before addition in s3

containers/daap-python-base/src/var/task/data_product_metadata.py
murdo-moj: I've been using os pathing, we should standardise our approach

containers/daap-create-metadata/src/var/task/create_metadata.py
murdo-moj: yea, maybe I should pick it up sooner rather than later

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: yep that's my thinking

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: good spot, yeah think this may need a slight re-working

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: My thinking was to expand this class to not only create the first data product metadata file, but subsequent updates to metadata. So need to test whether the v1 of the metadata exists already because if it does then a different endpoint should be used (but it'll use this class too) and so validation is not necessary if a version of the data product metadata already exists

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: will address during unit test development

MatMoore (approved):
All good from my point of view 👍🏻

julialawrence (approved):
Lgtm

PR #1375, opened 2023-09-05
DPL 879 - Add unit tests for get-glue-metadata container

### LINKED ISSUE #879: Write tests - unit  for get glue metadata lambda
Investigate how best to automate testing of lambda containers using pytest.

MatMoore:
Resolves #879 

This PR introduces pytest unit tests that use moto to mock out AWS services.

I think the final commit needs an extra review from @ministryofjustice/data-platform-core-infra  - this integrates pytest into the Github actions workflow that targets a single container. For now, I've restricted it to get-glue-metadata. Once we are confident it's working as intended we can extend it to the other containers.


containers/daap-get-glue-metadata/tests/unit/mock_test.py
```diff
@@ -1,2 +1,89 @@
...
+    test_table_name = 'test_table'
+    s3_location = f's3://my-bucket/{TEST_DATABASE_NAME}/{test_table_name}'
+    table_input = glue_table_input(test_table_name, s3_location)
+    glue_client.create_table(
+        DatabaseName=TEST_DATABASE_NAME,
+        TableInput=table_input
+    )
+
+    # Use the mock client in the lambda, not a real one
+    monkeypatch.setattr(boto3, 'client', lambda _name: glue_client)
```
MatMoore: I'm not sure if there is a better way to get the code under test to use the moto mock. Perhaps we could allow the lambda to take the client as an optional argument, but I didn't want to mess with the code itself in this PR.

containers/daap-get-glue-metadata/tests/unit/conftest.py
```diff
@@ -0,0 +1,41 @@
+import sys
+import os
+from os.path import dirname, join
+from dataclasses import dataclass
+
+from moto import mock_glue
+import boto3
+import pytest
+
+sys.path.append(join(dirname(__file__), '../', '../', 'src', 'var', 'task'))
```
MatMoore: This path hacking is necessary because our source is organised according to the container layout, rather than as a python package with `__init__.py` files.

Container code also depends on the shared libraries which are distributed in the base layer for the images.

.github/workflows/reusable-workflow-containers.yml
jacobwoffenden, line 1: I think I would prefer if application tests lived in their own workflow file instead of a dynamic job in containers, e.g. `test-daap-get-glue-metadata.yml` or maybe an even more generic `test-daap.yml`

MatMoore: @jacobwoffenden do you think it should be one workflow that always executes on push then, rather than being dynamically run with an `image` argument, like `reusable-workflow-containers.yaml` is?

I'm happy to rework this.

MatMoore: @jacobwoffenden I've now split this into `test-daap.yml`.

.github/workflows/test-daap.yml
```diff
@@ -0,0 +1,39 @@
...
+          python-version: 3.11
+          cache: pip
+          cache-dependency-path: containers/${{ matrix.container }}/tests/requirements.txt
+
+      - name: Install test dependencies
+        run: python -m pip install -r requirements.txt
+        working-directory: containers/${{ matrix.container }}/tests/
+
+      - name: Run tests
+        run: pytest
```
LavMatt: could we increase verbosity of output?
```suggestion
        run: pytest -vv

**Review by LavMatt — COMMENTED** [review]
looks good to me, one minor suggestion is all

**Comment by MatMoore** [comment]
@LavMatt @murdo-moj Would you mind taking another look? I've added some error handling to the lambda and included this in the unit tests, as discussed in standup:
- if arguments are missing, return 400
- if the table is not found, return 404

There are [a few other exceptions returned from get_table](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue/client/get_table.html) but none of them are anything the client can do anything about, so we can just let the lambda fail and the gateway will return a regular 500 response.

BTW I've raised the following ticket in the discussion board regarding formatting error responses in a consistent way:
https://trello.com/c/LeO1wiWP/39-helpers-for-formatting-json-responses

File: containers/daap-get-glue-metadata/tests/unit/mock_test.py (hunk 1 above)
Comment (PriyaBasker23) [reply containers/daap-get-glue-metadata/tests/unit/mock_test.py]: Alternately you can use the patch decorator to replace boto3.client with the custom function. But this seems ok to me

**Review by LavMatt — DISMISSED** [review]

**Comment by MatMoore** [comment]
@ministryofjustice/data-platform-core-infra would you mind reviewing this again? Thanks!

File: .github/workflows/test-daap.yml [hunk 4]
```diff
@@ -0,0 +1,39 @@
+---
+name: Test application code
```
jacobwoffenden:
```suggestion
name: Test DaaP Applications
```

.github/workflows/test-daap.yml
```diff
@@ -0,0 +1,39 @@
...
+
+on: # yamllint disable-line rule:truthy
+  pull_request:
+    branches:
+      - main
+    paths:
+      - containers/**
+  push:
+    paths:
+      - containers/**
```
jacobwoffenden:
```suggestion
    paths:
      - containers/daap-*/**
  push:
    paths:
      - containers/daap-*/**
```

There are none DaaP containers

.github/workflows/test-daap.yml
```diff
@@ -0,0 +1,39 @@
...
+  run-tests:
+    name: Test python
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        container:
+          - daap-get-glue-metadata
+    steps:
+      - uses: actions/checkout@v3
```
jacobwoffenden:
```suggestion
      - name: Checkout
        id: checkout
        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
```

.github/workflows/test-daap.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
```
jacobwoffenden:
```suggestion
      - name: Setup Python
        id: setup_python
        uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1 # v4.7.0
```

.github/workflows/test-daap.yml
jacobwoffenden, line 1: It might be worth implementing the same `detect-changes` functionality as https://github.com/ministryofjustice/data-platform/blob/main/.github/workflows/containers.yml#L21-L38 so we don't have to hardcode or trigger one DaaP apps that don't change

jacobwoffenden (changes requested):

.github/workflows/test-daap.yml
MatMoore: @jacobwoffenden I was wondering about that - I think I can implement that, but it would need to be a separate path filter config file so as to exclude the other containers that are in `.github/path-filter/containers.yml` and don't have python tests - is that what you had in mind? Something like `.github/path-filter/daap-python-containers.yml`?

Thanks for the other suggested changes - I will squash all these into one commit once I'm done.

containers/daap-get-glue-metadata/src/var/task/get_glue_metadata.py
```diff
@@ -26,6 +39,23 @@ def handler(event, context):
...
+    try:
+        resp = glue_client.get_table(DatabaseName=database, Name=table)
+    except glue_client.exceptions.EntityNotFoundException:
+        return {
+            "statusCode": 404,
+            "headers": {"Content-Type": "application/json"},
+            "body": json.dumps(
+                {
+                    "error": {
+                        "message": "Table {table} in database {database} does not exist in the glue catalog."
```
murdo-moj: missing f-string?
```suggestion
                        "message": f"Table {table} in database {database} does not exist in the glue catalog."
```

MatMoore:
@jacobwoffenden I think I've got this working with detect-changes now 🤞🏻 

I had to modify the bash script in `scripts/path-filter/configuration-generator.sh`  though. It's now detecting which containers have `_test.py` files, and if so it will trigger the pytest job. Will this be ok to merge?

scripts/path-filter/configuration-generator.sh
```diff
@@ -13,13 +13,23 @@ case ${MODE} in
     SEARCH_PATTERN=".terraform.lock.hcl"
     SKIP_FILE=".terraform-path-filter-ignore"
   ;;
+  pytest)
+    PATH_FILTER_CONFIGURATION_FILE=".github/path-filter/pytest-containers.yml"
```
MatMoore: rename to `pytest.yml`

jacobwoffenden (approved):
LGTM

PR #1384, opened 2023-09-05
DPL 1316 Add module for paths

MatMoore:
Currently we have a lot of hard-coded paths and code for extracting information from paths.

We want to extract this into one place in the base container layer, so that we don't duplicate logic between containers. Doing so will allow us to extend or modify our s3 bucket structure without having to change the whole codebase.

As a first step, I'm adding the new module. This has changed a little from my earlier draft; I decided to split out a new `ExtractionConfig` class from `DataProductConfig`, since I want DataProductConfig to map 1:1 with a data product, rather than having one for every uploaded file. The idea is that if you have an instance of `DataProductConfig` it should be easy to construct any paths you need by calling methods and properties.

I originally thought about making `DataProductConfig` take only the data product name as an argument, but it turns out I needed the `table_name` as well, since I don't think there is a way to derive this from the data product name / database name.

### `BucketPath(bucket: 'str', prefix: 'str')`
A path to an object in S3.
Readonly properties: `arn`, `parent` 

### `QueryTable(database:str, name: str)`
A reference to a table in athena.

### `DataProductConfig(name: str, table_name: str)`
Configures the name, S3 paths, and athena table names for a data product.

Properties:

- `name: str`
- `raw_data_prefix: BucketPath`
- `curated_data_prefix: BucketPath`
- `raw_data_table: QueryTable`
- `curated_data_table: QueryTable`

Methods:

- `raw_data_path(timestamp: datetime, uuid_value: UUID) -> BucketPath`
- `metadata_path() -> BucketPath`
- `extraction_config((timestamp: datetime, uuid_value: UUID) -> ExtractionConfig`

## `ExtractionConfig(data_product_config: DataProductConfig, path: BucketPath, timestamp: datetime)`
Metadata about a specific processing of raw data.

Static methods:

- `parse_extraction_timestamp(raw_data_key: str) -> datetime`
- `parse_from_uri(raw_data_uri, account_id=None) -> ExtractionConfig`

## Convenience functions
I've added wrapper functions for the classes to make it easier to refactor some of the existing code. These return plain strings instead of objects.

### `data_product_curated_data_prefix(database_name: 'str', table_name: 'str', bucket_name: 'str' = None) -> 'str'`
The S3 location for partitioned data files in parquet format. Bucket name is taken from the environment if not passed in.

### `data_product_metadata_path(data_product_name: 'str', bucket_name: 'str' = None) -> 'str'`
Generate the metadata path based on the data product name.  Bucket name is taken from the environment if not passed in.

### `data_product_raw_data_file_path(database_name: 'str', table_name: 'str', extraction_timestamp: 'datetime', uuid_value: 'uuid', bucket_name: 'str' = None) -> 'str'`
The S3 location for the raw uploaded data.  Bucket name is taken from the environment if not passed in.

### `get_bucket_name() -> 'str'`
Get the bucket name from the environment

### `get_account_id() -> 'str'`
Get the account ID from the environment / AWS configuration


containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -0,0 +1,123 @@
...
+
+    bucket: str
+    prefix: str
+
+    @property
+    def arn(self):
+        return f"s3://{self.bucket}/{self.prefix}"
+
+    @property
+    def parent(self):
```
MatMoore, line 44: this method might not be needed

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -0,0 +1,216 @@
...
+    extraction_timestamp: datetime  # Date the raw data was gathered
+    account_id: str  # AWS account ID
+
+    # S3 paths
+    raw_data_file: BucketPath
+    curated_data_path: BucketPath
+
+    # Athena database and table names
+    database_name: str
+    table_name: str
```
MatMoore: I think this should be replaced with `data_product_name`.

If we're talking about database_name and table_name it should specify whether we are talking about raw or curated data.

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -0,0 +1,219 @@
...
+)
+
+
+class BucketPath(NamedTuple):
+    """
+    A path to an object in S3
+    """
+
+    bucket: str
+    prefix: str
```
murdo-moj: in AWS they have a bucket and a key. is there any advantage in having prefix here rather than just carrying these concepts over from aws and calling this key?

murdo-moj:
I agree with your open question about how database_name and data product name have a bit of a namespace clash. We are assuming one database per data product, and the user should see data product in the input, but in the backend it is just a database in athena.

containers/daap-python-base/src/var/task/data_platform_paths.py
MatMoore: Nope, I was just confused about the correct AWS terminology here. I will change this to key. 👍🏻

containers/daap-python-base/src/var/task/data_platform_paths.py
murdo-moj: yeah I think it might not be, the main thing is to consistenly parse out the database and table from the key

containers/daap-python-base/src/var/task/data_platform_paths.py
murdo-moj: I can see sometimes you have a directory name in the prefix field, this does sometimes happen in s3 if you create folders in the s3 console but I think officially a key always relates to a specific file. I guess I'm trying to say that prefix might still be a valid concept in this class, but I can't think of a use case for it atm.

murdo-moj:
the planned API usage is super helpful btw, thanks for spelling it out

containers/daap-python-base/src/var/task/data_platform_paths.py
MatMoore: Yeah... I'm going to reserve `prefix` only for specific paths that do not point to a file.

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -0,0 +1,259 @@
...
+            os.path.join(
+                "curated_data",
+                f"database_name={name}",
+                f"table_name={table_name}",
+            )
+            + "/",
+        )
+
+        self.raw_data_table = QueryTable(
+            database=RAW_DATABASE_NAME, name=f"{table_name}_raw"
```
MatMoore, line 109: I noticed this table name does not contain the data product name. Could this cause problems if the table names clash?

(I based this on `containers/daap-athena-load/src/var/task/create_curated_athena_table.py`)

jacobwoffenden (dismissed):
LGTM

containers/daap-python-base/src/var/task/data_platform_paths.py
LavMatt: this table is temporary, and is deleted on every run. Saying that, there is an issue (https://github.com/ministryofjustice/data-platform/issues/1310) to add some random string to the name of it as it could cause potential issues with quick consecutive loads of the same data product

PriyaBasker23 (dismissed):

jacobwoffenden (dismissed):
LGTM

jacobwoffenden (approved):

murdo-moj (approved):

PR #1398, opened 2023-09-06
add path for metadata container

LavMatt:
needed to do this after as needed to build and push updated base image before building and pushing the new metadata container image (they were in the same PR)


julialawrence (approved):
Lgtm

PR #1442, opened 2023-09-08
DPL-1316 Refactor presigned url lambda to use shared path library

MatMoore:
This refactors the presigned URL lambda to use the new path construction module in our python base image. (see https://github.com/ministryofjustice/data-platform/pull/1384)

The goal of this refactor is that all the path construction logic is in one place so it's easier to change.

### DIFF
--- a/containers/daap-presigned-url/Dockerfile
+++ b/containers/daap-presigned-url/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/ministryofjustice/data-platform-daap-python-base:0.3.0
+FROM ghcr.io/ministryofjustice/data-platform-daap-python-base:0.5.0
 
 ARG VERSION
 
--- a/containers/daap-presigned-url/Dockerfile
+++ b/containers/daap-presigned-url/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/ministryofjustice/data-platform-daap-python-base:0.5.0
+FROM ghcr.io/ministryofjustice/data-platform-daap-python-base:1.0.0
 
 ARG VERSION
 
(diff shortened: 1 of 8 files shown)


containers/daap-presigned-url/tests/unit/mock_test.py
```diff
@@ -0,0 +1,97 @@
...
+    monkeypatch.setattr(boto3, "client", lambda _name: s3_client)
+    monkeypatch.setenv("BUCKET_NAME", bucket_name)
+    monkeypatch.setattr(uuid, "uuid4", lambda: a_uuid)
+
+    # Emulate the data product being set up in the bucket
+    s3_client.create_bucket(
+        Bucket=bucket_name,
+        CreateBucketConfiguration={"LocationConstraint": region_name},
+    )
+    s3_client.put_object(Bucket=bucket_name, Key=f"code/{database}/hello", Body="")
```
PriyaBasker23: Path doesnt seem relavant

PriyaBasker23 (dismissed):

jacobwoffenden (dismissed):

MatMoore:
Before merge: need to change the path being checked by the lambda to the metadata path instead of /code.

containers/daap-presigned-url/tests/unit/mock_test.py
MatMoore: @PriyaBasker23 I've now changed the check to the metadata path (v1 metadata file). Can you check the latest commit and reapprove if ok?

PriyaBasker23 (dismissed):

containers/daap-presigned-url/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/ministryofjustice/data-platform-daap-python-base:0.3.0
+FROM ghcr.io/ministryofjustice/data-platform-daap-python-base:0.5.0
```
jacobwoffenden: Could you bump this to `1.0.0` please

jacobwoffenden (changes requested):

containers/daap-presigned-url/Dockerfile
murdo-moj: were we gonna pin this at `:latest` for some images while we're developing?

jacobwoffenden (approved):

PriyaBasker23 (approved):

PR #1469, opened 2023-09-11
Refactor athena load lambda to use shared paths module

MatMoore:
This refactors the athena load lambda to use the new path construction module in our python base image. (see https://github.com/ministryofjustice/data-platform/pull/1384)

The goal of this refactor is that all the path construction logic is in one place so it's easier to change.

In addition to adding some unit tests, I manually tested this against the following scenarios:

#### Curated table exists, partition files exist
Preconditions:
- the curated table is present in AWS glue, e.g. example_prison_data_product.adjudications
- the data is partitioned and queryable via athena, e.g. "SELECT * FROM "example_prison_data_product"."adjudications" limit 10;"

Expected result:
- the lambda logs "partition for extraction_timestamp and table already exists so nothing more to be done"
- no table left over in data_products_raw database in AWS glue
#### Curated table exists, partition files don't exist
Preconditions
- the curated table is present in AWS glue, e.g. example_prison_data_product.adjudications
- there are no partition files in the curated_data directory, e.g. `s3://data-platform-products-development20230412155038755500000001/curated_data/database_name=example_prison_data_product/table_name=adjudications/extraction_timestamp=20230609T152430Z/`

Expected result:
- lambda logs "table does already exist but partition for timestamp does not"
- the curated table is still present in AWS glue, e.g. example_prison_data_product.adjudications
- no table left over in data_products_raw database in AWS glue
- the partitioned files now exist in the curated_data directory
- the data is partitioned and queryable via athena, e.g. "SELECT * FROM "example_prison_data_product"."adjudications" limit 10;"
#### Curated table doesn't exist, but partition files do exist
Preconditions:
- the curated table is not present in AWS glue, e.g. example_prison_data_product.adjudications
- there are partition files in the curated_data directory, e.g. `s3://data-platform-products-development20230412155038755500000001/curated_data/database_name=example_prison_data_product/table_name=adjudications/extraction_timestamp=20230609T152430Z/`

Expected result:
- the lambda logs "partition data exists but glue table does not"
- no table left over in data_products_raw database in AWS
- the data is partitioned and queryable via athena, e.g. "SELECT * FROM "example_prison_data_product"."adjudications" limit 10;"

#### Curated table doesn't exist, neither do partition files
Preconditions:
- the curated table is not present in AWS glue, e.g. example_prison_data_product.adjudications
- there are no partition files in the curated_data directory, e.g. `s3://data-platform-products-development20230412155038755500000001/curated_data/database_name=example_prison_data_product/table_name=adjudications/extraction_timestamp=20230609T152430Z/`

Expected result:
- the lambda logs "table and partition do not exist but other curated data do"
- no table left over in data_products_raw database in AWS
- the data is partitioned and queryable via athena, e.g. "SELECT * FROM "example_prison_data_product"."adjudications" limit 10;"


containers/daap-athena-load/src/var/task/athena_load_handler.py
```diff
@@ -28,29 +27,37 @@
...
+        f"config: {extraction.timestamp=} {data_product.raw_data_table=} {data_product.curated_data_table=}"
+    )
+
     metadata_types, metadata_str = infer_glue_schema(
         full_s3_path, "data_products_raw", logger=logger
     )
 
+    glue_client = boto3.client("glue")
+    athena_client = boto3.client("athena")
+    s3_client = boto3.client("s3")
```
murdo-moj: I think a few of these are defined at the beginning of the file

MatMoore: Yeah, they used to be, but I moved them into the handler so that their instantiation during the lifecycle of the tests (otherwise some stubbing broke).

I've since discovered a few different ways of handling import issues here: http://docs.getmoto.org/en/latest/docs/getting_started.html#what-about-those-pesky-imports

So if you prefer I can move this back and try and workaround this another way. Not sure if there is any other advantage/disadvantage to instantiating at module level vs handler level.

containers/daap-athena-load/src/var/task/create_curated_athena_table.py
```diff
@@ -1,33 +1,32 @@
...
-    table_name: str,
-    curated_path: str,
-    bucket: str,
-    extraction_timestamp: str,
+    data_product_config,
+    extraction_timestamp,
     metadata,
     logger: DataPlatformLogger,
     s3_security_opts,
+    athena_client,
```
MatMoore, line 16: Not sure if we need to add type hints for S3 clients. What do you think? I've left it for now because it seems like it requires adding a 3rd party library https://pypi.org/project/boto3-stubs/

containers/daap-athena-load/src/var/task/create_raw_athena_table.py
```diff
@@ -13,18 +13,17 @@ def create_raw_athena_table(
...
-        glue_client, "data_products_raw", logger, bucket, s3_security_opts
-    )
+    database_name = metadata_glue["DatabaseName"]
+    create_glue_database(glue_client, database_name, logger, bucket, s3_security_opts)
 
     # Create raw data table, recreating it if necessary
     table_name = metadata_glue["TableInput"]["Name"]
     try:
-        glue_client.delete_table(DatabaseName="data_products_raw", Name=table_name)
+        glue_client.delete_table(DatabaseName=database_name, Name=table_name)
```
MatMoore, line 22: This change is broken because when inferring the schema, we use the curated database name, but the raw table name

containers/daap-athena-load/tests/unit/athena_load_handler_test.py
```diff
@@ -0,0 +1,70 @@
...
+        sentinel.metadata_glue,
+        sentinel.metadata_glue_str,
+    )
+
+    athena_load_handler.handler(
+        {"detail": {"bucket": {"name": bucket_name}, "object": {"key": key}}},
+        fake_context,
+    )
+
+    mock_create_curated.assert_called_once_with(
```
MatMoore: Not sure if these assertions are worth keeping - I added them early in the development, but I don't think they provide much value now. We can just test the function runs to completion without raising an error.

containers/daap-athena-load/src/var/task/athena_load_handler.py
murdo-moj: leave it like this by all means, but remove the doubles from the beginning of the file eg athena and s3 on L10/11?

containers/daap-athena-load/src/var/task/create_curated_athena_table.py
murdo-moj: yeah just leave it if it's another lib

containers/daap-athena-load/tests/unit/infer_glue_schema_test.py
```diff
@@ -0,0 +1,194 @@
...
+
+    metadata_glue, metadata_glue_str = infer_glue_schema(
+        file_path=data_product.curated_data_prefix,
+        data_product_config=data_product,
+        logger=logger,
+        table_type="curated",
+        file_type="parquet",
+    )
+
+    assert metadata_glue == {
```
murdo-moj: to make the test files a bit more readable I guess you could test the datatypes bit alone like 
`assert metadata_glue["TableInput"]["StorageDescriptor"]["Columns]`

containers/daap-athena-load/tests/unit/athena_load_handler_test.py
murdo-moj: okay, remove them then

murdo-moj (dismissed):

containers/daap-athena-load/src/var/task/athena_load_handler.py
MatMoore: whoops - I didn't spot those duplicates. I've changed this now so they're defined at the module level but can be overridden by tests.

murdo-moj (approved):

Emterry (approved):
ltgm

PR #1506, opened 2023-09-12
♻️  Refactor `additional` security groups

mitchdawson1982:
This PR uses the term `additional` when describing security groups. This is because the Cluster Security Group - as defined in the AWS console, is implicitly created when an EKS cluster is spun up. This security group is not explicitly managed in Pulumi, that is to say, it is managed, but as part of the creation of the EKS cluster. That is the approach we are also taking going forward in Terraform. 

More detail is [here](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html) but the key point is: _"When you create a cluster, Amazon EKS creates a security group that's named` eks-cluster-sg-my-cluster-uniqueID`. This security group has the following default rules..."_ This is the security group which will be implicitly created by Terraform when the EKS cluster is created. 

Another PR will follow this which will remove the `moved` and `import` blocks.


julialawrence (approved):
LGTM

PR #1509, opened 2023-09-12
DMT-1508/add-high-mem-node-inst-type

### LINKED ISSUE #1508: Upgrade Airflow-Prod's High-Memory Node's Instance Class
Currently, Airflow Prod is running an `r6i.4xlarge` instance class for its 'high-memory' workloads. However, it turns out 128GB is not _quite_ high enough for some legitimate user workloads. As such, we'll be looking to increase the current instance to a `r6i.8xlarge` class, ideally as an 'in-place' update. I recommend familiarising yourself with the node group setup via the EKS area of the AWS Console. The high-memory node group can be seen via the 'compute' tab, and it's worth familiarising yourself with the current configuration. It's also worth familiarising yourself with the [monitoring DAG you can run](https://github.com/moj-analytical-services/airflow/blob/main/environments/prod/dags/monitoring/high_memory_daily_pass.py) to check you can schedule on the new node. It's worth noting it's not essential if it passes (although it should), but as long as the pod can get started you're doing alright. Any further questions, reach out to me via Slack and I can answer.

**Acceptance Criteria**

- [x] A PR with a plan detailing the changes brought on by the change in instance size is ready and can be applied.
- [x] All settings _other_ than instance size are retained, paying particular attention to kubernetes taints and tags, both of which are essential to scheduling on the node group.
- [x] Communications are made in #Ask-Data-Engineering detailing the planned timing of the change (should be done during business hours to allow for corrective action if needed prior to overnight runs), as well as updating once the new plan is applied.
- [x] Scheduling on the new node is tested via one of the monitoring DAGs.

tamsinforbes:
First pass to set instance types for standard as is in terraform and up for high-mem in prod

PROD
- set as is standard instance types: ["t3a.large", "t3.large", "t2.large"]
- up high-memory from ["r6i.4xlarge"] to ["r6i.8xlarge"]


jhpyke (approved):
lgtm - ready to roll out in prod

PR #1510, opened 2023-09-12
Adds EKS node Security Groups

Gary-H9:
The following block in `aws_eks_node_group` resources...

```terraform
# Allow external changes without Terraform plan difference
  lifecycle {
    ignore_changes = [scaling_config[0].desired_size]
  }
```

...has been included in relation to [this](https://mojdt.slack.com/archives/C04BRGD7KFH/p1694611558035269) thread.


terraform/aws/analytical-platform-data-production/airflow/variables.tf
```diff
@@ -89,9 +104,19 @@ variable "prod_eks_role_arn" {
   description = "ARN of role used by EKS cluster for Airflow-Prod"
 }
 
-variable "prod_cluster_sg_name" {
+variable "prod_eks_cluster_name" {
```
julialawrence, line 116: If I am understanding this right, this attribute is exported by the terraform EKS resource so there's no need to hardcode it:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster#cluster_id

terraform/aws/analytical-platform-data-production/airflow/variables.tf
```diff
@@ -89,9 +104,19 @@ variable "prod_eks_role_arn" {
...
+  type        = string
+  description = "Name of cluster for Airflow-Prod"
+}
+
+variable "prod_cluster_additional_sg_id" {
```
julialawrence: If I understand this right, this attribute is exported by the terraform EKS module, so there's no need to hardcode it:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster#cluster_security_group_id

terraform/aws/analytical-platform-data-production/airflow/variables.tf
```diff
@@ -89,9 +104,19 @@ variable "prod_eks_role_arn" {
...
+  type        = string
+  description = "Name of cluster additional security group for Airflow-Prod"
+}
+
+variable "prod_cluster_additional_sg_name" {
```
julialawrence: Assuming it's not a tag, I'd use the exported id to retrieve the name via a datasource call:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/security_group#name

Remember, in most data calls, all or most inputs are exported as attributes including name.

terraform/aws/analytical-platform-data-production/airflow/terraform.tfvars
```diff
@@ -47,10 +47,17 @@ prod_vpc_sg_name = "airflow-prod"
...
-
-prod_eks_role_arn    = "arn:aws:iam::593291632749:role/airflow-prod-eksRole-role-de6b4f5"
-prod_cluster_sg_name = "airflow-prod-eksClusterSecurityGroup-6ab84a6"
-prod_node_sg_id      = "sg-0f73e78564012634a"
+dev_eks_cluster_name           = "airflow-dev"
+dev_eks_role_arn               = "arn:aws:iam::593291632749:role/airflow-dev-eksRole-role-211908c"
+dev_cluster_additional_sg_name = "airflow-dev-eksClusterSecurityGroup-6a4dde4"
+dev_cluster_additional_sg_id   = "sg-0bcd3cf5dc6d7b314"
+dev_cluster_node_sg_id         = "sg-01930457ae391c7f0"
+dev_cluster_node_sg_name       = "airflow-dev-nodeSecurityGroup-f3e4ecd"
```
julialawrence: How many of these need hardcoding versus can be retrieved as attributes post eks cluster import?

julialawrence (changes requested):
Very nice start, but there might be a few more opportunities for levering terraform to avoid hard-coding values.

terraform/aws/analytical-platform-data-production/airflow/terraform.tfvars
Gary-H9: All but the `<env>_cluster_name` need to be hardcoded like this. Given this we feel it is more appropriate to keep them.

If we want to reference the clusters as data blocks this adds two new data block resources with potentially ugly names - due to 'dev' being deployed in the production account. e.g.

``` terraform
data "aws_eks_cluster" "analytical_platform_data_production_dev" {
  provider = aws.analytical-platform-data-production

  name = "airflow-dev"
}

data "aws_eks_cluster" "analytical_platform_data_production_prod" {
  provider = aws.analytical-platform-data-production

  name = "airflow-prod"
}
```

Ultimately this would lead to us updating the line name line in our `aws_eks_cluster` resource blocks from:
``` terraform
name     = var.dev_eks_cluster_name
```
to:
``` terraform
name     = data.aws_eks_cluster.analytical_platform_data_production_dev.id
```

This doesn't feel like win since we are:
- moving one variable away from comparable variables into a data block (it still needs to be defined in there)
- making data blocks with less than appealing names

terraform/aws/analytical-platform-data-production/airflow/variables.tf
Gary-H9: `prod_cluster_additional_sg_id` is manually defined, the EKS cluster SG is not defined in IaC.

terraform/aws/analytical-platform-data-production/airflow/variables.tf
Gary-H9: Same as the other one `prod_cluster_additional_sg_name` is manually defined, the EKS cluster SG is not defined in IaC.

julialawrence (approved):
LGTM

Issue #784, opened 2023-07-06
Change the backend storage architecture to have separate data lake style zones


PriyaBasker23:
Change the backend storage architecture to have separate data lake style zones (eg landing, raw, refined, curated) https://playbook.hackney.gov.uk/Data-Platform-Playbook/playbook/elements-of-the-platform/data-lake
Introduce the landing bucket layer to the pipeline.

**Definition of Done** 

* separate buckets for landing and raw/curated/Fail (3 buckets)
* Alter the base pathing object to reflect the new bucket structure
* Create new copy lambda to copy objects from landing to raw
* investigate and test disabling ACLs and potentially implement. See https://docs.aws.amazon.com/AmazonS3/latest/userguide/acls.html

See the diagram below 

![Image](https://github.com/ministryofjustice/data-platform/blob/main/docs/source/documentation/platform/images/daap-data-flow-architecture.excalidraw.png?raw=true)

murdo-moj:
Just doing #1755 as a separate PR first, so I can use the pathing in the landing to raw lambda I'm making

PR #1511, opened 2023-09-12
Dpl 1334 logging refactor

LavMatt:
This is related to https://github.com/ministryofjustice/data-platform/issues/1334
It does not complete the issue but implementing into lambda containers where used will form a separate PR
What it does do is:
1. Includes the security options for uploading to s3 within the module
2. Writes to s3 log file on every log call
3. Includes unit tests for the logging module
4. Moves function for generating log file path to the central paths module including a new unit test


containers/daap-python-base/src/var/task/data_platform_logging.py
```diff
@@ -8,11 +8,26 @@
...
+from botocore.exceptions import ClientError
+from data_platform_paths import data_product_log_bucket_and_key
 
 # aws lambda uses UTC so setting the timezone for correct time.
 os.putenv("TZ", "Europe/London")
 time.tzset()
 
+# additional arguments relating to data bucket security of data platform
+s3_security_opts = {
+    "ACL": "bucket-owner-full-control",
```
MatMoore, line 20: Side note: I was wondering about our use of ACLs as I thought they were a legacy thing in S3.

I found this documentation https://docs.aws.amazon.com/AmazonS3/latest/userguide/acls.html

> A majority of modern use cases in Amazon S3 no longer require the use of ACLs. **We recommend that you keep ACLs disabled, except in unusual circumstances where you need to control access for each object individually.** With ACLs disabled, you can use policies to control access to all objects in your bucket, regardless of who uploaded the objects to your bucket. For more information, see [Controlling ownership of objects and disabling ACLs for your bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html).

Maybe something to look into separately?

containers/daap-python-base/src/var/task/data_platform_logging.py
```diff
@@ -55,14 +70,26 @@ class DataPlatformLogger:
 
     def __init__(
         self,
+        lambda_name: str | None = None,
```
MatMoore: I think we could avoid passing this in by using the AWS_LAMBDA_FUNCTION_NAME environment variable

https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime

containers/daap-python-base/src/var/task/data_platform_logging.py
```diff
@@ -140,35 +167,36 @@ def error(self, msg, *args, **kwargs):
...
+
+        try:
+            response = s3_client.get_object(
+                Bucket=self.log_bucket_path.bucket, Key=self.log_bucket_path.key
             )
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "NoSuchKey":
+                jlines = ""
+            else:
+                raise e
```
MatMoore: I think we should completely suppress exceptions in this function, because it could cause the data ingestion to fail completely in a situation where it would otherwise continue working.

If there is something wrong with logging to s3, we can catch that separately by monitoring the bucket itself. We should be able to set up cloudwatch alerts to ping us if nothing has been written to the bucket in some time window.

Also I think we will still be able to access the default stdout logging if the s3 logging fails(?)

containers/daap-python-base/tests/unit/data_platform_logging_test.py
```diff
@@ -0,0 +1,155 @@
...
+    "lambda_name": "test_lambda",
+    "data_product_name": "test_database",
+    "table_name": "test_table",
+    "image_version": "test",
+    "base_image_version": "test",
+}
+
+
+@freeze_time("2023-01-01")
+def test__make_log_dict():
```
MatMoore: [Minor] I wouldn't normally bother testing this, as it's a private function that is already exercised by the logger tests below.

Other than this minor point these tests look useful though 👍🏻

MatMoore (changes requested):

containers/daap-python-base/src/var/task/data_platform_logging.py
PriyaBasker23: Can we check if the  ACL is disabled, will the lambdas or any AWS components accessing the data have permission . 

On reading the article, when ACLs are disabled, the bucket owner automatically owns and has full control over every object in the bucket. ( This wasnt the case before )

LavMatt: We could try disabling ACLs when we change the bucket architecture, so can add to #784?

containers/daap-python-base/src/var/task/data_platform_logging.py
LavMatt: This is true, should exit the method if unexpected error though to avoid creating a 1 line log file for the last log call. Yeah stdout should still be available if this method was to fail.

containers/daap-python-base/tests/unit/data_platform_logging_test.py
LavMatt: Did this one first, but you're right is covered by the last one so probably not needed

containers/daap-python-base/src/var/task/data_platform_logging.py
PriyaBasker23: yes thats good idea. Please add to the ticket

containers/daap-python-base/src/var/task/data_platform_logging.py
```diff
@@ -140,35 +168,40 @@ def error(self, msg, *args, **kwargs):
...
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "NoSuchKey":
+                jlines = ""
+            else:
+                # if an unexpected error we want to exit the method
+                # to avoid creating a log file with 1 line of last log
+                # but not raise an error, disrupting the application
+                print(e)
+                return None
         else:
```
MatMoore, line 196:
```suggestion
        except Exception:
            return None
        else:
```

MatMoore (dismissed):

MatMoore (approved):

PR #1547, opened 2023-09-15
Dlp metadata minor change plus more tests

LavMatt:
A minor change to `get_data_product_metadata_spec_path()` so semantic versioning is properly sorted.

Also adds further unit tests for the `data_product_metadata`

### DIFF
--- a/containers/daap-python-base/tests/unit/data_product_metadata_test.py
+++ b/containers/daap-python-base/tests/unit/data_product_metadata_test.py
@@ -15,7 +15,7 @@ test_schema_pass = {
     "dataProductOwnerDisplayName": "matt laverty",
     "email": "matthew.laverty@justice.gov.uk",
     "status": "draft",
-    "dpiaRequired": False
+    "dpiaRequired": False,
 }
 
 test_schema_fail = {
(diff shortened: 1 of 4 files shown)


containers/daap-python-base/tests/unit/data_product_metadata_test.py
```diff
@@ -31,3 +81,58 @@ def test_get_specific_metadata_spec_path(monkeypatch):
...
+
+def test_metadata_does_not_exist(s3_client, region_name, monkeypatch):
+    bucket_name = "bucket"
+    setup_bucket(bucket_name, s3_client, region_name, monkeypatch)
+    md = DataProductMetadata(test_schema_pass["name"], logging.getLogger())
+    assert not md.metadata_exists
+
+
+validation_inputs =[(test_schema_pass, True), (test_schema_fail, False)]
+# expected_outputs = [True, False]
```
MatMoore: I think this commented out line can be removed

containers/daap-python-base/tests/unit/data_product_metadata_test.py
```diff
@@ -31,3 +81,58 @@ def test_get_specific_metadata_spec_path(monkeypatch):
...
+    with NamedTemporaryFile(delete=True, suffix=".json") as tmp:
+        with open(tmp.name, "w", encoding="UTF-8") as f:
+            f.write(file_text)
+
+        s3_client.upload_file(
+            tmp.name, bucket_name, "metadata/test_product/v1.0/metadata.json"
+        )
+
+    md = DataProductMetadata(test_schema_pass["name"], logging.getLogger())
+    assert md.metadata_exists
```
MatMoore, line 104: You might have to monkeypatch the `s3_client` in the `data_product_metadata`, since it is a different instance to the one you are uploading the file to above.

A pattern I've used elsewhere is to make DataProductMetadata take an optional s3_client parameter that defaults to the module global value. That way you can pass it in from the tests.

MatMoore (dismissed):
Looks good to me, providing the tests pass ok

MatMoore (approved):

PR #1548, opened 2023-09-15
Excluding DaaP's Python based images from Dependabot

Emterry:


.github/dependabot.yml
```diff
@@ -27,83 +27,6 @@ updates:
       include: "scope"
     reviewers:
       - "ministryofjustice/data-platform-core-infra"
-  - package-ecosystem: "docker"
```
julialawrence, line 30: Is this not automatically generated?

PR #1549, opened 2023-09-15
Change data_platform_paths module to separate data products from data product elements, and allow multiple buckets

MatMoore:
This is a breaking change to the data_platform_paths module in the python base image.

1. Separate out the different buckets, and make them all attributes of  `DataProductConfig`. By default, these will be read from environment variables,  falling back to the `BUCKET_NAME` environment variable for backwards compatibility while we switch over. Note that all buckets must be provided in order to instantiate DataProductConfig objects, so we should set all the environment variables on all the lambdas.

2. Extract `DataProductElement` from `DataProductConfig`, so that a data product can have many elements, each with their own raw/curated tables and paths. This is needed to extract path logic from the `reload_data_product` lambda.

3. Add a random suffix to raw table names, as suggested in https://app.zenhub.com/workspaces/data-platform-labs-6439239ff652d87aef65297a/issues/gh/ministryofjustice/data-platform/1310

### DIFF
--- a/containers/daap-python-base/src/var/task/data_platform_paths.py
+++ b/containers/daap-python-base/src/var/task/data_platform_paths.py
@@ -112,7 +112,7 @@ def get_account_id() -> str:
 @dataclass
 class DataProductElement:
     """
-    A dataset within the data product. The curated data for each element
+    An entity within the data product. The curated data for each element
     is queryable in its own athena table.
     """
 
--- a/containers/daap-python-base/src/var/task/data_platform_paths.py
+++ b/containers/daap-python-base/src/var/task/data_platform_paths.py
@@ -15,9 +15,10 @@ from __future__ import annotations
 
 import os
 import re
+from dataclasses import dataclass, field
 from datetime import datetime
 from typing import NamedTuple
-from uuid import UUID
+from uuid import UUID, uuid4
 
 import boto3
 
@@ -26,6 +27,9 @@ EXTRACTION_TIMESTAMP_FORMAT = "%Y%m%dT%H%M%SZ"
 EXTRACTION_TIMESTAMP_REGEX = re.compile(
     r"^(.*)/(extraction_timestamp=)([0-9TZ]{1,16})/(.*)$"
 )
+
+# https://docs.aws.amazon.com/athena/latest/ug/tables-databases-columns-names.html
+MAX_IDENTIFIER_LENGTH = 255
 
 
 class BucketPath(NamedTuple):
@@ -63,11 +67,39 @@ class QueryTable(NamedTuple):
     name: str
 
 
-def get_bucket_name() -> str:
+def get_raw_data_bucket() -> str:
     """
-    Get the bucket name from the environment
+    Get the raw data bucket name from the environment
     """
-    return os.environ["BUCKET_NAME"]
+    return os.environ.get("RAW_DATA_BUCKET", "") or os.environ["BUCKET_NAME"]
+
+
+def get_curated_data_bucket() -> str:
+    """
+    Get the curated data bucket name from the environment
+    """
+    return os.environ.get("CURATED_DATA_BUCKET") or os.environ["BUCKET_NAME"]
+
+
+def get_log_bucket() -> str:
+    """
+    Get the log data bucket name from the environment
+    """
+    return os.environ.get("LOG_BUCKET") or os.environ["BUCKET_NAME"]
+
+
+def get_metadata_bucket() -> str:
+    """
+    Get the metadata data bucket name from the environment
+    """
+    return os.environ.get("METADATA_BUCKET") or os.environ["BUCKET_NAME"]
+
+
+def get_landing_zone_bucket() -> str:
+    """
+    Get the landing zone bucket name from the environment
+    """
+    return os.environ.get("LANDING_ZONE_BUCKET") or os.environ["BUCKET_NAME"]
 
 
 def get_account_id() -> str:
@@ -77,38 +109,51 @@ def get_account_id() -> str:
     return boto3.client("sts").get_caller_identity()["Account"]
 
 
-class DataProductConfig:
+@dataclass
+class DataProductElement:
     """
-    Configures the name, S3 paths, and athena table names for a data product.
+    A dataset within the data product. The curated data for each element
+    is queryable in its own athena table.
     """
 
-    def __init__(self, name: str, table_name: str, bucket_name: str | None = None):
-        """
-        Generate all the paths based on the data product name and a table name
-        """
-        if bucket_name is None:
-            bucket_name = get_bucket_name()
+    name: str
+    data_product: DataProductConfig
 
-        self.name = name
+    @staticmethod
+    def load(name, data_product_name):
+        data_product = DataProductConfig(name=data_product_name)
+        return DataProductElement(data_product=data_product, name=name)
 
-        self.raw_data_prefix = BucketPath(
-            bucket=bucket_name, key=os.path.join("raw_data", name, table_name) + "/"
+    @property
+    def raw_data_prefix(self):
+        return BucketPath(
+            bucket=self.data_product.raw_data_bucket,
+            key=os.path.join("raw_data", self.data_product.name, self.name) + "/",
         )
 
-        self.curated_data_prefix = BucketPath(
-            bucket_name,
+    @property
+    def curated_data_prefix(self):
+        return BucketPath(
+            self.data_product.curated_data_bucket,
             os.path.join(
                 "curated_data",
-                f"database_name={name}",
-                f"table_name={table_name}",
+                f"database_name={self.data_product.name}",
+                f"table_name={self.name}",
             )
             + "/",
         )
 
-        self.raw_data_table = QueryTable(
-            database=RAW_DATABASE_NAME, name=f"{table_name}_raw"
-        )
-        self.curated_data_table = QueryTable(database=name, name=table_name)
+    def raw_data_table_unique(self):
+        suffix = uuid4()
+        name = f"{self.name}_{suffix}_raw"
+        if len(name) > MAX_IDENTIFIER_LENGTH:
+            raise ValueError(f"Generated table name too long: {name}")
+
+        return QueryTable(database=RAW_DATABASE_NAME, name=name)
+
+    @property
+    def curated_data_table(self):
+        return QueryTable(database=self.data_product.name, name=self.name)
 
     def raw_data_path(self, timestamp: datetime, uuid_value: UUID) -> BucketPath:
         """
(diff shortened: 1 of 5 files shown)


containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -77,71 +109,130 @@ def get_account_id() -> str:
     return boto3.client("sts").get_caller_identity()["Account"]
 
 
-class DataProductConfig:
+@dataclass
+class DataProductElement:
     """
-    Configures the name, S3 paths, and athena table names for a data product.
+    A dataset within the data product. The curated data for each element
```
tom-webber:
```suggestion
    An entity within the data product. The curated data for each element
```

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -77,71 +109,130 @@ def get_account_id() -> str:
...
+            ),
+        )
+
+        return RawDataExtraction(timestamp=timestamp, element=self, path=path)
+
+
+@dataclass
+class DataProductConfig:
+    """
+    Configures the name, and S3 buckets for a data product.
```
tom-webber:
```suggestion
    Configures the name, elements, S3 buckets, and related paths for a data product.
```

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -77,71 +109,130 @@ def get_account_id() -> str:
...
-        if bucket_name is None:
-            bucket_name = get_bucket_name()
+    name: str
+    data_product: DataProductConfig
 
-        self.name = name
+    @staticmethod
+    def load(name, data_product_name):
+        data_product = DataProductConfig(name=data_product_name)
+        return DataProductElement(data_product=data_product, name=name)
```
tom-webber:
```suggestion
    def load(element_name, data_product_name):
        data_product = DataProductConfig(name=data_product_name)
        return DataProductElement(data_product=data_product, name=element_name)
```
Specifying `element_name` as this is a static method, so it's clear outside the context of the element class.

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -77,71 +109,130 @@ def get_account_id() -> str:
...
+            key=os.path.join("raw_data", self.data_product.name, self.name) + "/",
         )
 
-        self.curated_data_prefix = BucketPath(
-            bucket_name,
+    @property
+    def curated_data_prefix(self):
+        return BucketPath(
+            self.data_product.curated_data_bucket,
             os.path.join(
```
tom-webber:
```suggestion
            bucket=self.data_product.curated_data_bucket,
            key=os.path.join(
```

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -77,71 +109,130 @@ def get_account_id() -> str:
...
 
-        self.raw_data_table = QueryTable(
-            database=RAW_DATABASE_NAME, name=f"{table_name}_raw"
-        )
-        self.curated_data_table = QueryTable(database=name, name=table_name)
+    def raw_data_table_unique(self):
+        suffix = uuid4()
+        name = f"{self.name}_{suffix}_raw"
+        if len(name) > MAX_IDENTIFIER_LENGTH:
+            raise ValueError(f"Generated table name too long: {name}")
```
tom-webber, line 163: Do we want to try a shorter UUID if adding a UUID is too long?

The following suggestion is **completely** optional, but [gains 6 characters while losing almost no entropy](https://stackoverflow.com/a/55229089/10478115). It comes at the cost of the random string no longer being visually identifiable as UUID4.

e.g. 
```suggestion
        suffix_lim = suffix.hex[:12] + suffix.hex[13:16] + suffix.hex[17:]
        name = f"{self.name}_{suffix_lim}_raw"
        if len(name) > MAX_IDENTIFIER_LENGTH:
            raise ValueError(f"Generated table name too long: {name}")
```

PriyaBasker23:
1. Having docstring for all the methods would be good 
2. Example path output for functions would be intuitive 
3.  str(uuid_value) ( line 177) - extension from raw load should be preserved . Can be separate ticket 
4. Function from line 306 can be removed ( as discussed)

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -77,71 +109,130 @@ def get_account_id() -> str:
...
             )
             + "/",
         )
 
-        self.raw_data_table = QueryTable(
-            database=RAW_DATABASE_NAME, name=f"{table_name}_raw"
-        )
-        self.curated_data_table = QueryTable(database=name, name=table_name)
+    def raw_data_table_unique(self):
+        suffix = uuid4()
```
tom-webber, line 160: Should this UUID be the [same UUID as for the file path](https://github.com/ministryofjustice/data-platform/pull/1549/files#diff-f47b07dfcbf000013d2c9d5ec4009c2b492803a64344d8da662490171a6184dcR310)?

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -77,71 +109,163 @@ def get_account_id() -> str:
...
     def raw_data_path(self, timestamp: datetime, uuid_value: UUID) -> BucketPath:
         """
         Path to the raw data extracted at a particular timestamp.
+        E.g. raw_data/my-data-product/some-element/extraction_timestamp=20230101T000000Z/3d95ff89-b063-484d-b510-53742d0a6a64
         """
         return self.extraction_config(timestamp, uuid_value).path
 
-    @staticmethod
-    def _metadata_path(data_product_name: str, bucket_name: str | None = None):
+    def extraction_config(
```
tom-webber:
```suggestion
    def extraction_instance(
```

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -77,71 +109,163 @@ def get_account_id() -> str:
...
+        self, timestamp: datetime, uuid_value: UUID
+    ) -> RawDataExtraction:
         """
-        Path to the V1 metadata file
+        Config for the data extraction identified by uuid_value and timestamp.
```
tom-webber:
```suggestion
        Instance of the data extraction identified by uuid_value and timestamp.
```

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -77,71 +109,163 @@ def get_account_id() -> str:
...
+        E.g. ('my_data_product', 'some_element')
+        """
+        return QueryTable(database=self.data_product.name, name=self.name)
 
     def raw_data_path(self, timestamp: datetime, uuid_value: UUID) -> BucketPath:
         """
         Path to the raw data extracted at a particular timestamp.
+        E.g. raw_data/my-data-product/some-element/extraction_timestamp=20230101T000000Z/3d95ff89-b063-484d-b510-53742d0a6a64
         """
         return self.extraction_config(timestamp, uuid_value).path
```
tom-webber:
```suggestion
        return self.extraction_instance(timestamp, uuid_value).path
```
If changing [method name](https://github.com/ministryofjustice/data-platform/pull/1549/files#diff-f47b07dfcbf000013d2c9d5ec4009c2b492803a64344d8da662490171a6184dcR182), needs changing here too

containers/daap-python-base/tests/unit/data_platform_paths_test.py
```diff
@@ -140,11 +180,15 @@ def test_extraction_config():
...
+        raw_data_bucket="a-bucket",
+        curated_data_bucket="a-bucket",
+        landing_zone_bucket="a-bucket",
+        metadata_bucket="a-bucket",
     )
 
-    extraction = config.extraction_config(uuid_value=uuid_value, timestamp=timestamp)
+    element = config.element("some-table")
+
+    extraction = element.extraction_config(uuid_value=uuid_value, timestamp=timestamp)
```
tom-webber:
```suggestion
    extraction = element.extraction_instance(uuid_value=uuid_value, timestamp=timestamp)
```
If changing [method name](https://github.com/ministryofjustice/data-platform/pull/1549/files#diff-f47b07dfcbf000013d2c9d5ec4009c2b492803a64344d8da662490171a6184dcR182), needs changing here too

MatMoore:
@PriyaBasker23 I've committed your suggestions, except #3 which I've raised here https://github.com/orgs/ministryofjustice/projects/27/views/2?pane=issue&itemId=39286614

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, data_product_name: str, logger: DataPlatformLogger):
         self.data_product_name = data_product_name
         self.logger = logger
         bucket, key = split_bucket_and_key(
-            data_product_metadata_file_path(data_product_name)
+            DataProductConfig(data_product_name).metadata_path().uri
```
tom-webber:
```suggestion
            DataProductConfig(name=data_product_name).metadata_path().uri
```

containers/daap-python-base/tests/unit/data_platform_paths_test.py
```diff
@@ -39,79 +41,115 @@ def test_bucket_path_parent():
...
 
-    assert (
-        path
-        == f"s3://bucket/raw_data/data-product/table/extraction_timestamp=20230906T000000Z/{uuid_value}"
+    element = DataProductElement.load("some-table", "data-product")
+
+    raw_data_table = element.raw_data_table_unique()
+
+    assert re.match(
+        r"some-table_[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}_raw",
```
tom-webber, line 91: Note: not compatible (as is) with my suggestion for shorter UUIDs (not difficult to add though)

containers/daap-python-base/src/var/task/data_platform_paths.py
PriyaBasker23: its the name of the file

MatMoore: @tom-webber It could be, but I was thinking of this as a separate randomiser that is independent of the filename.

This is used to name a temporary table for the lambda to read from when populating or resyncing curated data.

The idea behind adding some random string was just to handle the edge case where the lambda runs twice in succession for the same data product element. That could be two different extractions or it could be the same extraction being repeated twice for some reason.

containers/daap-python-base/src/var/task/data_platform_paths.py
MatMoore: This is a nice suggestion, but I think I'll leave it for now as 255 characters is a lot to work with, and we might end up limiting the name length enough that this is fine anyway. If we hit this edge case in practice then I'm happy to revisit this

containers/daap-python-base/src/var/task/data_platform_paths.py
MatMoore: Thanks, good spot

tom-webber (approved):

PR #1563, opened 2023-09-18
Bump base version for Presigned URL lambda

MatMoore:
Update the base image to pick up the latest change to data_platform_paths.

The main change is to use the `DataProductElement` object for anything that requires a `table_name` instead of a `DataProductConfig` (which should be for configuration common to the whole data product`.

### DIFF
--- a/containers/daap-presigned-url/src/var/task/presigned_url.py
+++ b/containers/daap-presigned-url/src/var/task/presigned_url.py
@@ -5,7 +5,7 @@ from datetime import datetime
 
 import boto3
 from data_platform_logging import DataPlatformLogger
-from data_platform_paths import DataProductConfig
+from data_platform_paths import DataProductElement
 
 s3 = boto3.client("s3")
 
@@ -18,10 +18,10 @@ logger = DataPlatformLogger(
 
 
 def handler(event, context):
-    bucket_name = os.environ["BUCKET_NAME"]
     database = event["queryStringParameters"]["database"]
     table = event["queryStringParameters"]["table"]
     amz_date = datetime.utcnow()
+    formatted_date = amz_date.strftime("%Y%m%dT%H%M%SZ")
     md5 = str(event["queryStringParameters"]["contentMD5"])
     uuid_value = uuid.uuid4()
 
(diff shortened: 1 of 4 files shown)


containers/daap-presigned-url/src/var/task/presigned_url.py
```diff
@@ -103,16 +99,15 @@ def handler(event, context):
             ),
         }
 
-    if any(data_product_registration):
```
MatMoore, line 106: We can skip this extra check as we have already returned if there is no registration.

containers/daap-presigned-url/src/var/task/presigned_url.py
```diff
@@ -39,17 +39,14 @@ def handler(event, context):
             ),
         }
 
-    data_product = DataProductConfig(
-        name=database, table_name=table, bucket_name=bucket_name
-    )
-    file_name = data_product.raw_data_path(
-        timestamp=amz_date, uuid_value=uuid_value
-    ).key
+    model = DataProductElement.load(data_product_name=database, element_name=table)
```
murdo-moj:
```suggestion
    element = DataProductElement.load(data_product_name=database, element_name=table)
```

containers/daap-presigned-url/src/var/task/presigned_url.py
```diff
@@ -39,17 +39,14 @@ def handler(event, context):
...
+    data_product = model.data_product
```
murdo-moj:
```suggestion
    data_product = element.data_product
```

containers/daap-presigned-url/src/var/task/presigned_url.py
```diff
@@ -39,17 +39,14 @@ def handler(event, context):
...
+    raw_data_path = model.raw_data_path(timestamp=amz_date, uuid_value=uuid_value)
```
murdo-moj:
```suggestion
    raw_data_path = element.raw_data_path(timestamp=amz_date, uuid_value=uuid_value)
```

murdo-moj (approved):

PR #1574, opened 2023-09-18
Create metadata unit tests

murdo-moj:
* Test all the proper responses from this lambda
* Don't test the metadata validator object, as this is tested in the base image tests


containers/daap-create-metadata/src/var/task/create_metadata.py
```diff
@@ -35,7 +35,7 @@ def handler(event, context):
 
     data_product_metadata = DataProductMetadata(data_product_name, logger)
 
-    if not data_product_metadata.metadata_exists:
+    if data_product_metadata.metadata_exists is False:
```
tom-webber:
```suggestion
    if not data_product_metadata.metadata_exists:
```
Original is more standard python (`if not .exists`), and more typically what I'd expect.

If insisting on moving to comparing with False, use ` == False`

murdo-moj: 😢 ok

containers/daap-create-metadata/tests/requirements.txt
```diff
@@ -0,0 +1,7 @@
+boto3==1.28.40
```
MatMoore, line 1: in the other lambdas I've been including the source requirements here, i.e. starting the file with

`-r ../src/var/task/requirements.txt`

and then just adding the test requirements.

I don't know if at some point we want to switch over to poetry or something else that lets us define dev dependencies vs normal dependencies in one place?

MatMoore (dismissed):
looks good to me

MatMoore (approved):

MatMoore (approved):

PR #1644, opened 2023-09-22
Moved some functions to base image

PriyaBasker23:
Moved some common functions from resync processed file  container to base image

### DIFF
--- a/containers/daap-python-base/src/var/task/data_platform_paths.py
+++ b/containers/daap-python-base/src/var/task/data_platform_paths.py
@@ -19,6 +19,7 @@ from dataclasses import dataclass, field
 from datetime import datetime
 from typing import NamedTuple
 from uuid import UUID, uuid4
+from typing import Optional
 
 import boto3
 
@@ -113,7 +114,7 @@ def get_account_id() -> str:
     return boto3.client("sts").get_caller_identity()["Account"]
 
 
-def search_string_for_regex(string: str, regex: re.Pattern[str]) -> str:
+def search_string_for_regex(string: str, regex: re.Pattern[str]) -> Optional[str]:
     """Search a string for a regex pattern and return the first result"""
     search_output = regex.search(string)
     return search_output.groups()[0] if search_output else None
--- a/containers/daap-python-base/src/var/task/data_platform_paths.py
+++ b/containers/daap-python-base/src/var/task/data_platform_paths.py
@@ -27,6 +27,9 @@ EXTRACTION_TIMESTAMP_FORMAT = "%Y%m%dT%H%M%SZ"
 EXTRACTION_TIMESTAMP_REGEX = re.compile(
     r"^(.*)/(extraction_timestamp=)([0-9TZ]{1,16})/(.*)$"
 )
+DATABASE_NAME = re.compile(r"database_name=([^\/]*))\/")
+TABLE_NAME = re.compile(r"table_name=([^\/]*)\/")
+EXTRACTION_TIMESTAMP_CURATED = re.compile()
 
 # https://docs.aws.amazon.com/athena/latest/ug/tables-databases-columns-names.html
 MAX_IDENTIFIER_LENGTH = 255
@@ -108,25 +111,29 @@ def get_account_id() -> str:
     """
     return boto3.client("sts").get_caller_identity()["Account"]
 
+
 def search_string_for_regex(string: str, regex: str) -> str:
     """Search a string for a regex pattern and return the first result"""
-    database_name_search = re.search(regex, string)
-    if database_name_search:
-        return database_name_search.groups()[0]
+    search_output = regex.search(string)
+    if search_output:
+        return search_output.groups()[0]
     else:
         raise ValueError(f"{regex} not found in {string}")
 
 
-def database_name_regex() -> str:
-    return """database_name=([^\/]*)\/"""  # noqa: W605
+def extract_table_name_from_curated_path(string: str):
+    return search_string_for_regex(string=str, regex=TABLE_NAME)
 
 
-def table_name_regex() -> str:
-    return """table_name=([^\/]*)\/"""  # noqa: W605
+def extract_database_name_from_curated_path(string: str):
+    return search_string_for_regex(string=str, regex=DATABASE_NAME)
 
 
-def extraction_timestamp_regex() -> str:
-    return """(extraction_timestamp=[^\/]*)\/"""  # noqa: W605
+def extract_timestamp_from_curated_path(string: str):
+    return "extraction_timestamp=" + search_string_for_regex(
+        string=str, regex=EXTRACTION_TIMESTAMP_REGEX
+    )
+
 
 @dataclass
 class DataProductElement:
--- a/containers/daap-python-base/src/var/task/data_platform_paths.py
+++ b/containers/daap-python-base/src/var/task/data_platform_paths.py
@@ -27,10 +27,10 @@ EXTRACTION_TIMESTAMP_FORMAT = "%Y%m%dT%H%M%SZ"
 EXTRACTION_TIMESTAMP_REGEX = re.compile(
     r"^(.*)/(extraction_timestamp=)([0-9TZ]{1,16})/(.*)$"
 )
-DATABASE_NAME = re.compile(r"database_name=([^\/]*))\/")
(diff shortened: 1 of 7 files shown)


containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -108,6 +108,25 @@ def get_account_id() -> str:
...
+def search_string_for_regex(string: str, regex: str) -> str:
+    """Search a string for a regex pattern and return the first result"""
+    database_name_search = re.search(regex, string)
+    if database_name_search:
+        return database_name_search.groups()[0]
+    else:
+        raise ValueError(f"{regex} not found in {string}")
+
+
+def database_name_regex() -> str:
```
MatMoore: Rather than exposing the regex itself, could we have functions for each one, like 

- `extract_timestamp_from_curated_path(string: str)`
- `extract_table_name_from_curated_path(string: str)`
- `extract_database_name_from_curated_path(string: str)`

it should simplify the calling code

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -108,6 +108,25 @@ def get_account_id() -> str:
...
+    return """database_name=([^\/]*)\/"""  # noqa: W605
+
+
+def table_name_regex() -> str:
+    return """table_name=([^\/]*)\/"""  # noqa: W605
+
+
+def extraction_timestamp_regex() -> str:
+    return """(extraction_timestamp=[^\/]*)\/"""  # noqa: W605
```
MatMoore: There is no need to surpress W605 here, this warning is due to python trying to interpret `\/` as an escape sequence before it gets to the regex library.

For regexes you should use raw strings like `r"(extraction_timestamp=[^\/]*)\/"`

MatMoore: This particular one already exists as a precompiled regex at the start of the module, 

```python
EXTRACTION_TIMESTAMP_REGEX = re.compile(
    r"^(.*)/(extraction_timestamp=)([0-9TZ]{1,16})/(.*)$"
)
```

I think this is doing much the same thing except it's anchored at the end of the string - so we should choose one regex or the other.

It's unlikely to make a significant performance difference at the moment, but in general precompiling the regex with re.compile will speed things up if the regex gets called a lot of times (e.g. in deeply nested loops).

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -27,8 +27,10 @@
 EXTRACTION_TIMESTAMP_REGEX = re.compile(
     r"^(.*)/(extraction_timestamp=)([0-9TZ]{1,16})/(.*)$"
 )
+DATABASE_NAME = re.compile(r"database_name=([^\/]*)\/")
```
MatMoore:
```suggestion
DATABASE_NAME_REGEX = re.compile(r"database_name=([^\/]*)\/")
```

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -27,8 +27,10 @@
...
+TABLE_NAME = re.compile(r"table_name=([^\/]*)\/")
```
MatMoore:
```suggestion
TABLE_NAME_REGEX = re.compile(r"table_name=([^\/]*)\/")
```

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -109,6 +111,29 @@ def get_account_id() -> str:
...
+    """Search a string for a regex pattern and return the first result"""
+    search_output = regex.search(string)
+    if search_output:
+        return search_output.groups()[0]
+    else:
+        raise ValueError(f"{regex} not found in {string}")
+
+
+def extract_table_name_from_curated_path(string: str):
+    return search_string_for_regex(string, regex=TABLE_NAME)
```
MatMoore:
```suggestion
    return search_string_for_regex(string, regex=TABLE_NAME_REGEX)
```

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -109,6 +111,29 @@ def get_account_id() -> str:
...
+
+
+def extract_database_name_from_curated_path(string: str):
+    return search_string_for_regex(string, regex=DATABASE_NAME)
```
MatMoore:
```suggestion
    return search_string_for_regex(string, regex=DATABASE_NAME_REGEX)
```

MatMoore (dismissed):
Looks good, just a minor naming suggestion, take it or leave it 🙂

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -109,6 +114,26 @@ def get_account_id() -> str:
     return boto3.client("sts").get_caller_identity()["Account"]
 
 
+def search_string_for_regex(string: str, regex: re.Pattern[str]) -> Optional[str]:
+    """Search a string for a regex pattern and return the first result"""
+    search_output = regex.search(string)
+    return search_output.groups()[0] if search_output else None
```
tom-webber:
```suggestion
def search_string_for_regex(string: str, regex: re.Pattern[str]) -> str:
    """Search a string for a regex pattern and return the first result"""
    search_match = regex.search(string)
    if not search_match:
        raise Exception((f"{regex} not found in {string}"))
    return search_match.group(0)
```
Added an error catching for if the regex isn't matched, as `None` isn't handled by downstream uses.
[re.search only matches the first occurrence of the string](https://docs.python.org/3/library/re.html#functions), returning a match with only that group.

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -19,6 +19,7 @@
 from datetime import datetime
 from typing import NamedTuple
 from uuid import UUID, uuid4
+from typing import Optional
```
tom-webber:
```suggestion
from typing import NamedTuple, Optional
from uuid import UUID, uuid4
```
linting

PriyaBasker23: Thanks

containers/daap-python-base/src/var/task/data_platform_paths.py
PriyaBasker23: For my uses case,  if directory structure does match returning None will be useful hence i didnt want to Raise exception

tom-webber: I changed it here because None isn't handled in downstream use cases. 

If you want to return None, I'd recommend logging `f"{regex} not found in {string}"` instead of `raise`, then handle `None` in the downstream method calls that at the moment are expecting a string.

containers/daap-python-base/src/var/task/data_platform_paths.py
tom-webber: FYI You can hit 'Commit suggestion' to add the change

containers/daap-python-base/src/var/task/data_platform_paths.py
PriyaBasker23: ok

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -27,8 +28,12 @@
 EXTRACTION_TIMESTAMP_REGEX = re.compile(
     r"^(.*)/(extraction_timestamp=)([0-9TZ]{1,16})/(.*)$"
 )
+EXTRACTION_TIMESTAMP_CURATED_REGEX = re.compile(r"(extraction_timestamp=[^\/]*)\/")
+
+DATABASE_NAME_REGEX = re.compile(r"database_name=([^\/]*)\/")
+TABLE_NAME_REGEX = re.compile(r"table_name=([^\/]*)\/")
+
 
-# https://docs.aws.amazon.com/athena/latest/ug/tables-databases-columns-names.html
```
tom-webber, line 31: This feels like a useful comment for context. I would leave it in.

tom-webber (dismissed):
lgtm

tom-webber (approved):

julialawrence (approved):

PR #1649, opened 2023-09-25
Refactor athena load module for inferring schemas

MatMoore:
This refactor introduces an abstraction around the glue metadata to make it easier to pass data around and avoid having to parse the dict structure to access table name and database name elsewhere. I think this will also help with the refactor of `create_curated_table`.

I've also split the infer_glue_schema function into two distinct functions, one for raw CSVs and one for parquet files. Previously the logic dependend on the combination of `file_type` and `table_type` flags, and it was possible to pass a combination other than the 2 use cases we have properly implemented.

In the previous version `infer_glue_schema` handled a lot of different concerns: 
- loading either a single file, or choosing a file from s3
- preparing a sample of data in the case of s3
- the actual conversion of an arrow dataset to a glue schema
- standardising and setting table and column names
- converting column types to strings

In the new version:
- instead of calling infer_glue_schema, calling code calls `infer_glue_schema_for_raw_csv` or `infer_glue_schema_for_parquet`
- there is a GlueSchemaGenerator which operates on file-like objects, so that the s3 logic can be separated out
- there is a csv_sample function, which does the csv sampling on a file like object

### DIFF
--- a/containers/daap-athena-load/CHANGELOG.md
+++ b/containers/daap-athena-load/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Refactor infer_glue_schema
+
 ## [1.1.4] 2023-09-21
 
 ### Changed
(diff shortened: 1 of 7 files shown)


LavMatt (approved):
Looks good, much neater!

PR #1652, opened 2023-09-26
Refactored regex

PriyaBasker23:
Regex pattern is changed to get specific output 
eg:   In the pattern `"curated_data/database_name=data_product/table_name=table/extraction_timestamp=timestamp/file.parquet"`
using` search_match.group(0)` it was returning `database_name=data_product/`
but  changing to `search_match.group(1)`  , now able to extract  `data_product`


tom-webber (dismissed):
LGTM

tom-webber (approved):
Still lgtm

Issue #1310, opened 2023-08-30
further refactor of athena load code


LavMatt:
Some minor improvements have been discussed to be implemented to athena load code, https://github.com/ministryofjustice/data-platform/tree/main/containers/daap-athena-load/src/var/task

## Problems
- Two successive loads could conflict with each other due to sharing the state of the raw data table
- There is a case where we catch ClientError's other than not found. This multiplies the possible scenarios and shouldn't be needed - if we get another kind of ClientError we should let it fail and debug the code
- We are unsure if the scenario in L111 create_curated_athena_table is required. It looks like this covers a real possibility but there is some redundancy in the code between branches.

## Suggested solution
- [x] add random string to end of raw table name to protect against errors if quick successive loads to the same table. (Change in data_platform_paths)
- [ ] replace try except blocks with more intuitive does_exist logic for database and tables (or factor out into a function)
- [ ] separate the code that creates the table if it doesn't exist into one if statement, and the code executes `sql_unload_table_partition` if the partition file already exists into another if statement. `refresh_table_partitions` always happens so it can go outside of the ifs.

## Acceptance criteria
- [ ] Raw table name changes each time the method is run
- [ ] No unnecessary scenarios
- [ ] All remaining scenarios are exercised by the unit tests

MatMoore:
https://github.com/ministryofjustice/data-platform/blob/56c8a30f0d6300d2f95e3e1d5762c7c59985e256/containers/daap-athena-load/src/var/task/create_curated_athena_table.py#L111 <- version with line numbers before my refactor for path names

MatMoore:
I think create_raw_athena_table could be changed to a context manager that handles the clean up as well. If there is any error we want to make sure the table is still removed, or we will accumulate glue tables.

Also wondering if we could restructure to infer_glue_schema as this is a complex function.

MatMoore:
The extra case on L111 is not completely redundant. There are 5 cases:

1. Table exists and partitioned data exist -> do nothing
2. Table exists and partitioned data doesn't exist -> run UNLOAD query
3. Table does not exist and partitioned data exists -> infer schema from parquet, create table
4. Table does not exist and no partitoned data exists at all -> run CREATE TABLE
5. Table does not exists and partitioned data doesn't exist for the current extraction, BUT partitions exist for other extractions

[...]

PR #1653, opened 2023-09-26
Refactor create_curated_athena_table in athena data load lambda

MatMoore:
This is the final part of #1310.

1. Extract CuratedDataQueryBuilder for building athena queries. These queries will change if we switch from parquet to iceberg format.

2. Extract CuratedDataLoader for executing athena queries and inspecting the glue catalog. This part will need to change if we replace athena with dbt.

3. Remove edge case logic: If the table doesn't exist, but partitioned files do, it means something has gone wrong, and we should run reload_data_product. In this case the lambda should just return a 500 error.

4. Since the number of tests has increased, I've grouped them into classes based on the thing they are testing

This leaves `create_curated_athena_tables` as a chain of if statements, so hopefully it's easier to see the different scenarios (previously we had nested ifs and a try/catch). 

This might be simplified further once we implement the ability to upload schemas, at which point we will fetch and validate schemas in athena data load, instead of inferring them from raw csv and/or parquet files.

### DIFF
--- a/containers/daap-athena-load/src/var/task/create_curated_athena_table.py
+++ b/containers/daap-athena-load/src/var/task/create_curated_athena_table.py
@@ -283,7 +283,7 @@ def create_curated_athena_table(
         )
         return
 
-    logger.info(
+    logger.error(
         f"{loader.curated_data_table} does not exist,"
         f" but files exist in {loader.curated_table_path}."
         " Run reload_data_product to recreate the data product before continuing."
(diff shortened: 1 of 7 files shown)


containers/daap-athena-load/src/var/task/create_curated_athena_table.py
```diff
@@ -74,191 +257,38 @@ def create_curated_athena_table(
...
-def refresh_table_partitions(
-    database_name: str, table_name: str, athena_client
-) -> None:
-    """
-    refreshes partitions following an update to a table
-    """
-    athena_client.start_query_execution(
-        QueryString=f"MSCK REPAIR TABLE {database_name}.{table_name}",
-        WorkGroup="data_product_workgroup",
+    logger.info(
```
LavMatt: should this be logged as an error?
```suggestion
    logger.error(

LavMatt (commented):
LGTM. Just one minor comment/suggestion

LavMatt (approved):

murdo-moj (approved):

PR #1660, opened 2023-09-26
Ensure raw athena table is deleted on error

MatMoore:
Since we now use a unique name for this table, we should be careful to always delete it, otherwise we will have resources accumulate in AWS.

(Part of #1310)


murdo-moj:
have you tested this process still works plugged into actual athena?

MatMoore:
Actually, the current version is broken, and the fix depends on a base image update that I put in the other PR.

I've updated the commit now so that this PR works in isolation.

I'm testing it against real athena by deleting the curated data in development and then running the request against the local container.

MatMoore:
I've also just deleted all the raw tables that were hanging around in the development sandbox. There may be more created before this fix goes in though.

murdo-moj (approved):

PR #1688, opened 2023-09-27
Dp 1478 unit test for daap resync unprocessed files

PriyaBasker23:
Unit test has been implemented for the resynchronisation of unprocessed files. 

This Lambda function is designed to compare the contents of the 'raw_data' and 'Curated_data' buckets, identifying and filtering files that are present in the 'raw_data' bucket but missing in the 'curated_data' bucket. These identified files are then queued for reprocessing.


containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
```diff
@@ -16,77 +24,62 @@
...
 
 def handler(event, context):
     data_product_to_recreate = event.get("data_product", "")
-    raw_prefix = f"raw_data/{data_product_to_recreate}"
-    curated_prefix = f"curated_data/database_name={data_product_to_recreate}"
+
+    data_product = DataProductConfig(
+        name=data_product_to_recreate,
+        raw_data_bucket=raw_data_bucket,
+        curated_data_bucket=raw_data_bucket,
```
mitchdawson1982: Is this meant to be `curated_data_bucket=curated_data_bucket` ?

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
```diff
@@ -16,77 +24,62 @@
...
         data_product_prefix=raw_prefix,
     )
 
-    # Find extraction timestamps in the raw area, and not in the curated area
-    paginator = s3.get_paginator("list_objects_v2")
-    curated_pages = paginator.paginate(
-        Bucket=curated_data_bucket, Prefix=curated_prefix
+    # get data product has associated curated data
+    curated_pages = get_data_product_pages(
+        ucket=curated_data_bucket, data_product_prefix=curated_prefix
```
mitchdawson1982: I think this should be `bucket=curated_data_bucket`

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
```diff
@@ -16,77 +24,62 @@
...
 
 
 def handler(event, context):
     data_product_to_recreate = event.get("data_product", "")
-    raw_prefix = f"raw_data/{data_product_to_recreate}"
-    curated_prefix = f"curated_data/database_name={data_product_to_recreate}"
+
+    data_product = DataProductConfig(
+        name=data_product_to_recreate,
+        raw_data_bucket=raw_data_bucket,
```
MatMoore, line 34: you can leave these bucket params out if you want, I made it default to the environment variables

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
```diff
@@ -16,77 +24,62 @@
...
+        raw_table_timestamps, curated_table_timestamps, raw_pages
+    )
     logger.info(f"raw keys to resync: {raw_keys_to_resync}")
 
     # Feed unprocessed data files through the load process again.
     # If there are over 1000 files, the lambda will get jammed up
     aws_lambda = boto3.client("lambda")
     for key in raw_keys_to_resync:
-        payload = f'{{"detail":{{"bucket":{{"name":"{raw_data_bucket}"}}, "object":{{"key":"{key}"}}}}}}'
+        payload = f'{{"detail":{{"bucket":{{"name":"{raw_data_bucket}"}},\
```
MatMoore: [suggestion] this line hasn't changed but it might be a bit cleaner as a json.dumps call of a nested dict rather than a string interpolation. That way you don't have to worry about whether the brackets match - if not the code won't run.

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
```diff
@@ -107,36 +103,54 @@ def get_data_product_pages(
...
-
-def search_string_for_regex(string: str, regex: str) -> str:
-    """Search a string for a regex pattern and return the first result"""
-    database_name_search = re.search(regex, string)
-    if database_name_search:
-        return database_name_search.groups()[0]
-    else:
-        raise ValueError(f"{regex} not found in {string}")
+    filtered_pages = pages.search("Contents[?Size > `0`][]")
+    result_set = set("/".join(item["Key"].split("/")[1:-1]) for item in filtered_pages)
```
MatMoore, line 106: Just to check my understanding: did you decide not to use RawDataExtraction.parse_extraction_timestamp here because you need the whole "data_product/table/extraction_timestamp=timestamp" path here?

mitchdawson1982 (commented):
I wasn't sure if the curated_data_bucket argument should also be curated_data_bucket as opposed to raw_data_bucket.

```
    data_product = DataProductConfig(
        name=data_product_to_recreate,
        raw_data_bucket=raw_data_bucket,
        curated_data_bucket=**raw_data_bucket**,
```

MatMoore (dismissed):
This all looks good to me, thanks for picking this one up.

It bothers me a little that the data product object doesn't have methods to parse/generate the `f"{data_product}/{table}/extraction_timestamp={extraction_timestamp}"` expression, so we still have some path wrangling in both `get_unique_extraction_timestamps` and `get_curated_unique_extraction_timestamps`, but I think at the moment it's better to close off this ticket than go back and change the path library again. Maybe it's something we can consider again when we implement @murdo-moj's suggested path structure.

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
```diff
@@ -16,77 +24,62 @@
...
         aws_lambda.invoke(
             FunctionName=athena_load_lambda, InvocationType="Event", Payload=payload
         )
 
     logger.info(
-        f"data product {data_product_to_recreate} resynced with {len(raw_keys_to_resync)} files"
+        f"data product {data_product_to_recreate} resynced"
+        + "with {len(raw_keys_to_resync)} files"
     )
     logger.info(str(raw_keys_to_resync))
```
mitchdawson1982, line 80: Is this required ? Keys are logged on line 68.

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
PriyaBasker23: yes . Well spotted. I couldnt catch in unit test because both raw and curated bucket are same

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
PriyaBasker23: This is necessary to get the raw_prefix and curated_prefix isnt it

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
PriyaBasker23: yes

PriyaBasker23: so i can filter out the ones in curated

containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
PriyaBasker23: Will be useful for debugging

MatMoore (approved):

PR #1701, opened 2023-09-28
Add workflow to create a label to correspond to an epic.

julialawrence:
This workflow will generate a corresponding label when a new epic is created so the new stories in the epic can be tagged with it.


ymao2 (approved):

Gary-H9 (approved):
LGTM

PR #1760, opened 2023-10-02
Dp 1755 Reformat directory structure

murdo-moj:
All data directories are now of the form `{area}/{data-product}/{version}/{table}/load_timestamp={load_timestamp}/{filename}.{extension}`

metadata is of the form `{data-product}/{version}/metadata.json`

There's a `get_latest_version` function which defaults to v1.0 to populate the filepaths. This function calls s3, so loads more mocking is needed in the tests to stub the call to the AWS API. This bit felt messy, I think the tests could be improved, but this works for now.


containers/daap-python-base/CHANGELOG.md
```diff
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [3.0.0] - 2023-10-2
+
+### Added
+
+- landing_prefix method
```
murdo-moj:
```suggestion
- landing_data_prefix method
- get_latest_version function
```

containers/daap-python-base/tests/unit/infer_glue_schema_test.py
```diff
@@ -47,12 +47,7 @@ def test_csv_sample(test_input, expected, sample_size_in_bytes, logger):
     assert output == expected
```
murdo-moj, line 49: the bucket in this test is made in conftest now

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -240,30 +260,40 @@ class DataProductConfig:
     curated_data_bucket: str = field(default_factory=get_curated_data_bucket)
     metadata_bucket: str = field(default_factory=get_metadata_bucket)
 
+    def __post_init__(self):
+        self.latest_version = get_latest_version(self.name)
```
murdo-moj, line 264: the post init is a bit awkward in a dataclass but we only want 1 s3 call to find the latest version, and this was the easiest implementation I could think of

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -240,30 +260,40 @@ class DataProductConfig:
...
         return BucketPath(
             bucket=self.raw_data_bucket,
-            key=os.path.join("raw_data", self.name) + "/",
+            key=os.path.join("raw", self.name, self.latest_version) + "/",
+        )
+
+    @property
+    def landing_data_prefix(self):
+        """
+        The path to the raw data in s3 excluding the element name,
```
murdo-moj:
```suggestion
        The path to the landing data in s3 excluding the element name,
```

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -4,11 +4,10 @@
...
 Example for data product name "data_product", table name "table":
 
-- Raw data is stored at: raw_data/data_product/table/extraction_timestamp=timestamp/3d95ff89-...-53742d0a6a64
-- Curated data is stored at:
-  curated_data/database_name=data_product/table_name=table/extraction_timestamp=timestamp/file.parquet
+- Raw data is stored at: raw/{data-product}/{version}/{table}/load_timestamp={load_timestamp}/data.ext
+- Curated data is stored at: curated/{data-product}/{version}/{table}/load_timestamp={load_timestamp}/data.ext
 - Athena table name for curated data is: data_product.table
-- Athena table name for raw data has the format: data_products_raw.table_8ae7f9ee-7888-4b6a-b01f-8b8b0a537ad2_raw
+- Athena table name for raw data has the format: data_products_raw.table_data_3d95ff89-b063-484d-b510-53742d0a6a64_raw
```
LavMatt: isn't `-` now replaced with `_` for the unique id part of the raw table?

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -131,7 +130,28 @@ def extract_database_name_from_curated_path(string: str):
...
 def extract_timestamp_from_curated_path(string: str):
-    return search_string_for_regex(string, regex=EXTRACTION_TIMESTAMP_CURATED_REGEX)
+    return search_string_for_regex(string, regex=LOAD_TIMESTAMP_CURATED_REGEX)
+
+
+def get_latest_version(data_product_name: str):
+    """Get the latest version of a data product, else return v1.0"""
+    metadata_bucket = get_metadata_bucket()
+    metadata_versions = s3.list_objects_v2(
+        Bucket=metadata_bucket, Prefix=data_product_name
```
LavMatt: would the prefix simply be `data_product_name`? Is the metadata bucket going to just have data product metadata or will contain table schema as well? Currently the path is `metadata/{data_product_name}` but appreciate we only have the one bucket

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -280,7 +310,7 @@ def metadata_path(self):
         key = os.path.join(
             "metadata",
```
LavMatt: This path now doesn't quite align with the prefix in `get_latest_version` where there is no `metadata`

LavMatt (commented):
couple of minor comments. Main one being the metadata paths now don't quite align within some of the methods

containers/daap-python-base/src/var/task/data_platform_paths.py
murdo-moj, line 10:
```suggestion
- Athena table name for raw data has the format: data_products_raw.table_data_3d95ff89_b063_484d_b510_53742d0a6a64_raw
```

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -23,12 +22,12 @@
...
+
 RAW_DATABASE_NAME = "data_products_raw"
-EXTRACTION_TIMESTAMP_FORMAT = "%Y%m%dT%H%M%SZ"
-EXTRACTION_TIMESTAMP_REGEX = re.compile(
-    r"^(.*)/(extraction_timestamp=)([0-9TZ]{1,16})/(.*)$"
-)
-EXTRACTION_TIMESTAMP_CURATED_REGEX = re.compile(r"extraction_timestamp=([^/]+)")
+LOAD_TIMESTAMP_FORMAT = "%Y%m%dT%H%M%SZ"
+LOAD_TIMESTAMP_REGEX = re.compile(r"^(.*)/(load_timestamp=)([0-9TZ]{1,16})/(.*)$")
+LOAD_TIMESTAMP_CURATED_REGEX = re.compile(r"load_timestamp=([^/]+)")
```
MatMoore, line 30: I might have forgotten something here, but can we combine these two regexes? The first one has some extra capture groups and is a bit more strict with the format, but otherwise I think they're doing the same thing.

containers/daap-python-base/src/var/task/data_platform_paths.py
MatMoore: tbh I think even if we have schema as well it could use the same structure

I was imagining something like this

- {data_product_name}/v1.0/metadata.json
- {data_product_name}/v1.0/{table}/schema.json

containers/daap-python-base/src/var/task/data_platform_paths.py
MatMoore: the only other way I can think of is to completely separate the lookup from this class.

So instead of calling DataProductConfig directly from the lambdas we would call some helper function or static method.

Happy to go with this for now though

containers/daap-python-base/src/var/task/data_platform_paths.py
MatMoore: good spot, might be worth extracting this to a top level function as well if we want to ensure this is always consistent

containers/daap-python-base/tests/unit/data_platform_paths_test.py
```diff
@@ -88,152 +90,165 @@ def test_data_product_config_metadata_spec_prefix():
 
 
 def test_data_product_element_config(monkeypatch):
-    monkeypatch.setenv("RAW_DATA_BUCKET", "bucket")
-    monkeypatch.setenv("CURATED_DATA_BUCKET", "bucket")
-    monkeypatch.setenv("METADATA_BUCKET", "bucket")
-    monkeypatch.setenv("LANDING_ZONE_BUCKET", "bucket")
+    with patch("data_platform_paths.get_latest_version", lambda _: "v1.0"):
+        monkeypatch.setenv("RAW_DATA_BUCKET", "bucket")
```
MatMoore: if you want to get rid of these annoying setenv's everywhere, maybe we could put them in a fixture in conftest with autouse=True?

MatMoore (dismissed):
Looks good to me 👍🏻

MatMoore:
Regarding the region issue -

Elsewhere I've done

```
    s3_client.create_bucket(
        Bucket=bucket_name,
        CreateBucketConfiguration={"LocationConstraint": region_name},
    )
```

that should work with any region, but I think it lets you leave off the LocationConstraint if you use the default us-east-1

containers/daap-python-base/src/var/task/data_platform_paths.py
LavMatt: and if it is just the data product name it needs to end with `/` otherwise it'll pick up objects that start with that data product name
```suggestion
        Bucket=metadata_bucket, Prefix=data_product_name + "/"

murdo-moj:
I tried to default all the regions to us-east-1 but some tests didn't like it, so I put it back to the mix it is now

MatMoore (approved):

Issue #1656, opened 2023-09-26
API End point changes - Data Product operations - Register a data product (POST)


PriyaBasker23:
## User story
As a data producer
I want to register a data product (including metadata)
So that it becomes available in the data catalog
And I can start to publish my data

Note: Registering the data product end point already exists. The scope of the ticket would be to update the api end point based on the new design, and add in missing checks and responses.

## API Design
https://dsdmoj.atlassian.net/wiki/spaces/DataPlatform/pages/4511991882/Initial+APIs#Data-Product-operations

POST data-product/register

accepts application/json, representing the metadata for the data product.

### Validation
- Schema must be valid according to https://github.com/ministryofjustice/modernisation-platform-environments/blob/main/terraform/environments/data-platform/data-product-metadata-json-schema/v1.0.0/moj_data_product_metadata_spec.json
- Data product name must contain only letters, numbers, and underscores (TBC)
- Data product name length max limit TBC
- Data product with this name doesn't already exist

### Success response
There is no response body so we should return status code 201 on success.

### Error responses
- use the standardised response functions for responses in the base image
- Return 400 response on invalid schema
- Return 400 response on invalid data_product_name
- [Return 409](https://stackoverflow.com/a/70371989/10478115) if the data product already exists

## Lambda behaviour
This already exists as create_metadata.

Double check the functionalities:

- Validate schema
- Extract name from schema
- Validate name
- Check if data product already exists
- Store the metadata in the correct path for v1: `DataProductConfig(name).metadata_path()`
- Create a folder structure for the data product (e.g. raw_data, curated_data paths)

## Definition of done
- [x] Checked/updated the new lambda
- [x] Added swagger docs
- [x] Deployed the new endpoint for API gateway
- [ ] ~~Added to the smoke test (if possible)~~
- [ ] We are able to call the endpoint to create a new data product

tom-webber:
[Deploying new endpoint PR](https://github.com/ministryofjustice/modernisation-platform-environments/pull/3515)

tom-webber:
[PR for updating the lambda](https://github.com/ministryofjustice/data-platform/pull/1763)

tom-webber:
[...]

PR #1763, opened 2023-10-02
update create_metadata lambda

### LINKED ISSUE #1657: API End point - Data Product operations - Update a data product (PUT)
## User story

As a data producer
I want to update the metadata for an existing data product
So that I can change information in the catalogue
And consumers can be notified

The scope of the ticket would be to create an api end point for updating the existing data product in line with the design below. Note: this endpoint updates details about the data product. It's not to be used for data nor data schema.

### Value / Purpose

If we have this functionality we can trigger when its requested by user or delete test tables created. We can also automate this process

### Hypothesis

By having this option we will be able allow users to update the data product information when needed

## API design

The end point should be PUT /data-product/{data-product-name}

It accepts application/json, representing the metadata for the data product.

### Validation

- The metadata must be valid according to https://github.com/ministryofjustice/modernisation-platform-environments/blob/main/terraform/environments/data-platform/data-product-metadata-json-schema/v1.0.0/moj_data_product_metadata_spec.json
- The data product must exist already (i.e. there is a metadata file for a previous version)

### Success response

- The 200 response should be a json containing the new version number. E.g. `{"version_number": "1.1"}`

### Error responses

- Return a 400 response if the schema is not valid
- Return a 404 response if no existing data product
- Return a 400 response if fields have changed that aren't allowed to be changed

## Updatable vs not updatable fields

Only specific metadata should be updatable. We haven't defined specific rules, but for the scope of this ticket any fields that wont cause issues can updated, e.g. `description`.

Note that the data product cannot be renamed, a new one must be created instead.

## New lambda functionality
- Validate the metadata against the spec (same as create-metadata lambda)
- Fetch the most recent version of the metadata
- Increment the version number, e.g. 1.0 -> 1.1
- Compare the metadata to the previous one to see what changed
- Store the metadata in the correct path for the new version

## Checklist

- [ ] add code to terraform module for api infrastructure 
- [ ] create a lambda for updating  the data product
- [ ] DataProductConfig(name).metadata_path() will need to be updated to accept a version number
- [ ] Update GitHub docs with existing versioning logic (migrate [confluence doc](https://dsdmoj.atlassian.net/wiki/spaces/DataPlatform/pages/4479156541/Data+product+Schema+Versioning) to monorepo)
- [ ] Specify which fields are editable, and implement this being respected

## Definition of Done

- [ ] Able to update a data product to a new version
- [ ] ~~smoke test implemented (if possible)~~
- [ ] swagger documentation for the new endpoint

tom-webber:
- update lambda response codes to match [aws example](https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/python/example_code/lambda/lambda_handler_rest.py) 

update #1656

### DIFF
--- a/containers/daap-create-metadata/src/var/task/create_metadata.py
+++ b/containers/daap-create-metadata/src/var/task/create_metadata.py
@@ -88,6 +88,14 @@ def handler(event, context):
 
     logger.add_extras({"data_product_name": data_product_name})
 
+    if http_method == "POST":
+        pass
+    else:
+        error = f"Sorry, {http_method} isn't allowed."
+        response_code = HTTPStatus.METHOD_NOT_ALLOWED.value
+        logger.error(error)
+        return generate_response(response_code, event, error=error)
+
     data_product_metadata = DataProductMetadata(data_product_name, logger)
 
     if not data_product_metadata.metadata_exists:
@@ -107,13 +115,6 @@ def handler(event, context):
     else:
         response_code = HTTPStatus.CONFLICT.value
         error = f"Data Product {data_product_name} already has a version 1 registered metadata."  # noqa E501
-        logger.error(error)
-
-    if http_method == "POST":
-        pass
-    else:
-        error = f"Sorry, {http_method} isn't allowed."
-        response_code = HTTPStatus.METHOD_NOT_ALLOWED.value
         logger.error(error)
 
     response = generate_response(
--- a/containers/daap-create-metadata/src/var/task/create_metadata.py
+++ b/containers/daap-create-metadata/src/var/task/create_metadata.py
@@ -11,7 +11,6 @@ class HTTPStatus(Enum):
     SUCCESS = 200
     CREATED = 201
     BAD_REQUEST = 400
-    NOT_FOUND = 404
     METHOD_NOT_ALLOWED = 405
     CONFLICT = 409
 
--- a/containers/daap-create-metadata/src/var/task/create_metadata.py
+++ b/containers/daap-create-metadata/src/var/task/create_metadata.py
@@ -1,6 +1,7 @@
 import json
 import os
 from enum import Enum
+from typing import Optional
 
 from data_platform_logging import DataPlatformLogger
 from data_product_metadata import DataProductMetadata
@@ -43,9 +44,9 @@ def handler(event, context):
     def generate_response(
         response_code: int,
         event: dict,
-        response_message: str = None,
-        data_product_name: str = None,
-        error: str = None,
+        response_message: Optional[str] = None,
+        data_product_name: Optional[str] = None,
+        error: Optional[str] = None,
     ) -> dict:
         """
         Generate a response to return to API Gateway that contains the response code,
@@ -53,13 +54,13 @@ def handler(event, context):
         """
         response_body = {"input": event}
 
-        if response_message:
-            response_body.update({"message": response_message})
-        elif error:
+        if response_message is not None:
+            response_body.update({"message": response_message})  # type: ignore[dict-item]
+        elif error is not None:
             response_body.update({"error": {"message": error}})
 
         if data_product_name is not None:
-            response_body.update({"data_product_name": data_product_name})
+            response_body.update({"data_product_name": data_product_name})  # type: ignore[dict-item]
 
         response = {
             "statusCode": response_code,
--- a/containers/daap-create-metadata/src/var/task/create_metadata.py
+++ b/containers/daap-create-metadata/src/var/task/create_metadata.py
@@ -1,8 +1,18 @@
 import json
 import os
+from enum import Enum
 
 from data_platform_logging import DataPlatformLogger
 from data_product_metadata import DataProductMetadata
+
+
+class HTTPStatus(Enum):
+    SUCCESS = 200
+    CREATED = 201
+    BAD_REQUEST = 400
+    NOT_FOUND = 404
+    METHOD_NOT_ALLOWED = 405
+    CONFLICT = 409
 
 
 def handler(event, context):
@@ -66,12 +76,12 @@ def handler(event, context):
     data_product_name = None
 
     http_method = event.get("httpMethod")
-    body = event.get("body")
+    body = json.loads(event.get("body"))
 
     try:
         data_product_name = body["metadata"]["name"]
     except KeyError:
-        response_code = 400
+        response_code = HTTPStatus.BAD_REQUEST.value
         error = "Data product name is missing, it must be specified in the metadata against the 'name' key"  # noqa E501
         logger.error(error)
         return generate_response(response_code, event, error=error)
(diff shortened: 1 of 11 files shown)


tom-webber:
secret linter is picking up the AWS example key

containers/daap-docs/src/var/task/swagger.json
```diff
@@ -170,29 +170,59 @@
...
+          "201": {
+            "description": "successfully created",
             "content": {
               "application/json": {
                 "schema": {
                   "type": "object",
                   "example": {
-                    "data_product_name": "example_data_product",
-                    "message": "Data product metadata has been created."
+                    "message": "Data Product {data_product_name} was successfully created."
```
MatMoore: Is there a use case for the `message` field in the success response? Is it something we want to return across all endpoints or specific to this one?

It seems superfluous to me because the meaning of 201 itself means that the resource has been successfully created.

If in doubt, I would lean towards YAGNI (you ain't gonna need it), since once something is in our API it's hard to remove, but we can always add stuff in later.

containers/daap-create-metadata/src/var/task/create_metadata.py
```diff
@@ -1,74 +1,112 @@
 import json
 import os
+from enum import Enum
 
 from data_platform_logging import DataPlatformLogger
 from data_product_metadata import DataProductMetadata
 
 
+class HTTPStatus(Enum):
```
MatMoore: I recommend using `IntEnum` here instead of just `Enum`. https://docs.python.org/3/library/enum.html#enum.IntEnum

That way you can type `format_reponse` to accept `status_code: IntEnum` (which will forbid nonsense codes like -1) and you don't have to add `.value` to everything, you can just call

```python
format_response(HTTPStatus.METHOD_NOT_ALLOWED, event, response_body)
```

containers/daap-create-metadata/src/var/task/create_metadata.py
```diff
@@ -1,74 +1,112 @@
...
 
+    def format_response(response_code: int, event: dict, body_dict: dict) -> dict:
+        """
+        Generate a response to return to API Gateway that contains the initial event,
+        response code, and contents of the response body
+        (i.e. body['message'] or body['error']['message'])
+        """
+        response_body = {"input": event}
+        response_body.update(body_dict)
+        response = {
```
MatMoore, line 66: As discussed earlier, we do want a function like this in the base image that handles the basic formatting.

So I think there are two options:
1. Make that change prior to merging this PR
2. Leave it in this lambda for now, make sure all reviewers agree with the signature and response format, then migrate the changes to the base image after the PR is merged.

(I'm happy to pair on either)

Comments below are cross posted from original discussion https://mojdt.slack.com/archives/C03QZ776JVA/p1696591089567449?thread_ts=1696584870.001379&cid=C03QZ776JVA

@tom-webber 
> My aim was to have a single function for all of my 'response statuses', rather than a function per code
I also wanted to ensure we were passing errors as errors, rather than messages (and optionally in place of messages).
The data_product_name field can probably be removed from everywhere.

@murdo-moj 
> I'm very unprecious about the way we standardise responses. It might be easier to implelement later down the line when we have an idea of the variability of what we want to return. I do like the idea of including inputs in the return for ease of debugging.

@MatMoore 
> I was talking to 
[@Matt Laverty](https://mojdt.slack.com/team/U01FBCH07B8)
 about this earlier as well. I was thinking we don't have much common structure in our API responses yet, so a function where we can just pass in a dict or an error works is probably ok for now. The main thing I was concerned about originally was just reporting errors in a consistent way and getting rid of some boilerplate :slightly_smiling_face:

@tom-webber 
> Gotcha. So rather than providing error or message as bespoke body.update() options, simply pass in a body_content dict and update with that?
That makes sense to me.
def format_response(response_code: int, event: dict, body_content: dict) -> dict:
    response_body = {"input": event}
    response_body.update(body_content)
    response = {
        "statusCode": response_code,
        "headers": {"Content-Type": "application/json"},
        "body": json.dumps(response_body),
    }
    return response

MatMoore: Just to confirm, I'm in favour of getting rid of `response_status_200` `response_status_404` and using the enum argument to indicate status.

Regarding the response format, there are a few changes this will make to the output. Since we are not consistently using the existing helpers anyway, I think it's ok to make breaking changes now, I just want to highlight that this is what we're doing.

1. `response_status_200` wraps everything in a `message`, whereas `format_response` has the caller decide the body structure. IMO if responses don't have any metadata like `input` (see below) then `message` is redundant and the `format_response` that does not wrap the body in another field.

    However, if we are keep the `input` field then we could assign the response_body in `message` to clearly delineate the two (but either approach is fine imo).

2. `format_response` adds the input field to echo the request event in the response. I think this is not really that useful to the client at the time of making the request, but is very useful for debugging afterwards. **My suggestion is to change the function to just log the event and response** instead of returning the event as its own field. I think this would be extremely useful information to see in the log bucket, particularly if we expose those logs via the API.

3. The existing `response_status_4xx` functions handle wrapping errors in a special `error` object, but with `response` we need to construct a dict when we call it. Have you considered adding a separate function for this e.g. `format_error_response(status_code: HTTPStatusCode, event: dict, message: str) -> dict`

containers/daap-create-metadata/src/var/task/create_metadata.py
```diff
@@ -1,74 +1,112 @@
...
-            "statusCode": 400,
-            "headers": {"Content-Type": "application/json"},
-            "body": json.dumps(
-                {
-                    "data_product_name": data_product_name,
-                    "message": "Your data product already has a version 1 registered metadata.",
-                }
-            ),
-        }
+        data_product_metadata.validate(event_body["metadata"])
```
MatMoore, line 102: Very minor suggestion: it might be more readable to invert the condition of this if, and do an early return if it does exist. Then all the special cases are at the top and the main logic as the end of the function.

containers/daap-create-metadata/src/var/task/create_metadata.py
```diff
@@ -1,74 +1,112 @@
...
-                    "data_product_name": data_product_name,
-                    "message": "Data product metadata has been created.",
-                }
-            ),
-        }
+        if data_product_metadata.valid_metadata:
+            data_product_metadata.write_json_to_s3()
+            response_code = HTTPStatus.CREATED.value
+            message = f"Data Product {data_product_name} was successfully created."
+            response_body = {"message": message}
```
MatMoore: See below note about 201 responses. Normally in a REST API you would have 201 with an empty body or 200 with the body being the resource itself.

containers/daap-create-metadata/src/var/task/create_metadata.py
```diff
@@ -1,74 +1,112 @@
...
+    """
+
     logger = DataPlatformLogger(
         extra={
             "image_version": os.getenv("VERSION", "unknown"),
             "base_image_version": os.getenv("BASE_VERSION", "unknown"),
         }
     )
 
+    def format_response(response_code: int, event: dict, body_dict: dict) -> dict:
```
MatMoore: Minor suggestion: Body dict could be marked optional here and the `None` case handled.

MatMoore (changes requested):

containers/daap-create-metadata/src/var/task/create_metadata.py
tom-webber: Great suggestion. Looking into it I've just found that the http core package has [HTTPStatus already implemented as an IntEnum](https://docs.python.org/3/library/http.html). I'll just use that.

containers/daap-create-metadata/src/var/task/create_metadata.py
tom-webber: It may be a style choice, but I favour preserving the consistency of 

`try/if` _happy path_ 
`except/else` _unhappy path_

throughout the handler

containers/daap-create-metadata/src/var/task/create_metadata.py
tom-webber: - 1 / 2: I have no preference. And I agree that it may not be useful to the client at the time to include the `input` event. Using the logger in the base image may be the best bet.
- 3: This is a good suggestion. I'll implement it here, then propagate to the base image later.

containers/daap-create-metadata/src/var/task/create_metadata.py
```diff
@@ -1,74 +1,119 @@
...
+            "body": json.dumps(response_body),
+        }
+
+        logger.debug(
+            f"code: {response_code}, response body: {body_dict}, input: {event}"
+        )
+
+        return response
+
+    def format_error_response(response_code: int, event: dict, message: dict) -> dict:
```
MatMoore: the type is still `int` on this one, should be `HTTPStatus` as well I think

MatMoore (dismissed):

MatMoore (approved):

jacobwoffenden (approved):

PR #1774, opened 2023-10-03
✨ Added landing to raw lambda code

murdo-moj:
* Added landing to raw lambda
* This lambda is triggered upon data arriving to landing and doesn't need to return anything


containers/daap-landing-to-raw/src/var/task/landing_to_raw.py
```diff
@@ -0,0 +1,34 @@
+import os
+
+import boto3
+from data_platform_logging import DataPlatformLogger, s3_security_opts
+from data_platform_paths import get_raw_data_bucket
+
+logger = DataPlatformLogger(
+    extra={
```
MatMoore: do we still need this `extra` argument? I thought we'd moved this into the logging module

MatMoore (dismissed):
👍🏻

containers/daap-landing-to-raw/src/var/task/landing_to_raw.py
murdo-moj: I removed it, thanks

LavMatt: Do still need the extra args i think, the `security_opts` were removed

LavMatt: Yeah it doesn't set the base image or lambda container image versions within the logging module. It does get name of the lambda from an environment variable but also if you wanted to add in data_product_name and table_name you would do it in extra too

murdo-moj: ok reverted

MatMoore (approved):
Reapproved!

containers/daap-landing-to-raw/src/var/task/landing_to_raw.py
```diff
@@ -0,0 +1,36 @@
...
+    copy_source = {"Bucket": bucket_name, "Key": file_key}
+    destination_key = file_key.replace("landing/", "raw/")
+
+    config = RawDataExtraction.parse_from_uri("s3://bucket/" + file_key)
+    logger = DataPlatformLogger(
+        extra={
+            "image_version": os.getenv("VERSION", "unknown"),
+            "base_image_version": os.getenv("BASE_VERSION", "unknown"),
+            "data_product": config.element.data_product.name,
+            "table": config.element.curated_data_table.name,
```
LavMatt, line 23: Could these be consistently named? like https://github.com/ministryofjustice/data-platform/blob/main/containers/daap-athena-load/src/var/task/athena_load_handler.py#L33-L34
We should probably think of a better way to do this. Maybe within the base image logging module when passed a `RawDaraExtraction` object, but think that's not one for this PR

PR #1781, opened 2023-10-04
Automate containers.yml generation

murdo-moj:
- Incorporated path-filter file generation into `containers.yml` workflow


MatMoore (approved):

julialawrence (approved):
LGTM

PR #1785, opened 2023-10-04
DPL-1194 additions to path module for create schema

LavMatt:
So far just a slight refactor and addition to the path module, adding in schema path.


containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -36,6 +37,29 @@
...
+
+    and
+
+    data product schema - relates to data describing a table within a data product
+    """
+    data_product_metadata: str = "metadata"
+    data_product_schema: str = "schema"
+
+    @classmethod
+    def _missing_(cls, value):
```
MatMoore, line 55: Is this `__missing__` just to customise the error message? I get a ValueError by default if I call it with an undefined value, just a slightly less helpful one

```python
>>> MetadataTypes('xyz')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/matt.moore/.pyenv/versions/3.11.5/lib/python3.11/enum.py", line 712, in __call__
    return cls.__new__(cls, value)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/matt.moore/.pyenv/versions/3.11.5/lib/python3.11/enum.py", line 1128, in __new__
    raise ve_exc
ValueError: 'xyz' is not a valid MetadataTypes
```

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -314,33 +338,46 @@ def metadata_path(self):
...
+            self.name,
+            self.latest_version,
+            table_name,
+            "schema.json",
+        )
+        return BucketPath(bucket=self.metadata_bucket, key=key)
+
     @staticmethod
-    def metadata_spec_prefix(bucket_name: str | None = None) -> BucketPath:
+    def specification_prefix(metadata_type: str, bucket_name: str | None = None) -> BucketPath:
```
MatMoore: I don't think you need to accept a string here - you could type it to `MetadataType` and force the caller to pass in the correct value, instead of doing a lookup in the method. That way the type checker can catch the errors instead of it only being caught at runtime.

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -314,33 +338,46 @@ def metadata_path(self):
...
             bucket_name or get_metadata_bucket(),
             os.path.join(
-                "data_product_metadata_spec",
+                f"{spec_type.name}_spec",
             ),
         )
 
     @staticmethod
-    def metadata_spec_path(version: str, bucket_name: str | None = None) -> BucketPath:
+    def specification_path(metadata_type: str, version: str, bucket_name: str | None = None) -> BucketPath:
```
MatMoore: I like the idea of this handling both the JSON schema for the glue schemas and the JSON schema for the metadata. 👍🏻 

Maybe rather than making this a static method we could move it to a top level function though, I feel like it's quite separate from what the dataclass is doing.

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -314,33 +338,46 @@ def metadata_path(self):
         )
         return BucketPath(bucket=self.metadata_bucket, key=key)
 
+    def schema_path(self, table_name: str) -> BucketPath:
```
MatMoore, line 374: 👍🏻 this is exactly what I need

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -36,6 +37,29 @@
 MAX_IDENTIFIER_LENGTH = 255
 
 
+class MetadataTypes(Enum):
```
MatMoore: Suggestion: Maybe something like `SpecificationType` or `JsonSchemaName` would be a more helpful name here - I get it now that I know what it's doing, but it's so easy to confuse the specification paths with the actual schema or metadata paths.

LavMatt: yeah, am trying to avoid using `metadata` where possible, so i like it

containers/daap-python-base/src/var/task/data_platform_paths.py
LavMatt: yeah it is

containers/daap-python-base/tests/unit/conftest.py
```diff
@@ -76,12 +76,15 @@ def raw_data_table(data_product_element):
 
 
 @pytest.fixture
-def data_product_element(s3_client, monkeypatch):
-    monkeypatch.setenv("METADATA_BUCKET", "test")
+def data_product_element(region_name, s3_client, monkeypatch):
+    # monkeypatch.setenv("METADATA_BUCKET", "metadata")
```
MatMoore, line 80: should this be commented out?

MatMoore (dismissed):

containers/daap-python-base/tests/unit/conftest.py
LavMatt: i think it's covered by `set_env_vars` fixture, can remove as tests run and pass without it

MatMoore (approved):

PR #1787, opened 2023-10-04
DPL-1655 Add schema validation to landing bucket -> raw lambda

MatMoore:
## Description of changes
https://github.com/ministryofjustice/data-platform/issues/1655

1. Infer the schema using the infer schema functionality
2. Check against the registered schema
3. If it doesn't match, move to fail path in the data bucket
4. If it does match it will get moved to the raw location (already implemented)

Notification via slack or email would be implemented later.

## Data type handling
We decided to make the schema checks very lenient because there are lots of cases where the registered schema could be a different type to what we can infer from the data. E.g. the data is inferred as integer, but the schema says bigint.

## Related
- The types Athena supports: https://docs.aws.amazon.com/athena/latest/ug/data-types.html
- The types mojap_metadata handles: https://github.com/moj-analytical-services/mojap-metadata/blob/main/mojap_metadata/converters/glue_converter/__init__.py


containers/daap-landing-to-raw/src/var/task/requirements.txt
```diff
@@ -1,2 +1,15 @@
 boto3==1.28.9
 botocore==1.31.9
+attrs==23.1.0
```
MatMoore, line 3: I had to copy all the base image requirements here, as otherwise they aren't installed by the test job, and they aren't installed in the image.

This problem could be solved by building a python package for the base image as proposed here https://github.com/ministryofjustice/data-platform/discussions/1613#discussioncomment-7185500

MatMoore:
I haven't integration tested this on the development environment yet - I'm going to wait until the new landing zone infrastructure is deployed first. However, the fail location is in the same bucket as the raw history so I'm not expecting any permission issues.

containers/daap-landing-to-raw/src/var/task/landing_to_raw.py
```diff
@@ -28,9 +45,34 @@ def handler(event, context):
...
-    )
+    schema_path = config.data_product_config.schema_path(table_name=config.element.name)
+
+    inferred_schema = infer_glue_schema_from_raw_csv(
+        file_path=BucketPath(key=file_key, bucket=bucket_name),
+        data_product_element=config.element,
+        logger=logger,
+    ).metadata
+    inferred_columns = extract_columns_from_schema(inferred_schema)
+    registered_schema = read_json_from_s3(schema_path)
```
MatMoore: TODO: handle errors if the schema is missing (should we skip validation if there is no schema?)

murdo-moj: I think we do skip validation if there is no schema yeah

containers/daap-landing-to-raw/src/var/task/validation.py
```diff
@@ -0,0 +1,86 @@
+class DataInvalid(Exception):
+    pass
+
+
+def type_is_compatable(registered_type: str, inferred_type: str) -> bool:
```
LavMatt: spelt `compatible`

containers/daap-landing-to-raw/src/var/task/validation.py
```diff
@@ -0,0 +1,86 @@
+class DataInvalid(Exception):
+    pass
+
+
+def type_is_compatible(registered_type: str, inferred_type: str) -> bool:
```
LavMatt, line 5: Think this function needs some work still and more complex logic. e.g. if a `tinyint` is registered but a `bigint` is inferred then this should fail as `bigint` will be outside the range of `tinyint`. So `tinyint` registered should only pass for `tinyint` inferred, `smallint` for `tinyint` and `smallint`, etc

containers/daap-landing-to-raw/src/var/task/validation.py
```diff
@@ -0,0 +1,86 @@
...
+    extracting from the source system, or the schema has changed, and the data producer needs to register a new
+    version. We do not care about row-level data quality checks at this stage.
+    """
+    registered_names = set(registered_schema_columns.keys())
+    actual_names = set(inferred_columns.keys())
+    missing_names = registered_names - actual_names
+    extra_names = actual_names - registered_names
+    if missing_names:
+        raise DataInvalid(
+            f"Columns do not match schema (missing: {missing_names}, extra: {extra_names})"
```
LavMatt, line 68: This will return extra as `extra: set()` if columns are missing but there are no extra. Might be confusing for a user

LavMatt (commented):
Couple of comments. Main thing is i think we'll need logic that's a bit more complex to pass validation for compatible types

containers/daap-landing-to-raw/src/var/task/validation.py
MatMoore: Yeah, I'm aware this will let through some cases that aren't strictly valid, and I considered this for some time before I arrived at the current version.

It may be that there is some more cases we can reject, but it seems like too big of a big rabbit hole to go down at this stage of development.

In the scenario you mentioned, it may be that 90% of the data is a valid tinyint, but just 1 row has a bigint in, causing us to infer it as a bigint. In this situation, I think ultimately in this scenario we would want to accept the dataset, and treat it as a data quality issue later in the pipeline. Data quality is not in scope for the alpha, but still we made the decision to prefer lenient validation rather than strict validation, so I decided to keep the tinyint/smallint/int/bigint relatively simple.

It's also worthing keeping in mind that rAPId's implementation is similar but (at least for now) only has one special case (for dates iirc). So I think we've already gone quite far in trying to predict possible issues, and it shouldn't be too difficult to add in more special cases later once we get it in front of some users 🙂

containers/daap-landing-to-raw/src/var/task/validation.py
MatMoore: Yeah makes sense, I'll add an extra `if` to handle it

containers/daap-landing-to-raw/tests/unit/validation_test.py
```diff
@@ -0,0 +1,169 @@
...
+        ("float", "string", False),
+        ("float", "boolean", False),
+        ("float", "timestamp", False),
+        ("float", "date", False),
+        ("double", "tinyint", True),
+        ("double", "smallint", True),
+        ("double", "int", True),
+        ("double", "bigint", True),
+        ("double", "double", True),
+        ("double", "decimal", True),
```
MatMoore, line 80: Just FYI these tests are quite slow because there's 100 of them. We could always remove some cases if we want to speed this up

containers/daap-landing-to-raw/src/var/task/validation.py
LavMatt: Hmm, but i'm not sure we could run data quality tests if the data are stored as in a scenario above because you won't be able to query that data at all. `select * from db.tbl` would just return an error, or any query with an impacted column. Could we not do something like assign a number to int types and then have some relatively simple if >= logic? so if tinyint were 1 and bigint 4 your case could have an `if registered_type => inferred_type` and return true if satisfied or else false 

Kind of thinking out loud. But if you think this wouldn't be appropriate i suppose for alpha we're unlikely to run into anything like this and it could always be developed further in beta

MatMoore: Hi @LavMatt, I gave this a try this morning, but with the current schema inference code, there is no way to make it work.

This is because infer schema doesn't return the smallest possible integer for the dataset, it just always picks a bigint. This is down to the behaviour of arrow, for example:

```python
>>> from pyarrow import csv as pa_csv
>>> from io import BytesIO
>>> csv=b"a,b,c\n1,2,3\n4,5,6"
>>> f = BytesIO(csv)
>>> table = pa_csv.read_csv(f, convert_options=pa_csv.ConvertOptions(null_values=[""]))
>>> table
pyarrow.Table
a: int64
b: int64
c: int64
----
a: [[1,4]]
b: [[2,5]]
c: [[3,6]]
```

This is the very first step done by the infer schema code, before it even calls the glue converter.

So when we take a sample of the dataset, we will infer *A schema* but not *the strictest schema*. This means if the registered schema is `tinyint` we **cannot** know whether the data is going to overflow just by using infer schema on the data set.

For this reason I don't think we can make the validation any stricter without changing the infer schema logic.

For future reference this is the variant I tested

```python
INTEGER_WIDENESS = {"tinyint": 0, "smallint": 1, "int": 2, "integer": 2, "bigint": 3}

def type_is_compatible(registered_type: str, inferred_type: str) -> bool:
    """
    Validate a type inferred from the dataset is compatible with the schema.

    This implementation adds some leniency for types that are different, but probably compatible.
    For example: a `long` type is wider than an `integer` which is wider than `short` and `byte`.
    It is always valid for the dataset to contain a narrower type than what is registered in the schema,
    such as when the schema requires long but we detected integers.

    For a full list of supported types, see the AWS athena documentation:
    https://docs.aws.amazon.com/athena/latest/ug/data-types.html
    """
    if registered_type == inferred_type:
        return True

    match (registered_type, inferred_type):
        case ("string", _):
            # Anything can go in a string
            return True
        case ("timestamp", "date"):
            # Date can go in timestamp
            return True
        case (
            "tinyint" | "smallint" | "int" | "integer" | "bigint",
            "tinyint" | "smallint" | "int" | "integer" | "bigint",
        ) if INTEGER_WIDENESS[inferred_type] <= INTEGER_WIDENESS[registered_type]:
            # Integers can go in wider integers
            return True
        case (
            "float" | "double" | "decimal",
            "tinyint" | "smallint" | "int" | "integer" | "bigint",
        ):
            # Integers can go in floats
            return True
        case "double", "float":
            # Floats can go in doubles
            return True
        case _:
            return False
```

LavMatt: Ah yes very true, we can't implement what i was thinking then you're correct.

containers/daap-landing-to-raw/src/var/task/landing_to_raw.py
```diff
@@ -28,6 +46,40 @@ def handler(event, context):
...
+            logger.error(f"{file_key} invalid; moving to fail location", exc_info=True)
+            s3.copy(
+                CopySource=copy_source,
+                Bucket=raw_data_bucket,
+                Key=fail_key,
+                ExtraArgs=s3_security_opts,
+            )
+            return
+
     s3.copy(
```
LavMatt, line 83: Priya mentioned about deleting file from land after copy, which we could add in to this PR?

MatMoore: 'm gonna create another ticket for this - I think it's low priority for now so I want to move onto the OpenMetadata library

MatMoore: https://github.com/ministryofjustice/data-platform/issues/1891

LavMatt (approved):

PR #1793, opened 2023-10-05
Add get-schema lambda

### LINKED ISSUE #1659: APi End point : Display a Table schema
The scope of the ticket would be to  create a api end point for display table schema with  path parameter  based on the document below 

The end point should be ~~/{data-product-name}/{table-schema-name}/~~ see API doc above

**Checklist** 

- [x] add code to terraform module for api infrastructure 
- [x] create a lambda for displaying the table schema

**Definition of Done** 

- [x] Able to display the  schema  for the table requested in a particular data product
- [x] unit test changed or implemented as per new change

MatMoore:
This is an initial implementation of the lambda that will serve the `/data-product/{data-product}/table/{table-name}` endpoint. Once this is in, I can define the API gateway part in terraform.

<strike>https://github.com/ministryofjustice/modernisation-platform/pull/5155 needs to be merged before this one.</strike>

## Next steps
This will eventually replace daap-get-glue-metadata, since the source of truth for a schema is now s3, rather than glue. Once this endpoint has been successfully deployed, I will remove daap-get-glue-metadata.

The response is not finalised yet - this endpoint should be consistent with https://github.com/ministryofjustice/data-platform/issues/1194 so we are thinking of simplifying the schema to just the essential parts instead of returning the full glue metadata. (See https://github.com/ministryofjustice/modernisation-platform-environments/blob/dpl-1194-create-table-schema/terraform/environments/data-platform/data-product-table-schema-json-schema/v1.0.0/moj_data_product_table_spec.json).

### DIFF
--- a/containers/daap-get-schema/src/var/task/get_schema.py
+++ b/containers/daap-get-schema/src/var/task/get_schema.py
@@ -7,17 +7,19 @@ from data_platform_logging import DataPlatformLogger
 from data_platform_paths import DataProductConfig
 from dataengineeringutils3.s3 import read_json_from_s3
 
-logger = DataPlatformLogger(
-    extra={
-        "image_version": os.getenv("VERSION", "unknown"),
-        "base_image_version": os.getenv("BASE_VERSION", "unknown"),
-    }
-)
-
 
 def handler(event, context):
     data_product_name = event["pathParameters"]["data-product-name"]
     table_name = event["pathParameters"]["table-name"]
+
+    logger = DataPlatformLogger(
+        data_product_name=data_product_name,
+        extra={
+            "image_version": os.getenv("VERSION", "unknown"),
+            "base_image_version": os.getenv("BASE_VERSION", "unknown"),
+            "table_name": table_name,
+        },
+    )
 
     config = DataProductConfig(data_product_name)
     schema_path = config.schema_path(table_name)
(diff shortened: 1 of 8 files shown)


murdo-moj:
This link is useful to check your ECR repo has the image in once you've merged (assume shared services from one of the MP accounts) https://user-guide.modernisation-platform.service.justice.gov.uk/user-guide/member-read-only-core-accounts.html#switching-roles

containers/daap-get-schema/src/var/task/get_schema.py
```diff
@@ -0,0 +1,38 @@
+import json
+import os
+
+import botocore
+from data_platform_api_responses import format_response_json, response_status_404
+from data_platform_logging import DataPlatformLogger
+from data_platform_paths import DataProductConfig
+from dataengineeringutils3.s3 import read_json_from_s3
+
+logger = DataPlatformLogger(
```
LavMatt: Could this be moved inside the handler?
Then it can be instantiated including the `"data_product_name": data_product_name` and `"table_name": table_name`. That means the log files created will have this information on every row that's logged

LavMatt:
Is your plan to merge this returning the fully registered schema, including the information that's probably not very interesting to a user and then refine once my create schema PRs are merged? Looks good other than one minor comment

MatMoore:
> Is your plan to merge this returning the fully registered schema, including the information that's probably not very interesting to a user and then refine once my create schema PRs are merged? Looks good other than one minor comment

Yeah that's the plan - if I merge it now I can work on the terraform in the meantime

LavMatt (approved):

Issue #1797, opened 2023-10-06
Use schema generated via api endpoint to create raw table in athena load


LavMatt:
### User Story

As a data producer
I need my defined schema for a table to be represented accurately in the eventual athena table created via data platform ingestion service
So that my table in consumed as intended

### Value / Purpose
Data presented to consumers will be as intended by owners of that data

### Hypothesis

If we ensure we create parquet files according to the types specified in the saved schema
Then data will be presented as intended by the owner of that data

### Proposal

To use the respective version of each schema saved to s3 to create parquet files and athena tables, rather than the schema that is inferred (as currently done in the athena load process).

This will involve casting to the types held in the schema saved in s3 for each column in a table, rather than casting to types inferred through the infer_schema function.

We also want to add the major version suffix to the curated database name as part of this ticket.

### Additional Information

Schema collected through the api endpoint will need to be trimmed where descriptions exceed the limit imposed by glue 2048 for table descriptions and 255 for column descriptions(https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-catalog-tables.html).

We want to use the types in the schema saved to s3 via the endpoint for the curated tabled creation in athena load.

I think the current process for creating the raw table can remain, i.e. use inferred schema stringyfied, but for `create_curated_athena_table` pass metadata with types and descriptions/comments from the saved schema to the `metadata` arg.

We also want to suffix the major version of a data product to the database name as part of the load
For example, the default name of a database for the data product `hmpps_use_of_force` would be `hmpps_use_of_force_v1`

### Definition of Done

- [x] athena_load process uses type and comment/description (trimmed where necessary) data from the registered schema to create the curated parquet files. Be confident metadata registered in the glue catalog matches that in the s3 schema.
- [x] data product database names are suffixed with their major version suffix.
- [x] create tests for any additional functions required.

PR #1799, opened 2023-10-06
Dpl 1194 add data product schema class

LavMatt:
This refactors `data_product_metadata.py` creating a base class and two child classes for metadata and schema. 

The `DataProductSchema` class is mostly new code to deal with creation of table schema by a user.

- Expects from user, a streamlined version of table schema with just table description and column information. We will get data product and table names from the path of the api, e.g. `data-product/{data-product-name}/table/{table-name}/schema` See initial draft of json schema [here](https://github.com/ministryofjustice/modernisation-platform-environments/blob/dpl-1194-create-table-schema/terraform/environments/data-platform/data-product-table-schema-json-schema/v1.0.0/moj_data_product_table_spec.json). Other glue table input is added from a template dict
- ~Table json schema may need to change based on discussions around what level we collect certain metadata of retention and data owners ([slack thread](https://mojdt.slack.com/archives/C03QZ776JVA/p1696597590797119)). Currently not collecting owner or retention period at the table level.~ we will collect at data product level (retention and owner) and each schema will inherit this. To be reviewed in beta
- ~Has included some blank methods that might be useful when we begin to develop using the class to update metadata or schema.~ Removed these as this now handles some versioning but outside of the class and sub-classes.
- Has a blank method to remind us we need to code adding auto generated metadata to data product metadata.
- Initial thinking is that the endpoint won't register the schema in glue catalog, this will happen when data first arrive and athena load will use the schema saved to s3 to create the tables see this ticket: #1797  

Following this will be a PR to create a new lambda container that handles the api request and calls functions to validate and write schema to s3

**UPDATE**
- Now expects a `write_key` to be passed to `write_json_to_s3()` and does not attempt to infer this
- Has a `load` method to load in latest version of metadata or schema saved to s3.
- There needs to be some interaction between schema and metadata. e.g. when a schema is created it should be added to the metadata, and if one or more schema already exist at the `schemas` key in the metadata then data product version has a minor increment applied. This will be handles in the lambda function though.


containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -360,55 +360,37 @@ def element(self, name):
         """
         return DataProductElement(name=name, data_product=self)
 
-    def metadata_path(self) -> BucketPath:
+    def metadata_path(self, version: str | None = None) -> BucketPath:
         """
         Path to the latest version of a data product metadata file
```
MatMoore: This docstring is out of date now

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,291 @@
...
-    return bucket, key
-
-
-class DataProductMetadata:
-    """
-    class to handle creation and updating of
-    metadata relating to a dataproduct
-    """
+class DataProductBaseJsonSchema:
+    """base class for operations on json type metadata and schema for data products"""
```
MatMoore, line 76: Are these classes going to handle all operations on metadata/schema, or is it really just about write operations?

I'm just thinking if it's the latter we could probably tweak the docs/names to make that more clear.

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,291 @@
...
             raise ValidationError(
                 "Metadata must be validated before being written to s3."
             )
+
+    # compare old and new metadata/schema and use
+    # logic to set version scale.
+    # will need new path function to reflect new version number schema and metadatas 3 paths
+    def _classify_update(self):
+        """infer type of increment to version, major/minor"""
+        pass
```
MatMoore: I think if you add these functions now they should raise `NotImplementedError` to prompt us to implement them all

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,291 @@
...
-        )
+
+def get_data_product_specification_path(
+    spec_type: JsonSchemaName, version: None | str = None
+) -> str:
+    """gets path for given version or latest version of metadata and schema json schema"""
+
+    # if version is empty we get the latest version
+    if version is None:
+        file_paths = get_filepaths_from_s3_folder(specification_prefix(spec_type).uri)
```
tom-webber, line 62: should we use `get_latest_version` here? ([definition](https://github.com/ministryofjustice/data-platform/blob/cd6846a08234e02376780af790365ba674c04a0d/containers/daap-python-base/src/var/task/data_platform_paths.py#L193); [already imported](https://github.com/ministryofjustice/data-platform/pull/1799/files#diff-b6cacd5f702c5b90ca80f6359737c2543b425e49aecdf98c4014e09bc518f267R13))

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,291 @@
...
-    bucket, key = path.replace("s3://", "").split("/", 1)
-    return bucket, key
-
-
-class DataProductMetadata:
-    """
-    class to handle creation and updating of
-    metadata relating to a dataproduct
-    """
+class DataProductBaseJsonSchema:
```
MatMoore: [Minor] Finding this name a bit hard to parse due to the placement of `base` in the middle. Could we perhaps drop `DataProduct` from the beginning?

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,291 @@
...
+    def __init__(
+        self, data_product_name: str, table_name: str, logger: DataPlatformLogger
+    ):
+        bucket_path = BucketPath.from_uri(
+            DataProductConfig(name=data_product_name).schema_path(table_name).uri
+        )
+        self.table_name = table_name
+
+        super().__init__(
+            data_product_name, logger, JsonSchemaName("schema"), bucket_path
```
MatMoore: [Very minor] I would write this as JsonSchemaName.schema - I think it's basically the same but you can save a few keystrokes 🤷🏻

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: Is slightly different. This gets the latest version of the metadata or schema json schema file, generic to all data products and `get_latest_version` returns the latest version of a data product. Could potentially refactor `get_latest_version` so it gets the latest version of wherever it's pointed though

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,291 @@
...
-class DataProductMetadata:
-    """
-    class to handle creation and updating of
-    metadata relating to a dataproduct
-    """
+class DataProductBaseJsonSchema:
+    """base class for operations on json type metadata and schema for data products"""
 
-    def __init__(self, data_product_name: str, logger: DataPlatformLogger):
+    def __init__(
```
MatMoore, line 78: I think this abstraction of `DataProductMetadata` is a good idea, it feels like this class extracts all the stuff that relates a working copy of some json to it's persisted version in s3. It doesn't really have much to do with data products at all any more.

If inheriting from this class is too awkward, another option is to use composition, where the other 2 class instantiate this class directly, and then delegate method calls to it. I think either approach will work here though.

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,291 @@
...
 
-    def _check_if_metadata_exists(self) -> object:
+        if not self.exists:
+            self.data_product_version = "v1.0"
+            self.is_update = False
+        else:
+            self.data_product_version = get_latest_version(self.data_product_name)
+            self.is_update = True
+
+    def _check_if_metadata_or_schema_exists(
```
MatMoore: This is partly a coding convention thing, but I think if a method is intended to be called from a subclass, it shouldn't be marked as private, and we would still want to test this as part of the tests for the base class.

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,291 @@
...
+        self, bucket: str, key: str, json_type: JsonSchemaName
```
MatMoore: It looks like this could use `self.write_bucket`, `self.write_key` and `self.type` but it's parameterised so that the schema object can query the metadata as well.

I think it would be simpler to have each object only operate on one schema file, which would mean that the DataProductSchema would instantiate a separate object to interact with the metadata (see suggested change below)

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,291 @@
...
+        self.write_key = write_bucket_path.key
+        self.exists = self._check_if_metadata_or_schema_exists(
+            self.write_bucket, self.write_key, self.type
         )
-        self.metadata_bucket = bucket
-        self.metadata_key = key
-        self._check_if_metadata_exists()
 
-    def _check_if_metadata_exists(self) -> object:
+        if not self.exists:
```
MatMoore: I think this gives us two sources of truth for the data product version - this logic, and the `write_key`.

I'm concerned we could get a conflict where the table schema is written using 1.1 but it's used metadata from 1.0 or something.

I'm also not sure I follow the logic - if write path is /foo/v1.1/metadata.json, and it doesn't exist, why should it fall back to v1.0?

Maybe we should extract the version from the write key instead of basing it on the existence of the schema/metadata.

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,291 @@
...
+    def _does_data_product_metadata_exist(self):
+        """checks wheter data product for schema has metadata registered"""
+        data_product_metadata_path = BucketPath.from_uri(
+            DataProductConfig(name=self.data_product_name).metadata_path().uri
+        )
+        md_bucket = data_product_metadata_path.bucket
+        md_key = data_product_metadata_path.key
+        return self._check_if_metadata_or_schema_exists(
+            md_bucket, md_key, JsonSchemaName("metadata")
+        )
```
MatMoore: alternatively, if every `DataProductBaseJsonSchema` instance has a `check_if_exists()` method:

```suggestion
        metadata = DataProductMetadata(self.data_product_name, self.logger)
        return metadata.check_if_exists()
```

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,291 @@
...
+        return self
+
+    def _get_parent_data_product_metadata(self) -> Dict:
+        data_product_metadata_path = BucketPath.from_uri(
+            DataProductConfig(name=self.data_product_name)
+            .metadata_path(version=self.data_product_version)
+            .uri
+        )
+
+        metadata = read_json_from_s3(data_product_metadata_path.uri)
```
MatMoore, line 299: Suggestion: We could perhaps implement this as a static `load()` method on the base class if we want it to support reads in addition to writes.

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,291 @@
...
+    # Maybe this will be better suited outside of the class.
+    def _generate_new_version_number(self):
+        """generate a version number to pass to get path function"""
+        pass
+
+
+class DataProductSchema(DataProductBaseJsonSchema):
+    """
+    class to handle creation and updating of
+    schema json files relating to data product tables
```
MatMoore, line 196: This might be a good place to also introduce what we mean by a schema, or give a simplified example, since the terminology we're using is easy to confuse, what with metadata files, table schema files, and JSON schemas for both of these

containers/daap-python-base/src/var/task/data_product_metadata.py
MatMoore: Or if we don't have a single function, we could have one for each, like `get_latest_json_schema_version` vs `get_latest_data_product_version`. I think we have to be specific whenever we talk about schemas and versions or it will get very confusing 😱

MatMoore:
I've left some suggestions and questions inline but the main approach looks solid to me.

Potentially we could also add a generic load function here too (there is a call to fetch metadata here, and we fetch schemas and metadata in other lambdas too), but we can leave it as just the validate/write methods if you prefer.

containers/daap-python-base/src/var/task/data_platform_api_responses.py
```diff
@@ -18,6 +18,13 @@ def response_status_400(error) -> dict:
     return formatted_response
 
 
+def response_status_403(error) -> dict:
```
MatMoore, line 21: Just a heads up, @tom-webber is looking at changing this module so you might need to rework it depending on what gets merged first https://github.com/ministryofjustice/data-platform/pull/1763

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: Yeah is a bit long. Happy to trim

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: This method is kind of included in the unit tests already https://github.com/ministryofjustice/data-platform/blob/dpl-1194-add-data_product_schema-class/containers/daap-python-base/tests/unit/data_product_metadata_test.py#L202-L224

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: I think reads will be useful, especially when the class is extended to deal with updates

containers/daap-python-base/src/var/task/data_platform_api_responses.py
LavMatt: I'm happy to wait and amend as saw the convo after i'd done this

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: I agree, unnecessary duplication to infer version based on this

containers/daap-python-base/src/var/task/data_product_metadata.py
MatMoore: Yeah good point - we will need it to be able to classify major/minor

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,290 @@
...
+        self.write_bucket = write_bucket_path.bucket
+        self.latest_version_key = write_bucket_path.key
+        if not self.exists:
+            self.write_key = write_bucket_path.key
+        self.version = write_bucket_path.key.split("/")[1]
 
-    def _check_if_metadata_exists(self) -> object:
+    def _check_exists(self) -> object:
+        if self.type == JsonSchemaName.data_product_schema:
+            bucket_path = BucketPath.from_uri(
```
MatMoore: is this different from the `write_key` / `write_bucket`? I was thinking it would be the same thing

LavMatt: yeah is different because when this is called from the eventual update endpoints the `write_bucket` will be the same as the latest version path, but the `write_key` will not as they'll need to be some logic applied to infer the increment type and the path for updated metadata and schema. So actually the path set at `__init__` will only be a `write_key` for new data products' metadata or schema, otherwise it's a latest version key and `write_key` should be set elsewhere after the version increment type is known.

`_check_exists` also now only checks if the instantiated object exists by just checking if there's a `v1.0`

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -93,6 +93,14 @@ def specification_path(
     )
 
 
+def get_new_version(version, increment_type):
```
MatMoore, line 96: Since this just an if/else maybe we could make it two functions instead: `get_next_major_version` `get_next_minor_version`?

Saves having the special value for `increment_type` which could be a value other than 'minor', 'major'

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,326 @@
...
+    specification_path,
+    specification_prefix,
+)
 from dataengineeringutils3.s3 import get_filepaths_from_s3_folder, read_json_from_s3
 from jsonschema import validate
 from jsonschema.exceptions import ValidationError
 
 s3_client = boto3.client("s3")
 
+glue_csv_table_input_template = {
```
MatMoore, line 22: I think this is alright for now, but @tom-webber was floating the idea earlier of having another layer in our codebase for modelling the schema and metadata structure, so that we can ensure that the representations are consistent and valid against our JSON schemas. It's hard to plan out while we are working on different endpoints in parallel, but maybe we can iterate on this once we have an initial version of each one.

The way I see it, we have two representations that we want to transform back and forth between: the one we have specced out in JSON schema, and the simpler one we are going to expose through the API. I'm curious if we can make use of something like this in powertools: https://docs.powertools.aws.dev/lambda/python/latest/utilities/parser/

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,326 @@
...
-    # if version is empty we'll get the latest version
-    if version == "":
-        file_paths = get_filepaths_from_s3_folder(
-            DataProductConfig.metadata_spec_prefix().uri
-        )
+
+def get_data_product_specification_path(
+    spec_type: JsonSchemaName, version: None | str = None
+) -> str:
+    """gets path for given version or latest version of metadata and schema json schema"""
```
MatMoore:
```suggestion
    """
    Gets the specification path for the JSON schema that validates a data product metadata or table schema.
    If a version is not specified, the latest JSON schema version is assumed.
"""
```

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,326 @@
...
+
+    # if version is empty we get the latest version
+    if version is None:
+        file_paths = get_filepaths_from_s3_folder(specification_prefix(spec_type).uri)
         versions = list(
```
MatMoore, line 63: Maybe this latter part of the function could be extracted out and reused as `sort_versions` or something.

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,326 @@
...
+            logger=logger,
+            json_type=JsonSchemaName("schema"),
+            bucket_path=bucket_path,
+            input_data=input_data,
+            table_name=table_name,
+        )
+
+        if not self._does_data_product_metadata_exist():
+            self.logger.error("Data product metadata not yet registered.")
+            self.has_registered_data_product = False
```
MatMoore, line 220: I think you could drop this boolean and instead set `parent_data_product_metadata` to `None` if it's missing

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -1,116 +1,326 @@
...
+                .latest_version_saved_data
+            )
+
+        self.parent_product_has_registered_schema = (
+            self._does_data_product_have_other_schema_registered()
+        )
+
+    def _does_data_product_metadata_exist(self):
+        """checks whether data product for schema has metadata registered"""
+        metadata = DataProductMetadata(
```
MatMoore, line 239: this can be replaced with `self.parent_data_product_metadata` now (checking it's not None first)

MatMoore (dismissed):
Had another look through and I think it looks good to me, I don't have any blocking concerns, just a few take it or leave it suggestions.

Regarding splitting out the versioning logic, I'm not sure what it's going to look like until I see the lambda itself, but intuitively that seems like a good idea if it will keep these classes smaller & simpler, and especially if there's less mutable state to keep track of, as that can make the code harder to reason about in your head.

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: yeah that sounds like a good idea to me

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: Thanks, going to keep as is at the moment as it interacts with the way i've coded the lambda handler

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: as above

MatMoore (approved):

PR #1831, opened 2023-10-09
Dp 1618 Preserve file extensions

### LINKED ISSUE #1618: Preserve filenames and file extensions in raw data
Currently raw data is upload with a special filename we generate, and we drop the file extension.

Instead, we should allow providers to choose the filename.

https://github.com/ministryofjustice/data-platform/pull/1549#issuecomment-1727944755

mitchdawson1982:
Key changes:

- implements event.body as a string type and makes associated tests compliant
- implements uploading of filename parameter in the event body
- implements validation for empty filename and file extension
- Implements additional test for empty filename validation
- Adds filename support to existing tests responsible for generating file paths
- Presigned url uses python base version 3.2 
- Uplifts app versions

### DIFF
--- a/containers/daap-presigned-url/Dockerfile
+++ b/containers/daap-presigned-url/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/ministryofjustice/data-platform-daap-python-base:3.1.0
+FROM ghcr.io/ministryofjustice/data-platform-daap-python-base:3.2.0
 
 ARG VERSION
 
--- a/containers/daap-presigned-url/Dockerfile
+++ b/containers/daap-presigned-url/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/ministryofjustice/data-platform-daap-python-base:3.2.0
+FROM ghcr.io/ministryofjustice/data-platform-daap-python-base:3.3.0
 
 ARG VERSION
 
--- a/containers/daap-presigned-url/src/var/task/presigned_url.py
+++ b/containers/daap-presigned-url/src/var/task/presigned_url.py
@@ -2,6 +2,7 @@ import json
 import os
 import uuid
 from datetime import datetime
+from pathlib import PurePath
 
 import boto3
 from data_platform_logging import DataPlatformLogger
@@ -45,7 +46,7 @@ def handler(event, context):
             ),
         }
 
-    file_extension = os.path.splitext(filename)[1]
+    file_extension = PurePath(filename).suffix
     if file_extension == "":
         return {
             "statusCode": 400,
--- a/containers/daap-presigned-url/src/var/task/presigned_url.py
+++ b/containers/daap-presigned-url/src/var/task/presigned_url.py
@@ -50,7 +50,7 @@ def handler(event, context):
         return {
             "statusCode": 400,
             "headers": {"Content-Type": "application/json"},
-            "body": json.dumps({"error": {"message": "file extension is invalid"}}),
+            "body": json.dumps({"error": {"message": "file extension is invalid."}}),
         }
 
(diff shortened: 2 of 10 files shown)


containers/daap-presigned-url/src/var/task/presigned_url.py
```diff
@@ -20,30 +20,46 @@
...
                     "error": {
-                        "message": "data product name or table name value are not"
+                        "message": "data product name, table name or filename are not"
                         + " convertible to string type."
                     }
                 }
             ),
         }
 
+    file_extension = os.path.splitext(filename)[1]
```
tom-webber:
```suggestion
    from pathlib import PurePath  # top of file import if using this suggestion
    file_extension = PurePath.suffix(filename)
```
🎨 suggestion: I highly recommend the pathlib module for interacting with paths. As we're not actually representing an OS path, use PurePath in place of Path
Note: this will return the whole extension, so `.csv` for a csv file.

tom-webber: There's also a way to augment/extend this later for more complex suffixes e.g. `.SNAPPY.parquet`, using Path.suffixes, which collects all suffixes on a file.

containers/daap-presigned-url/tests/unit/mock_test.py
```diff
@@ -44,7 +49,7 @@ def test_success(s3_client, fake_context, region_name, monkeypatch):
     assert body["URL"]["url"] == "https://bucket.s3.amazonaws.com/"
     assert (
         body["URL"]["fields"]["key"]
-        == f"raw/database1/v1.0/table1/load_timestamp=20230101T000000Z/{a_uuid}"
+        == f"raw/database1/v1.0/table1/load_timestamp=20230101T000000Z/{a_uuid}{os.path.splitext(filename)[1]}"
```
tom-webber: See pathlib comment above - if implementing that way, likely want to reflect here.

I would make this more explicit for the expectation, potentially by splitting out the filename into filename + extension above
```suggestion
... filename = "testdata"  # defined at the top of the function
... extension = ".csv"
... file_ = filename + extension
        == f"raw/database1/v1.0/table1/load_timestamp=20230101T000000Z/{a_uuid}{extension}"
```

containers/daap-presigned-url/tests/unit/mock_test.py
```diff
@@ -11,6 +12,7 @@ def test_success(s3_client, fake_context, region_name, monkeypatch):
     bucket_name = "bucket"
```
tom-webber, line 12: 💡 Suggestion: Can we pull these constants out into a dict we're passing to mark.parameterize?

containers/daap-presigned-url/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/ministryofjustice/data-platform-daap-python-base:3.1.0
+FROM ghcr.io/ministryofjustice/data-platform-daap-python-base:3.2.0
```
MatMoore: I don't think this is correct as the version below is 3.3.0.

I think you will need to split the base image part into a separate PR, so that the image will be available to this build.

containers/daap-presigned-url/tests/unit/mock_test.py
mitchdawson1982: Now using PurePath and new file_extension variable has ben created and referenced in the assertion.

containers/daap-presigned-url/src/var/task/presigned_url.py
mitchdawson1982: Now using PurePath as suggested 👍🏻

MatMoore (approved):

PR #1844, opened 2023-10-10
Add data platform labs to create epic filter

PriyaBasker23:
Added data platform labs to create epic filter


julialawrence (changes requested):
Lines 26-31 were only put in as a temp measure until you could review the code. If you're happy to use these now, just delete lines 26-31 and you're good to go.

julialawrence (approved):
Looks good.

PR #1873, opened 2023-10-11
create_schema lambda handler

### LINKED ISSUE #1194: Create Table  Schema and register
### User Story

As a data producer
I want to define a schema for my data product table and register it with the data platform.
So that best i can ensure conformance against that schema in future uploads, and supply useful information to the data platform catalogue for consumers of my data product.

### Proposal

1. Write a python function/class that will allow to create a schema, plus unit tests
2. Create an api end point , where schema can be send and created 
3. Metadata information about the file extension type and load type should be included  Create a schema and store in folder eg: /data-product-bucket/scheme/databasename=x/table_name=x/schema.json     

  ( No Schema Versioning)           

### Definition of Done

- [x] will use glue schema format and have a schema file for the schema to validate table schema against (using pythons jsonschema.validate)
- [x] define the spec of the schema object in a json schema
- [x] create new shared ecr repo for each version of the lambda container - modernisation-platform repo.
- [x] create new lambda container in data-platform repo, with the code api calls to execute the required actions.
- [x] make all infra in modernisation-platform-environments. api-gateway, lambda, iam
- [ ] Once above tasks complete, add documentation for api endpoint.

LavMatt:
lambda code to be invoked by endpoint `data-product/{data-product-name}/table/{table-name}/schema`

it implements versioning to a point, where new schema added to an existing data product with one or more tables already results in a minor increment to version and so a new version of metadata and schema are created.

This currently on handles single schema loads. Further development or maybe a different endpoint would be needed for batch schema loads


containers/daap-create-schema/src/var/task/create_schema.py
```diff
@@ -0,0 +1,168 @@
...
+def s3_copy_folder_to_new_folder(
+    bucket, source_folder, latest_version, new_version, logger
+):
+    paginator = s3_client.get_paginator("list_objects_v2")
+    page_iterator = paginator.paginate(
+        Bucket=bucket,
+        Prefix=source_folder,
+    )
+    keys_to_copy = []
+    try:
```
PriyaBasker23, line 27: The  page_iterator will return a empty list if the paginator does not contain any items.so you can test using 

```
if not any(pages):
    print("The folder is empty.")
else: 
    Loop 
```

Dont need a try / expect here

containers/daap-create-schema/src/var/task/create_schema.py
```diff
@@ -0,0 +1,168 @@
...
+        if not schema.parent_product_has_registered_schema:
+            metadata_dict = (
+                DataProductMetadata(
+                    data_product_name=data_product_name,
+                    logger=logger,
+                    input_data=None,
+                )
+                .load()
+                .latest_version_saved_data
+            )
```
tom-webber:
```suggestion
            metadata_dict = schema.parent_data_product_metadata
```
This looks to be reproducing [the logic from the DataProductSchema class](https://github.com/ministryofjustice/data-platform/blob/main/containers/daap-python-base/src/var/task/data_product_metadata.py#L223-L231)

LavMatt: Yeah you're right, nice spot. Good to get a fresh pair of eyes on it

containers/daap-create-schema/src/var/task/create_schema.py
LavMatt, line 27: Doesn't seem to be the case, the page_iterator contains data even if a specified prefix does not exist so it fails with a `KeyError`

containers/daap-create-schema/src/var/task/create_schema.py
```diff
@@ -0,0 +1,154 @@
...
+        return response_status_403(error_msg)
+
+    # Code below that handles verisoning of a data product will be moved to a central module eventually.
+
+    # if schema already exist then we need to minor version increment to dataproduct metadata and schema
+    if schema.valid:
+        schema.convert_schema_to_glue_table_input_csv()
+        if not schema.parent_product_has_registered_schema:
+            metadata_dict = schema.parent_data_product_metadata
+            # metadata_with_schema = metadata_dict
```
PriyaBasker23: Why is  this commented out ?

containers/daap-create-schema/src/var/task/create_schema.py
PriyaBasker23: The paginate method then returns an iterable PageIterator
```
for page in PageIterator:
  if page['KeyCount'] != 0:
     continue 
```

containers/daap-create-schema/src/var/task/create_schema.py
LavMatt: It shouldn't be there, i'll remove

containers/daap-create-schema/src/var/task/create_schema.py
```diff
@@ -0,0 +1,153 @@
...
+                DataProductConfig(data_product_name).metadata_path("v1.0").key
+            )
+            schema.write_json_to_s3(
+                DataProductConfig(data_product_name).schema_path(table_name, "v1.0").key
+            )
+            msg = f"Schema for {table_name} has been created in the {data_product_name} data product"
+            logger.info("Schema successfully created")
+            return response_status_200(msg)
+        else:
+            # write to next minor version increment
```
PriyaBasker23, line 115: Just want to reiterate the logic 

 When there is a schema already and new schema is added , minor version  increment happens 
1. New version is generated  for schema
2. New folder created 
3. Data product metadata is copied to it se folder
4. New schema is added to the s3 folder

PriyaBasker23: Does   `  s3_copy_folder_to_new_folder` copies other schema to the new version folder ?

LavMatt: Yeah so the `DataProductSchema` class checks whether any schema have already been registered in its parent data product and assigns a boolean to a property for it. That property is used to know whether to add the schema to the current version of the metadata and save schema in v1.0  folder, or to copy latest version folder's contents to next minor version folder, append schema name to the data product metadata and save schema and metadata files in that next version folder.

If you look at the old products-development bucket in the test_product folder you'll see some different version folders i created when testing (this will go in the metadata bucket now but it didn't exist at the time)

PriyaBasker23 (approved):

PR #1897, opened 2023-10-12
📝 Update development documentation

jacobwoffenden:
Adds:
- 1Password setup

TODO:
- [x] Decide on language on how we make it mandatory to use 1Password for SSH key
- [x] Decide on if we make it mandatory to sign commits
  - NO: Requiring signed commits doesn't play nice with automation


Gary-H9 (dismissed):
LGTM

Gary-H9 (dismissed):

docs/source/documentation/platform/infrastructure/developing.html.md.erb
```diff
@@ -58,6 +66,16 @@ Once we begin our journey to the monorepo, this should become redundant, but tha
 
 Your GPG and SSH agents are mounted from macOS, so GPG and SSH commands inside the container should work without any configuration updates.
 
+1. Follow 1Password's instructions on how to setup SSH keys and their SSH agent ([link](https://developer.1password.com/docs/ssh/get-started/))
+
+> If you already have a key, you can skip creating a new one and just add your existing key to 1Password. ([link](https://developer.1password.com/docs/ssh/manage-keys/#import-an-ssh-key))
```
michaeljcollinsuk:
```suggestion
> If you already have an existing key, you can add it to 1Password ([link](https://developer.1password.com/docs/ssh/manage-keys/#import-an-ssh-key)) and skip to step 3 of 1Password's instructions.
```
I foolishly interpreted this as meaning I just needed to follow the second link instead, and got a bit confused as I thought there still be extra steps about update my ssh config. So I think this would be a bit clearer?

michaeljcollinsuk: Also, odd thing that I discovered - when downloading the unencrypted private key from 1password, it is different to the original, but still works.

bagg3rs (approved):

PR #1901, opened 2023-10-12
Add a new repo for new library

MatMoore:
This library will wrap the OpenMetadata SDK and enable the API to write metadata to the data catalogue.


MatMoore:
Closing for now - we're going to initially deploy this as part of the monorepo.

PR #1935, opened 2023-10-16
Pre-signed url lambda version 1.27

mitchdawson1982:
Removes
- formatted_data variable
- {"x-amz-date": formatted_date} from fields and conditions.


MatMoore (approved):
Just to check my understanding, does this change mean we create a URL that's valid for 7 days as per https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html#:~:text=A%20presigned%20URL%20can%20be%20valid%20for%20a%20maximum%20of%20seven%20days%20because%20the%20signing%20key%20you%20use%20in%20signature%20calculation%20is%20valid%20for%20up%20to%20seven%20days.?

PR #1941, opened 2023-10-17
create-a-derived-table OIDC role

jhpyke:
This PR seeks to create an OIDC based role that we can use to deploy tables from Create-A-Derived-Table. 

To Do:

- [x] Add longer timeout on new role.
- [x] Figure out why things are being removed from the path filter.


BrianEllwood (approved):
LGTM

PR #1960, opened 2023-10-17
DP-1833 First draft of the catalogue library (python 3.10)

MatMoore:
https://github.com/ministryofjustice/data-catalogue/issues/108

This PR introduces a class for interacting with the data catalogue (OpenMetadata). This is meant to be a wrapper around the OpenMetadata SDK to simplify the integration and abstract away implementation details that would change if we switched to another catalogue.

The goals of this PR is to prove that we can actually get data into the catalogue, and get some initial feedback on the proposed API.

I've based a lot of this on this guide: https://docs.open-metadata.org/v1.1.x/sdk/python

After this I will need to
- create a workflow to deploy to Pypi
- ensure that we can pass through all the metadata we currently have in our JSON schemas
- implement deletion methods

## Caveats
- This does not yet support 3.11 because the OpenMetadata SDK doesn't support it. However, I asked on the slack and they said it will be fixed in upcoming releases. I suggest we temporarily downgrade lambdas to 3.10 until this is resolved.

## Answered questions
1. I'm not sure what I should be passing for the `connection` configuration when creating the database service. I used CustomDatabase for now as it doesn't require any special credentials.

    **Resolved**: we can pass `{"connection": {"config": {}}}` here, although v1.1.7 makes this awkward

2. Does the way I've described the hierarchy make sense? I'm assuming we can group everything under one service, and have 1 database and 1 schema per data product. 

   **Resolved**: see diagram below

4. I'm not sure yet how owner metadata will work - can we assume that the data producer will be added as a user to the catalogue at the time when we grant them API access? Or should we add extra methods to add users through this library?

    **Resolved**:  Creating users is out of scope for this ticket. We will use the data labs user to test with.


MatMoore:
In case we need it later, here is an example of creating a service with the glue connection instead of custom database:

```python
    credentials = AWSCredentials(
        awsRegion="eu-west-1",
        awsAccessKeyId=aws_access_key,
        awsSecretAccessKey=secret_key,
        awsSessionToken=session_token,
    )

    create_service = CreateDatabaseServiceRequest(
        name="mat-test-database-service",
        serviceType=DatabaseServiceType.Glue,
        connection=DatabaseConnection(
            config=GlueConnection(
                awsConfig=credentials
            )
        ),
    )
```

MatMoore:
This change covers syncing of data **from** the ingestion component **to** the catalogue. However, @tom-webber pointed out that it is possible for data owners to update descriptions, owners etc. in OpenMetadata as well, which would imply a flow of metadata in the other direction as well.

If we configure the database service with a Glue connection, then OpenMetadata is able to update the Glue catalogue, but this wouldn't automatically update the schema/metadata files in s3, which is what we are currently considering the source of truth.

Syncing data in 2 directions is complex and error prone, so we would like to avoid this if possible. But it would also be nice if users could make use of OpenMetadata's built-in governance functionality, rather than forcing all updates to go via our API.

This problem would go away if we used OpenMetadata as our entity store, and got rid of the metadata bucket, but we're reluctant to do this in alpha as this will tightly couple the solution to OpenMetadata.

PriyaBasker23:
Database topology 
image

PriyaBasker23:
creating users - For now users can be created manually. We can liaise with the app/tools team about what is plan to populate users . I think this might be out of more remot @jacobwoffenden @julialawrence

PriyaBasker23:
Can we create a new user[dataplatformlabs@digital.justice.gov.uk](mailto:dataplatformlabs@digital.justice.gov.uk) and user it add owners for test case. We can create using postman/ api  end point. Let me know if you need creating

MatMoore:
Created https://catalogue.apps-tools.development.data-platform.service.justice.gov.uk/users/dataplatformlabs

PriyaBasker23:
I could create a Database service using api endpoint without needing AWS credentials 

image

MatMoore:
> I could create a Database service using apiend with needing aws credentials
> 
> image

Unfortunately it gives me a validation error if I do this through python

i.e.

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
data_platform_catalogue/client.py:88: in create_or_update_database_service
    connection=DatabaseConnection(config=GlueConnection()),
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   pydantic.error_wrappers.ValidationError: 1 validation error for GlueConnection
E   awsConfig
E     field required (type=value_error.missing)
```

It's marked as required in this schema: https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-spec/src/main/resources/json/schema/entity/services/connections/database/glueConnection.json#L49

PriyaBasker23:
Can you please if you can create a connection using the auth token rather than AWS credentials

PriyaBasker23:
Currently the scope of the ticket is to send data to Openmetadata endpoint. we dont have to connect to glue Database in AWS account. 

For the scenario,  automatic pull from database can be set as Glue connection workflow ( as part of IAC code to have trigger in DAG)

PriyaBasker23:
Possiblility for data owners to update descriptions -> resync, is out of scope of for Alpha. Again we can have discussions with apps and tools team to decide how to implement it. @julialawrence @jacobwoffenden

MatMoore:
I had some difficulty creating a Glue service without a connection in version 1.1.7 of openmetadata.

Apparently this is fixed in 1.2.0:

Screenshot 2023-10-19 at 14 39 18


After some experimentation, this works:

```json
{"name": "dp-glue3", "displayName": "data platform glue catalogue test3", "description": null, "tags": null, "serviceType": "Glue", "connection": {"config": {}}}
```

But this doesn't:

```
{"name": "dp-glue4", "displayName": "data platform glue catalogue test3", "description": null, "tags": null, "serviceType": "Glue", "connection": {"config": null}}
```

->

```
{
	"code": 400,
	"message": "InvalidServiceConnectionException for service [Glue] due to [Failed to encrypt connection instance of Glue]."
}
```

The problem I was having was that the following code generates the non-working version

```python
        create_service = CreateDatabaseServiceRequest(
            name=name,
            displayName=display_name,
            serviceType=DatabaseServiceType.Glue,
            connection=DatabaseConnection(),
        )
```

So as a temporary workaround I've bypassed the python objects so that we construct the json ourselves. We should be able to remove this once we've updated our instance, but for now it will give us something to work with.

PriyaBasker23 (approved):

PriyaBasker23:
LGTM

jacobwoffenden (approved):

PR #1973, opened 2023-10-18
:fire: dpl 1662 delete table for a data product

### LINKED ISSUE #1662: API end point : Remove Table Data and Schema
Image

### User Story

As a data producer / controller 
I need the platform to provide the ability to remove a table and all associated data
So that I can easily remove unwanted tables,  schemas and data when I no longer require them.

### Value / Purpose
With this functionality users will be able to remove tables and schemas that they no longer require. This will avoid scenarios where data that is no longer required is left in situ using resources that could otherwise be deployed elsewhere. Additionally buy automating the process we empower users, avoiding the need to log into the aws console and or requiring administrators input.

### Hypothesis
For Alpha we will delete the relevant schema and table referenced with the url path (below) when a DELETE request is received.

API Endpoint - /data-product/{data-product-name}/table/{table-name}
HTTP Verb - DELETE

### Additional Information
- All data deleted
- Major version bump required

###
Key Considerations
- although a schema could in theory be loaded to s3 without data and if there’s no data then a table in athena wouldn’t exist. So probably needs to check if associated data and tables exist before trying to delete them as well as schema file
- Metadata bucket file structure 
- `/test_product/v1.2/all_types1/schema.json` this is the path to version 1.2 of the schema.json file schema "all_types1" for data-product "test_product".

# Delete Schema Checklist
- [ ] Does raw and or curated data exist
- [ ] If exists delete raw and curated data
- [ ] for each version of the data product check the metadata.json file for the existence of the schema under "schemas"
`{
    "name": "test_product",
    "description": "just testing the metadata json validation/registration",
    "domain": "MoJ",
    "dataProductOwner": "matthew.laverty@justice.gov.uk",
    "dataProductOwnerDisplayName": "matt laverty",
    "email": "matthew.laverty@justice.gov.uk",
    "status": "draft",
    "dpiaRequired": false,
    "retentionPeriod": 3000,
    "schemas": [
        "all_types1",
        "all_types2",
        "all_types3"
    ]
}`

If the schema is present delete the schema.json for the version  i.e. if deleting "all_types1":
delete `/test_product/v1.0/all_types1/schema.json`

- [ ] Remove glue table?
- [ ] Create new major version

The remove table schema should remove all the glue table and data in curated bucket. data in raw data bucket should not be removed

Checking against the data retention is out of scope 

### Checklist

- [ ] add code to terraform module for api infrastructure change 
- [ ] create a lambda for removing the table schema

### Definition of Done

- [ ] Able to remove the  schema  for the table requested in a particular data product

mitchdawson1982:


containers/daap-delete-table-for-data-product/src/var/task/delete_table.py
```diff
@@ -0,0 +1,48 @@
...
+            "image_version": os.getenv("VERSION", "unknown"),
+            "base_image_version": os.getenv("BASE_VERSION", "unknown"),
+            "data_product_name": data_product_name,
+            "table_name": table_name,
+        }
+    )
+
+    logger.info(f"event: {event}")
+
+    if not http_method == HTTPMethod.POST:
```
MatMoore: I'm not sure if we need to worry about checking the method here - I would expect this to be enforced at the API gateway?

There is also a specific code for this, 405 Method Not Allowed

containers/daap-delete-table-for-data-product/src/var/task/delete_table.py
```diff
@@ -0,0 +1,123 @@
...
+        "base_image_version": os.getenv("BASE_VERSION", "unknown"),
+    }
+)
+
+s3_client = boto3.client("s3")
+glue_client = boto3.client("glue")
+athena_client = boto3.client("athena")
+
+
+def s3_recursive_delete(bucket, prefix) -> None:
```
MatMoore: Just wondering if you tried using the `filter` method described in this stack overflow answer? https://stackoverflow.com/questions/11426560/amazon-s3-boto-how-to-delete-folder

Seems like you can do the same thing more simply, but I haven't tried it out

containers/daap-delete-table-for-data-product/src/var/task/delete_table.py
```diff
@@ -0,0 +1,123 @@
...
+        input_data=None,
+    )
+
+    # Validate the existence of the table schema for the latest data product version
+    if not table_schema.exists:
+        error_message = f"Could not locate valid schema for table: {table_name}."
+        logger.error(error_message)
+        return format_error_response(HTTPStatus.BAD_REQUEST, event, error_message)
+
+    # Attempt deletion of the glue table
```
MatMoore, line 147: (minor) might be worth extracting this part into a helper function, just to help with readability

containers/daap-delete-table-for-data-product/src/var/task/delete_table.py
```diff
@@ -0,0 +1,123 @@
...
+        return format_error_response(HTTPStatus.BAD_REQUEST, event, error_message)
+    else:
+        glue_client.delete_table(DatabaseName=data_product_name, Name=table_name)
+
+    # Proceed to delete the raw data
+    element = DataProductElement.load(
+        element_name=table_name, data_product_name=data_product_name
+    )
+
+    raw_bucket = element.data_product.raw_data_bucket
```
MatMoore: I think this will be scoped to the current version only, so we won't delete any historical data (which I think we wanted to do)

I'm wondering if we need to add some way to get all the paths to historical versions of a table, since they won't have a common prefix.

`/raw/some-product/v1.2/some-table/...` vs `/raw/some-product/v1.1/some-table/...`

mitchdawson1982: Yes agree, thought that there was agreement just to delete current data but actually not sure that this makes sense any longer.

containers/daap-delete-table-for-data-product/src/var/task/delete_table.py
```diff
@@ -0,0 +1,173 @@
...
+    if not metadata_versions.get("Contents"):
+        return ["v1.0"]
+
+    versions = [
+        version["Key"].split("/")[1]
+        for version in metadata_versions["Contents"]
+        if version["Size"] > 0
+    ]
+    versions.sort(key=lambda x: [int(y.replace("v", "")) for y in x.split(".")])
+    print(versions)
```
MatMoore: I think this print was left in by accident?

containers/daap-delete-table-for-data-product/src/var/task/delete_table.py
```diff
@@ -0,0 +1,173 @@
...
+    # This is the case in which the data product is new.
+    if not metadata_versions.get("Contents"):
+        return ["v1.0"]
+
+    versions = [
+        version["Key"].split("/")[1]
+        for version in metadata_versions["Contents"]
+        if version["Size"] > 0
+    ]
+    versions.sort(key=lambda x: [int(y.replace("v", "")) for y in x.split(".")])
```
MatMoore: does it matter if they're sorted? Maybe we could skip this line?

containers/daap-delete-table-for-data-product/src/var/task/delete_table.py
```diff
@@ -0,0 +1,173 @@
...
+)
+
+s3_client = boto3.client("s3")
+s3_resource = boto3.resource("s3")
+glue_client = boto3.client("glue")
+athena_client = boto3.client("athena")
+
+
+def get_all_versions(data_product_name: str) -> list[str]:
+    """Get the latest version of a data product, else return v1.0"""
```
MatMoore: This docstring needs updating I think

containers/daap-delete-table-for-data-product/src/var/task/delete_table.py
```diff
@@ -0,0 +1,173 @@
...
+    )
+    if glue_error is not None:
+        return glue_error
+
+    # Delete a given elements raw and curated data for all versions of the data product
+    delete_all_element_version_data_files(
+        data_product_name=data_product_name, table_name=table_name
+    )
+
+    # Proceed to delete the raw data
```
MatMoore: I guess these lines can be deleted now

containers/daap-delete-table-for-data-product/src/var/task/delete_table.py
```diff
@@ -0,0 +1,173 @@
...
+        for version in metadata_versions["Contents"]
+        if version["Size"] > 0
+    ]
+    versions.sort(key=lambda x: [int(y.replace("v", "")) for y in x.split(".")])
+    print(versions)
+
+    return versions
+
+
+def generate_all_element_version_prefixes(
```
MatMoore, line 43: we might want to move some of this into the base image, but the logic looks good.

I wish there was a way to just pass a wildcard into the delete function but I don't think there is 😞

MatMoore (dismissed):
Looks good, just noticed a few things that can be removed/cleaned up

MatMoore (approved):

PR #2014, opened 2023-10-20
Modify catalogue client to accept full metadata objects and hide internal exceptions

MatMoore:
Previously, the client was creating minimal Tables and Schemas in OpenMetadata with just a name.

However, we should be passing through any metadata that we capture when creating a data product or table.
See JSON schemas: https://github.com/ministryofjustice/modernisation-platform-environments/tree/main/terraform/environments/data-platform

This change forces the user of this library to pass through the following metadata:

- name, description for database/schema/table
- retention period at schema/table level
- version, owner, email, dpi_required, domain at schema level

We can also accept optional tags at any level. These are expected to correspond to classifications in OpenMetadata. By default we have tags for PII sensitivity and "tier". https://catalogue.apps-tools.development.data-platform.service.justice.gov.uk/tags/

**Note**: Email, dpia_required, domain, versions are not yet passed through to the catalogue. These may require use of custom attributes. We can only do these at the table level at the moment, but 1.2.0 will add support at other levels.

Tags and owner are assumed to reference things that already exist in OpenMetadata - if not, we raise an error.

### DIFF
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
@@ -17,7 +17,17 @@ from metadata.generated.schema.entity.services.databaseService import (
 from metadata.generated.schema.security.client.openMetadataJWTClientConfig import (
     OpenMetadataJWTClientConfig,
 )
+from metadata.generated.schema.type.basic import Duration
+from metadata.generated.schema.type.entityReference import EntityReference
+from metadata.generated.schema.type.tagLabel import (
+    LabelType,
+    State,
+    TagLabel,
+    TagSource,
+)
 from metadata.ingestion.ometa.ometa_api import OpenMetadata
+
+from .entities import CatalogueMetadata, DataProductMetadata, TableMetadata
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


jacobwoffenden:
@MatMoore can you set `@ministryofjustice/data-platform-labs` to be a CODEOWNER for `python-libraries/data-platform-catalogue/*`, thanks! 🙏

jacobwoffenden (approved):
Approving CODEOWNERS change

python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
```diff
@@ -94,46 +121,66 @@ def create_or_update_database_service(
...
         return self._create_or_update_entity(create_db)
 
-    def create_or_update_schema(self, name: str, database_fqn: str):
+    def create_or_update_schema(self, metadata: DataProductMetadata, database_fqn: str):
         """
         Define a database schema.
         There should be one schema per data product.
         """
-        create_schema = CreateDatabaseSchemaRequest(name=name, database=database_fqn)
+        create_schema = CreateDatabaseSchemaRequest(
```
PriyaBasker23, line 142: You are collecting extra information  for schema like DPIA required but can see any custom properties created for those ?

PriyaBasker23: Functionality for custom property is available for tables  but not for database

PriyaBasker23: Discussed with   Mat. Will be implemented later

PriyaBasker23 (approved):

python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
MatMoore: New ticket: https://github.com/ministryofjustice/data-platform/issues/2022

PR #2020, opened 2023-10-23
Update metadata lambda

MatMoore:
This adds the new lambda for the update metadata endpoint https://github.com/ministryofjustice/data-platform/issues/1657.

This should 
- create a new minor version in the metadata bucket
- copy all schemas from the previous version

<s>I decided the schema copying should be handled in the base image so I created a separate PR for that.
https://github.com/ministryofjustice/data-platform/pull/2053</s>

This endpoint shouldn't copy any actual data. In the future, we will have a athena database per major version only, so minor versions will produce the same product.


containers/daap-update-metadata/src/var/task/update_metadata.py
```diff
@@ -0,0 +1,56 @@
...
+    extra={
+        "image_version": os.getenv("VERSION", "unknown"),
+        "base_image_version": os.getenv("BASE_VERSION", "unknown"),
+    }
+)
+
+
+def handler(event, context):
+    data_product_name = event["pathParameters"].get("data-product-name")
+    table_name = event["pathParameters"].get("table-name")
```
murdo-moj: is this needed to update the version within the metadata object? (we could just remove it from the code if it's not)

containers/daap-update-metadata/tests/unit/update_metadata_test.py
```diff
@@ -0,0 +1,36 @@
...
+def test_handler(s3_client, fake_event, fake_context):
+    s3_client.put_object(
+        Body=json.dumps({"description": "test_p"}),
+        Bucket="test",
+        Key="test_p/v1.0/metadata.json",
+    )
+
+    response = handler(fake_event, fake_context)
+
+    assert response == {
```
murdo-moj, line 31: could also assert that the new metadata file exists in s3? or is this tested in the tests for update_version?

containers/daap-update-metadata/tests/unit/update_metadata_test.py
```diff
@@ -0,0 +1,36 @@
...
+        "body": json.dumps({"version": "v1.1"}),
+        "statusCode": HTTPStatus.OK,
+        "headers": {"Content-Type": "application/json"},
+    }
```
murdo-moj, line 35: You could also tests a few of the error cases to make sure they work

containers/daap-update-metadata/src/var/task/update_metadata.py
MatMoore: Oh no, this shouldn't be here at all. I'll remove it

containers/daap-update-metadata/tests/unit/update_metadata_test.py
MatMoore: Yeah I'm gonna unit test this in the versioning module so not too worried about the details here

murdo-moj (approved):

PR #2025, opened 2023-10-23
Remove Parminder and Samuel for JML

julialawrence:
At the request of GMoller

One doesn't have the right checks and the other has left MOJ.


jacobwoffenden:
Will action in https://github.com/ministryofjustice/data-platform/pull/1985

PR #2027, opened 2023-10-23
DP-1657 Add update metadata function to base image

MatMoore:
This PR is preparation for adding an "update metadata" endpoint.

I've added some functionality to the metadata class to detect whether an update is minor or not; and if it is minor, to create a new minor version (we'll deal with major updates later).

This is implemented at the DataProductMetadata level rather than in the base class, since the schema update case is more complicated and until we start working on that, I'm not sure what we'll be able to reuse from this implementation.

I've also moved the `format_error_response` function into the base since I saw we were creating this in a few different places. For consistency, I made format_response_json also do the `json.dumps` which is a breaking change but very easy to update.

### DIFF
--- a/containers/daap-python-base/src/var/task/data_product_metadata.py
+++ b/containers/daap-python-base/src/var/task/data_product_metadata.py
@@ -98,7 +98,20 @@ def get_data_product_specification_path(
 
 
 class BaseJsonSchema:
-    """base class for operations on json type metadata and schema for data products"""
+    """
+    Base class for operations on json type metadata and schema for data products.
+
+    Parameters:
+    - bucket_path should be the path to the latest version, or 1.0.0 if no version exists
+    - input_data is optional data to be validated and written
+
+    Attributes:
+    - exists: does *any* version of this schema exist?
+    - valid: is the input_data valid?
+    - version: the version string corresponding to the `bucket_path`
+    - write_bucket / latest_version_key: the bucket and key components of `bucket_path`, respectively
+    - latest_version_saved_data: the metadata stored at `bucket_path`
+    """
 
     def __init__(
         self,
@@ -118,6 +131,7 @@ class BaseJsonSchema:
         self.exists = self._check_a_version_exists()
         self.write_bucket = bucket_path.bucket
         self.latest_version_key = bucket_path.key
+        self.latest_version_saved_data = None
         self.version = bucket_path.key.split("/")[1]
         if input_data is not None:
             self.validate(input_data)
--- a/containers/daap-python-base/src/var/task/data_product_metadata.py
+++ b/containers/daap-python-base/src/var/task/data_product_metadata.py
@@ -136,6 +136,7 @@ class UpdateType(Enum):
     Unchanged = 0
     MinorUpdate = 1
     MajorUpdate = 2
+    NotAllowed = 3
 
 
 class BaseJsonSchema:
@@ -416,11 +417,18 @@ class DataProductMetadata(BaseJsonSchema):
     schema json files relating to data products as a whole
     """
 
-    MINOR_UPDATE_FIELDS = {
+    UPDATABLE_FIELDS = {
         "description",
         "email",
         "dataProductOwner",
         "dataProductOwnerDisplayName",
+        "domain",
+        "status",
+        "dpiaRequired",
+        "retentionPeriod",
+        "dataProductMaintainer",
+        "dataProductMaintainerDisplayName",
+        "tags",
     }
 
     def __init__(
@@ -451,6 +459,7 @@ class DataProductMetadata(BaseJsonSchema):
 
         self.load()
         state = self._detect_update_type()
+        self.logger.info(f"Update type {state}")
 
         if state == UpdateType.MinorUpdate:
             new_version = self.generate_next_version_string()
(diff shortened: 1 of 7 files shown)


containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -332,23 +418,71 @@ class to handle creation and updating of
     schema json files relating to data products as a whole
     """
 
-    # returns path of latest metadata or v1 if it doesn't exist
+    MINOR_UPDATE_FIELDS = {
```
murdo-moj: Changing any data product level (non schema level) fields which are editable by the user will trigger a minor version update I think. Looking through the spec, only schema changes would be a breaking change for the user.

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -332,23 +418,71 @@ class to handle creation and updating of
...
+            new_version = self.generate_next_version_string()
+            self.logger.info(f"Creating {self.data_product_name}:{new_version}")
+            new_metadata = DataProductMetadata(
+                data_product_name=self.data_product_name,
+                logger=self.logger,
+                input_data=self.data,
+                version=new_version,
+            )
+            new_metadata.write_json_to_s3()
+            return new_metadata
```
LavMatt: might be personal preference but i kind of liked enforcing the write method to be passed a path, and think i'd find something like below clearer
```suggestion
            new_version_key = (
                DataProductConfig(self.data_product_name).metadata_path(new_version).key
            )
            self.write_json_to_s3(new_version_key)
            return self.data

LavMatt (commented):
This looks good and sets a nice framework i can work into the schema updates (which i'm looking at now), but couple of points:
1. is there a reason we need to distinguish, once we have the update type, whether it's minor or major to write it? For the metadata (and table schema) all that really differs is the path, although with major we would need to follow that up-versioning to the data as well i guess (so maybe that's why actually).
2. Even with minor updates we'll still need to write all associated schema for the data product into the new version folder which isn't currently happening. I wrote this function to do that https://github.com/ministryofjustice/data-platform/blob/main/containers/daap-create-schema/src/var/task/create_schema.py#L18 for schema

containers/daap-python-base/src/var/task/data_product_metadata.py
MatMoore: That makes sense, although I'm assuming some of these are not updatable at all (e.g. name)

https://github.com/ministryofjustice/modernisation-platform-environments/blob/main/terraform/environments/data-platform/data-product-metadata-json-schema/v1.1.0/moj_data_product_metadata_spec.json

The full list is

name
description
domain
dataProductOwner
dataProductOwnerDisplayName
email
status
dpiaRequired
retentionPeriod
dataProductMaintainer
dataProductMaintainerDisplayName
tags
version
id
lastUpdatedDate
creationDate
s3Location
rowCount
schema

So I think I need to add in:
domain
status
dpiaRequired
retentionPeriod
dataProductMaintainer
dataProductMaintainerDisplayName
tags

We also need to make sure that any generated fields are preserved when updating, since they won't be passed into the API call. But since none of that is implemented I think we can skip that for now.

MatMoore:
> This looks good and sets a nice framework i can work into the schema updates (which i'm looking at now), but couple of points:
> 
> 1. is there a reason we need to distinguish, once we have the update type, whether it's minor or major to write it? For the metadata (and table schema) all that really differs is the path, although with major we would need to follow that up-versioning to the data as well i guess (so maybe that's why actually).

No I don't think so, my goal was just to ignore major updates while I implement the minor ones. But as @murdo-moj pointed out, I don't think there are any major updates possible when updating metadata, so actually it's more that we need to distinguish valid and invalid updates (for example changing the name or one of the generated fields).

> 2. Even with minor updates we'll still need to write all associated schema for the data product into the new version folder which isn't currently happening. I wrote this function to do that https://github.com/ministryofjustice/data-platform/blob/main/containers/daap-create-schema/src/var/task/create_schema.py#L18 for schema

Ah thanks. I was planning to implement the actual deletion in the lambda, so that this class only manages the metadata files. Does that make sense to you?

containers/daap-python-base/src/var/task/data_product_metadata.py
MatMoore: Fair enough, I'll just drop that commit then, as this is what it looked like before.

It makes me wonder though if we should lose the `bucket_path` parameter altogether, so that `load` is consistent with `write_json_to_s3` (both are called with a path)?

containers/daap-python-base/src/var/task/data_product_metadata.py
MatMoore: Just pushed a new commit with some changes to the fields

LavMatt:
> Ah thanks. I was planning to implement the actual deletion in the lambda, so that this class only manages the metadata files. Does that make sense to you?

Should there be any deletions? Yeah that does make sense to have the wider collection of metadata and schema managed outside of the class. Suppose we could have a function that copies everything from the previous version folder except files (either data product metadata or table schema) that already exist in the new versions folder. Could live in the paths module possibly

containers/daap-python-base/src/var/task/data_product_metadata.py
LavMatt: as in the get rid of `latest_version_key`? 

I think `load` could maybe be more suitably named, like `load_latest_version` but if there might be any use for loading in older versions then possibly but i can't think of any at the moment, otherwise we'd always want to load the `latest_version_key`

MatMoore:
> > Ah thanks. I was planning to implement the actual deletion in the lambda, so that this class only manages the metadata files. Does that make sense to you?
> 
> Should there be any deletions? Yeah that does make sense to have the wider collection of metadata and schema managed outside of the class. Suppose we could have a function that copies everything from the previous version folder except files (either data product metadata or table schema) that already exist in the new versions folder. Could live in the paths module possibly

Oops sorry I meant copying, not deletion! A function that copies everything might be useful, I think I'll have a go at implementing it in the lambda first and then if I have anything reusable I'll move it.

LavMatt (approved):

PR #2047, opened 2023-10-24
add provider for kms

Emterry:
correcting an oversight - changing provider for KMS alias


jacobwoffenden (changes requested):
Can you restructure to follow the pattern of the rest of the resources that have a provider reference

jacobwoffenden (approved):

PR #2049, opened 2023-10-24
✨ Actions Runner deployment

jacobwoffenden:
This pull request:

- Creates a container for running GitHub Self-Hosted runners
- Creates a new Terraform component for deploying to Cloud Platform
- Creates a Helm chart for deploying GitHub Self-Hosted runners
- Updates create-a-derived-table role to use IRSA

Notes:

- Authentication to GitHub is done using a repo scoped fine grained token from our robot account, this is just used to obtain a self-hosted runner registration token
- There is no de-registration mechanism for a runner because I was unsure how to handle SIGTERM from the kubelet, HOWEVER inactive runners are removed from GitHub after 14 days of being offline
- Ephemeral runners aren't possible because when the runner program exits, Kubernetes just restarts the pod instead of replacing it

TODO:

- [x] Strategy for registering and unregistering runners
  - The token obtained from GitHub's UI is short-lived so when a pod cycles it doesn't come backonline
- [x] Fix path-filter, not all workflows dynamically generate it 😢 
- [x] Add serviceaccount creation permissions to data-platform-production serviceaccount [(link)](https://github.com/ministryofjustice/cloud-platform-environments/blob/main/namespaces/live.cloud-platform.service.justice.gov.uk/data-platform-app-gold-scorecard-form-dev/resources/serviceaccount.tf)
- [x] Remove Actions Runner package from install


Gary-H9 (approved):
LGTM

PR #2053, opened 2023-10-24
Move functionality creating new versions of data product metadata & schemas

MatMoore:
This PR moves a bunch of versioning related code into a new module for use in the update data product endpoint (https://github.com/ministryofjustice/data-platform/issues/1657)

This is mostly moving functionality I added to the metadata module in https://github.com/ministryofjustice/data-platform/pull/2027. We can potentially build on this in https://github.com/ministryofjustice/data-platform/issues/1925.

This separates some class responsibilities:
- DataPlatformMetadata and DataPlatformSchema more focused on reading/writing these files and converting to glue format
- VersionCreator is responsible for determining what is a major/minor change, and updating the metadata bucket accordingly

I've also moved `s3_copy_folder_to_new_folder` from the create_schema lambda, and ensured that is called when updating metadata fields. This is required to preserve the schemas from the previous version.

### DIFF
--- a/containers/daap-python-base/src/var/task/data_product_metadata.py
+++ b/containers/daap-python-base/src/var/task/data_product_metadata.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import json
 import traceback
 from copy import deepcopy
-from typing import Dict, NamedTuple
+from typing import Dict
 
 import boto3
 import botocore
@@ -49,29 +49,6 @@ glue_csv_table_input_template = {
         "Parameters": {"classification": "csv", "skip.header.line.count": "1"},
     },
 }
-
-
-class Version(NamedTuple):
-    """
-    Helper object for manipulating version strings.
-    """
-
-    major: int
-    minor: int
-
-    def __str__(self):
-        return f"v{self.major}.{self.minor}"
-
-    @staticmethod
-    def parse(version_str) -> Version:
-        major, minor = [int(i) for i in version_str.lstrip("v").split(".")]
-        return Version(major, minor)
-
-    def increment_major(self) -> Version:
-        return Version(self.major + 1, self.minor)
-
-    def increment_minor(self) -> Version:
-        return Version(self.major, self.minor + 1)
 
 
 def format_table_schema(glue_schema: dict) -> dict:
@@ -186,16 +163,6 @@ class BaseJsonSchema:
                 f"version 1 of {self.type.value} already exists for this data product"
             )
             return True
-
-    def generate_next_version_string(self, major: bool = False) -> str:
-        """
-        Generate the next version
-        """
-        current_version = Version.parse(self.version)
-        if major:
-            return str(current_version.increment_major())
-        else:
-            return str(current_version.increment_minor())
 
     def validate(
         self,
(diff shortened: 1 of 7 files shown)


containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -495,3 +473,83 @@ def _detect_update_type(self) -> UpdateType:
...
+        s3_copy_folder_to_new_folder(
+            bucket=bucket,
+            source_folder=source_folder,
+            latest_version=latest_version,
+            new_version=new_version,
+            logger=self.logger,
+        )
+
+
+def s3_copy_folder_to_new_folder(
```
MatMoore: This is copied from create_schema. Not sure whether to stick it here or in the paths module?

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -495,3 +473,83 @@ def _detect_update_type(self) -> UpdateType:
...
+
+        state = metadata.detect_update_type()
+        self.logger.info(f"Update type {state}")
+
+        if state != UpdateType.MinorUpdate:
+            raise InvalidUpdate(state)
+
+        latest_version = metadata.version
+        new_version = metadata.generate_next_version_string()
+        self._copy_from_previous_version(
```
MatMoore: This line is new. It copies the whole metadata directory from the previous version, so the `write_json_to_s3` call will then overwrite metadata.json.

This will ensure we inherit all the schemas from the previous version.

containers/daap-python-base/src/var/task/data_product_metadata.py
```diff
@@ -495,3 +473,83 @@ def _detect_update_type(self) -> UpdateType:
     def add_auto_generated_metadata(self):
         """adds key value pairs to metadata that are not given by a user."""
         raise NotImplementedError
+
+
+class VersionCreator:
```
MatMoore: Move this to a separate module. We can build on it in https://github.com/ministryofjustice/data-platform/issues/1925

containers/daap-python-base/tests/unit/data_product_metadata_test.py
```diff
@@ -131,51 +128,23 @@
 }
 
 
-def load_v1_metadata_schema_to_mock_s3(bucket_name, s3_client):
```
MatMoore, line 134: I moved some of this setup into conftest so it can be used in other tests.

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -0,0 +1,169 @@
...
+
+        if state != UpdateType.MinorUpdate:
+            raise InvalidUpdate(state)
+
+        latest_version = metadata.version
+        new_version = str(Version.parse(latest_version).increment_minor())
+        self._copy_from_previous_version(
+            latest_version=latest_version, new_version=new_version
+        )
+        latest_version_key = self.data_product_config.metadata_path(new_version).key
```
LavMatt: Should we call this `new_version_key`, as latest version is used elsewhere and refers to the latest saved version before the update

LavMatt (commented):
looks good to me, one pretty minor comment/observation.

Also, i think it's worth collecting which updates have been made somewhere as even if we aren't using that now will be useful to return somehow to notify consumers of what has changed in a metadata update (is already done in changed fields for metadata, so could just use this actually)

LavMatt (approved):

PR #2057, opened 2023-10-25
Dp 1891

murdo-moj:
- Files are now deleted from landing after copying them to raw
- The destination bucket for the presigned url has been changed from raw to landing


MatMoore (approved):
Nice 👍🏻

PR #2067, opened 2023-10-25
📝 Update tech docs

jacobwoffenden:
- Update CODEOWNERS to use apps-and-tools team
- Update Dependabot to use apps-and-tools team
- Remove path-filter configuration from docs
- Update WoW


julialawrence (approved):
Lgtm

PR #2082, opened 2023-10-26
Dpl 2021 add schema update to versioning

LavMatt:
This PR is change to the base image to allow updates of individual schemas.

It adds:
- `update_schema` and  `schema_update_type` methods to the `VersionCreator` in `versioning`, which write new schema versions to s3 and contains the versioning logic to infer update type based on changes indetified between new and previous versions of schema.
- `detect_column_differences_in_new_version` method to the `DataProductSchema` class in `data_product_metadata`. It detects and returns to differences, of columns only, between new and previous versions of schema.
- `MINOR_UPDATE_SCHEMA_FIELDS` to versioning. This is so keys in the schema that are not nested (like columns), which is all other keys can have their update type inferred based on the some method as metadata updates. Can be added to if we collect more information at the top level of table schema
- `generate_next_version_string` function to `versioning`. This generates the string for the next version based on `UpdateType`


containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -92,7 +94,7 @@ def update_metadata(self, input_data) -> str:
             logger=self.logger,
             input_data=input_data,
         ).load()
-        if not metadata.valid or not metadata.exists:
+        if not metadata.valid:
```
MatMoore: how come this was removed? I think this will break the update metadata endpoint

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -111,6 +113,38 @@ def update_metadata(self, input_data) -> str:
...
+        self._copy_from_previous_version(
+            latest_version=latest_version, new_version=new_version
+        )
+        new_version_key = (
+            DataProductConfig(schema.data_product_name)
+            .schema_path(table_name, new_version)
+            .key
+        )
+        schema.convert_schema_to_glue_table_input_csv()
+        schema.write_json_to_s3(new_version_key)
```
MatMoore: I think you will also need to write a new version of the metadata.json here, since it has the version in, and the list of schemas

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -167,3 +256,16 @@ def s3_copy_folder_to_new_folder(
             Key=destination_key,
             ExtraArgs=s3_security_opts,
         )
+
+
+def generate_next_version_string(
```
MatMoore, line 277: This could be called in the update metadata method as well

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -141,6 +175,61 @@ def metadata_update_type(data_product_metadata) -> UpdateType:
...
+) -> tuple[UpdateType, dict]:
+    """
+    Figure out whether changes to the input data represent a major or minor schema update
+    and return the changes as a dict, e.g.
+        {table_name: {columns:{...}, non_column_fields: {...}}}
+    """
+    if not data_product_schema.exists or not data_product_schema.valid:
+        return UpdateType.NotAllowed, {}
+
+    changed_fields = data_product_schema.changed_fields()
```
MatMoore, line 192: We might not need to use this method, since there is only table_name and columns for the schema, so you could inline the checks here if you prefer

containers/daap-python-base/src/var/task/versioning.py
LavMatt: was not intended, will add back in. Think it may have done it on some update from main conflict resolution

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -141,6 +175,61 @@ def metadata_update_type(data_product_metadata) -> UpdateType:
...
+            ]
+        ):
+            update_type = UpdateType.Unchanged
+
+    if update_type == UpdateType.Unchanged:
+        if not changed_fields:
+            update_type = UpdateType.Unchanged
+        if not changed_fields.difference(MINOR_UPDATE_SCHEMA_FIELDS):
+            update_type = UpdateType.MinorUpdate
+
```
MatMoore: this should be equivalent:

```suggestion
    if update_type == UpdateType.Unchanged and changed_fields and not changed_fields.difference(MINOR_UPDATE_SCHEMA_FIELDS):
        update_type = UpdateType.MinorUpdate
```

not sure I understand the edge case this is handling though. Is there anything that can change other than `columns`?

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -141,6 +175,61 @@ def metadata_update_type(data_product_metadata) -> UpdateType:
...
+    if "columns" in changed_fields:
+        column_changes = data_product_schema.detect_column_differences_in_new_version()
+
+        if any([column_changes["removed_columns"], column_changes["types_changed"]]):
+            update_type = UpdateType.MajorUpdate
+        elif any(
+            [column_changes["added_columns"], column_changes["descriptions_changed"]]
+        ):
+            update_type = UpdateType.MinorUpdate
```
MatMoore, line 202: I might be missing something, but is `update_type` initialised if "columns" is not in changed_fields?

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -141,6 +175,61 @@ def metadata_update_type(data_product_metadata) -> UpdateType:
...
+        elif not any(
```
MatMoore, line 203: I think there should be a branch for column_changes["types_changed"] that is a major change. Then this branch is just an `else` case.

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -141,6 +175,61 @@ def metadata_update_type(data_product_metadata) -> UpdateType:
         return UpdateType.MinorUpdate
 
 
+def schema_update_type(
```
MatMoore, line 181: Can you add more unit tests to cover the various scenarios this function handles? (It's quite a complicated one)

containers/daap-python-base/src/var/task/versioning.py
LavMatt: Possibly due to having the version within the metadata json, but this isn't currently populated as is an autogenerated field which we haven't done any of yet.

I think schemas will be unchanged because this is for updating and is only existing schemas

containers/daap-python-base/src/var/task/versioning.py
LavMatt: Added this in because it's likely data captured at that level for schema could increase and then we can just add into `MINOR_UPDATE_SCHEMA_FIELDS` or even a `MAJOR_UPDATE_SCHEMA_FIELDS`

containers/daap-python-base/src/var/task/versioning.py
LavMatt: Nope don't think you are, that's a bit dodgy

containers/daap-python-base/src/var/task/versioning.py
LavMatt: currently only `tableDescription` although this is almost certain to increase. I'm going to look at this bit again

containers/daap-python-base/src/var/task/versioning.py
LavMatt: covered by `if any([column_changes["removed_columns"], column_changes["types_changed"]])` i believe

containers/daap-python-base/src/var/task/versioning.py
LavMatt: we should probably make sure the version key is added to metadata and schema on write to s3. Maybe a separate issue though

containers/daap-python-base/src/var/task/versioning.py
MatMoore: Ah, missed that due to line wrapping. Ignore me 👍🏻

containers/daap-python-base/src/var/task/versioning.py
MatMoore: Yeah, maybe that should be a separate ticket then?

MatMoore (approved):

PR #2105, opened 2023-10-27
Dp 1846 migrate data product docs

### LINKED ISSUE #1846: Update Metadata product specification document in Data platform repo
### User story
As a user i want to know about the metadata specification allowed in a data product 

### Value / Purpose
By having this the users will have idea of the meta data that needed ( required/ optional for a data product)

### Hypothesis

By having this documentation , we can give it to users for user feedback and also it help data producers to know what are possible metadata they can provide 

### Additional information 

Refactor  and move the product specification from https://github.com/ministryofjustice/data-platform-products/blob/main/data-products/_docs/product-specification.md

to data platform docs

use swagger link and Metadat reserach docs to update the documentation https://docs.google.com/spreadsheets/d/1O1EjO96lyqDbyuzFWU8SNBky3b7tKM8dccZVS7-_t9g/edit#gid=0

tom-webber:
Migrate [Data Platform Product documentation](https://github.com/ministryofjustice/data-platform-products/tree/main/data-products) to Data Platform repo, and update to current and future states of the product definition.

Documentation should describe the metadata requested by the platform, and the purpose of each of the fields. 

Comments in the readme files detail the contents of later iteration metadata ([described in internal documentation](https://dsdmoj.atlassian.net/wiki/spaces/DataPlatform/pages/4561568040/Data+Product+Metadata+Specification))

Non-commented docs can be previewed by running `make preview check` from the `docs/` directory of the repo on a machine running docker.

closes #1846


docs/source/documentation/data-products/example_data_product/data-dictionary.yml
```diff
@@ -6,80 +6,24 @@
...
-      columns:
-        - name: "Offence"
-          description: "Offence group"
-          data-type: "string"
-          column-type: "dimension"
-        - name: "Description"
-          description: "Offence group description"
-          data-type: "string"
-          column-type: "supplementary"
+# Add a description for your collection to help users find it.
```
MatMoore, line 9: Is this example still useful, now that we have the swagger documentation? It seems like that would also provide examples of this format, and as user facing documentation it is more likely to stay up to date.

There is also a mismatch between this file being YAML, vs the actual files we store being JSON.

jemnery (dismissed):

docs/source/documentation/index.html.md.erb
```diff
@@ -35,6 +35,10 @@ This repository holds the Ministry of Justice’s Data Platforms architecture an
 
 - [Terraform](/documentation/platform/infrastructure/terraform.html)
 
+## Products
```
MatMoore, line 40: Should the overview section above be replaced with a link to https://data-platform.service.justice.gov.uk/about/ ?

docs/source/documentation/data-products/index.html.md.erb
```diff
@@ -0,0 +1,68 @@
...
+
+> **IMPORTANT**
+
+> The structure and usage of these metadata definitions are likely to change substantially throughout development
+
+A Data Product will have a unique name, and is defined in code: YAML or JSON
+(we show YAML in these docs, as it is a more human-readable format).
+The code representation of the Data Product is referred to as Data Product Metadata (data *about* the data), and is currently captured in a single file.
+
+This file contains the following sections:
```
MatMoore, line 55: I'm finding this part confusing because the current metadata schema doesn't distinguish between these sections. Is this outlining plans for how we plan to evolve it in the future?

If so I think we should clearly mark out what's implemented and what isn't, like you have below with "Sections of metadata that we plan to capture within a Data Product, but are not yet implemented:"

IMO the source of truth for how to define a data product at the moment are these two directories:
- https://github.com/ministryofjustice/modernisation-platform-environments/tree/main/terraform/environments/data-platform/data-product-metadata-json-schema
- https://github.com/ministryofjustice/modernisation-platform-environments/tree/main/terraform/environments/data-platform/data-product-table-schema-json-schema/

Is it possible with the tech docs template to directly embed those in the docs, with syntax highlighting? And if not, could we just link to the files in github?

MatMoore (dismissed):
I think this is definitely an improvement from before, so happy to merge as is.

Reading through it though I'm worried there is a lot of potential for confusion, due to the varying sources of information about DaaP that will be publicly available.

As far as I'm aware, we will have:

- The front door for the service https://data-platform.service.justice.gov.uk/about/
- The API documentation https://hsolkci589.execute-api.eu-west-2.amazonaws.com/development/docs/  (needs republishing I think)
- Metadata and JSON schemas (which get incorporated into the API documentation) https://github.com/ministryofjustice/modernisation-platform-environments/tree/main/terraform/environments/data-platform
- This one, which lives at https://technical-documentation.data-platform.service.justice.gov.uk/

Since this repo and it's docs are public, I would suggest we very clearly mark this site as being intended for an internal audience (preferably with a big banner at the top) with a link to the front door to redirect any users who wind up on the wrong site.

I'm not sure there is even an a need for us to describe the metadata definitions here, since all this information should be in the API documentation now(?)

docs/source/documentation/data-products/index.html.md.erb
tom-webber: Yes this is a proposal for how we evolve the schema in the future, see the [internal confluence doc on the proposed evolution](https://dsdmoj.atlassian.net/wiki/spaces/DataPlatform/pages/4561568040/Data+Product+Metadata+Specification).

Those files are great, and I didn't realise I'd missed so many fields that we currently accept in the Data Product metadata (I only listed the required). I'd want these docs to be a more human-readable, and beginner-friendly version of those files. It'd be worth checking out if we can convert those to yaml and present the 'properties' field

tom-webber:
> I'm not sure there is even an a need for us to describe the metadata definitions here, since all this information should be in the API documentation now(?)

@MatMoore The purpose of the documentation here is to provide a more gentle/structured description and walkthrough of our data for non-expert / less experienced users - ones who might be intimidated by autogenerated API docs, or parsing raw JSON specs.

docs/source/documentation/index.html.md.erb
tom-webber: most likely yes, but I'll leave that for now

MatMoore (approved):
@tom-webber Ah ok, I didn't realise these were for that audience as well.

I guess my main point was just that we should make sure that (eventually) the end user documentation is joined up together as one resource rather than being split up across multiple URLs. Even for technical API documentation I would expect that if any developers/architects outside the data platform team end up using it then we would need some introductory material and guides in addition to the autogenerated reference material.

Btw I don't know if you saw already but there's another issue here regarding consolidating internal docs: https://github.com/ministryofjustice/data-platform/issues/2106

PR #2108, opened 2023-10-27
created lambda for querying data

PriyaBasker23:
As a user, I want to confirm the successful creation of my data product. This feature eliminates the need for users to possess AWS accounts or tools in order to access and view the data.

The code is designed to execute a query and provide the results in a tabular format containing 10 rows


containers/daap-query-data/src/var/task/query_data.py
```diff
@@ -0,0 +1,124 @@
...
+        logger:DataPlatformLogger
+    ) -> str:
+        """
+        runs query for given sql and waits for completion
+        """
+        try:
+            res = athena_client.start_query_execution(
+                QueryString=sql,
+                QueryExecutionContext={"Database": database_name},
+                WorkGroup="data_product_workgroup",
```
MatMoore: I wonder if we should also store this constant in the paths library. Not sure if it's gonna change at some point.

containers/daap-query-data/src/var/task/query_data.py
```diff
@@ -0,0 +1,124 @@
...
+def format_athena_results_to_table(results):
+    # Format the results into a table string
+    if not results:
+        return "No data to display"
+
+    table = ''
+
+    # Data rows
+    headers= [value for header_dict in results['ResultSet']['Rows'][0]['Data'] for value in header_dict.values()]
+    print(headers)
```
MatMoore: [minor] I don't think this print should still be here

containers/daap-query-data/src/var/task/query_data.py
```diff
@@ -0,0 +1,124 @@
...
+
+def format_athena_results_to_table(results):
+    # Format the results into a table string
+    if not results:
+        return "No data to display"
+
+    table = ''
+
+    # Data rows
+    headers= [value for header_dict in results['ResultSet']['Rows'][0]['Data'] for value in header_dict.values()]
```
MatMoore: Might need a check here to exit early if there are no rows in the result set. I don't think this is being caught by the `if not results` above.

containers/daap-query-data/src/var/task/query_data.py
```diff
@@ -0,0 +1,124 @@
...
+    data_result=[]
+    for row in results['ResultSet']['Rows'][1:]:
+        data = [value for data_dict in row['Data'] for value in data_dict.values()]
+        data_result.append({key: value for key, value in zip(headers, data)})
+
+    # Calculate the maximum width for each column based on header and data values
+    for row in data_result:
+        for header in headers:
+            value = row[header]
+            column_widths[header] = max(column_widths[header], len(str(value)))
```
MatMoore, line 76: Suggestion: At the moment this will expand to fit the content, but I'm wondering if we should also have a max width before we truncate the data, since a dataset could have fields with really long values, like a long string of text.

containers/daap-query-data/src/var/task/query_data.py
```diff
@@ -0,0 +1,124 @@
...
+    logger = DataPlatformLogger(
+        data_product_name=data_product_name,
+        extra={
+            "image_version": os.getenv("VERSION", "unknown"),
+            "base_image_version": os.getenv("BASE_VERSION", "unknown"),
+            "table_name": table_name,
+        },
+    )
+
+    query = f'SELECT * FROM "{data_product_name}"."{table_name}" LIMIT 10'
```
MatMoore, line 108: Should we also order by execution timestamp descending? Doesn't really matter for now, I'm just thinking ahead to partial loads

containers/daap-query-data/tests/unit/mock_test.py
```diff
@@ -0,0 +1,65 @@
+from query_data import handler
+from unittest.mock import Mock
+import pytest
+import boto3
+
+
+def test_query_athena(athena_client, fake_context):
```
MatMoore: it would be good to have a test for an empty result set as well, to make sure the lambda handles both

MatMoore (commented):
I've left a few suggestions inline but this looks really nice 👍🏻

containers/daap-query-data/src/var/task/query_data.py
LavMatt: I think the query should be for latest partition, as the use case was for someone to check the data they just loaded had flowed through, so add
`where load_timestamp = (select max(load_timestamp) from {data_product_name}.{table_name})`

containers/daap-query-data/src/var/task/query_data.py
PriyaBasker23: WorkGroup="data_product_workgroup" is coded in few of our lambda. i will add a ticket to make as constant in seperate ticket

containers/daap-query-data/src/var/task/query_data.py
PriyaBasker23: In Athena - Select picks the latest partition by default isnt it

LavMatt: No it doesn't, need to specify in query

PriyaBasker23: In Amazon Athena, when you use a query like SELECT * FROM table_name, Athena will return data from the latest partition, assuming the table you're querying is a partitioned table and it's set up to use partitions based on a date or other criteria. If you have a table with partitions based on time (e.g., date), and you don't specify a specific partition in your query, Athena will query the partition with the latest data.

containers/daap-query-data/src/var/task/query_data.py
PriyaBasker23: I dont want to truncated the results rather wrapping would be better. But it gets complicated with wrapping  text with adding spacing for before and after columns. So will leave it like this to adjust width based on column values

MatMoore (dismissed):

MatMoore (approved):

PR #2113, opened 2023-10-30
Dpl 2021 daap update schema

LavMatt:
adds new lambda container for endpoint to update table schema of a data product, the endpoint path will be the PUT method of `/data-product/{data-product-name}/table/{table-name}/schema`


containers/daap-update-schema/src/var/task/update_schema.py
```diff
@@ -0,0 +1,69 @@
...
+    except InvalidUpdate as exception:
+        logger.error("Unable to update the data product", exc_info=exception)
+        return format_error_response(
+            response_code=HTTPStatus.BAD_REQUEST,
+            event=event,
+            message="Update not allowed",
+        )
+
+    return format_response_json(
+        status_code=HTTPStatus.OK, body={"version": new_version, "changes": changes}
```
MatMoore, line 68: I'm wondering if we should return the full schema here rather than just the changes. Not sure at this point what would be more useful for the API users.

However, this way does make it easy to understand why a version update was major or minor, and it'll definitely be useful for notifications.

MatMoore (approved):
Looks good to me 👍🏻

containers/daap-update-schema/src/var/task/update_schema.py
LavMatt: I think from a user perspective it might be better to return changes. Say if they have a large table with lots of columns and change a couple of things. This would be a clearer sense check that they changed the right columns opposed to returning full schema

PR #2132, opened 2023-10-30
:construction: Delete table metadata versioning

### LINKED ISSUE #1662: API end point : Remove Table Data and Schema

mitchdawson1982:
This PR moves much of the existing delete table code out into the base package.


containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -198,25 +199,68 @@ def extract_timestamp_from_curated_path(string: str):
...
+    for version in data_product_versions:
+        element_prefixes.append(
+            f"{path_prefix}/{data_product_name}/{version}/{table_name}/"
+        )
+
+    return element_prefixes
+
+
+def delete_all_element_version_data_files(data_product_name: str, table_name: str):
+    """Deletes raw and curated data for all element versions"""
```
LavMatt: To me this doesn't so much feel like a pathing related function but more a versioning related function

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -198,25 +199,68 @@ def extract_timestamp_from_curated_path(string: str):
...
+    # Proceed to delete the raw data
+    element = DataProductElement.load(
+        element_name=table_name, data_product_name=data_product_name
+    )
+    raw_prefixes = generate_all_element_version_prefixes(
```
LavMatt: this returns all versions as per the metadata, and that won;t match to the data, e.g. you won;t have 1.1, 1.2, but only major versions for data

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -198,25 +199,68 @@ def extract_timestamp_from_curated_path(string: str):
...
+        "raw", data_product_name, table_name
+    )
+    curated_prefixes = generate_all_element_version_prefixes(
```
LavMatt: same for this. Maybe we could just infer all major versions from what's returned

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -198,25 +199,68 @@ def extract_timestamp_from_curated_path(string: str):
...
+        "curated", data_product_name, table_name
+    )
+
+    s3_recursive_delete(element.data_product.raw_data_bucket, raw_prefixes)
```
LavMatt: Seems to me slightly conflicting behaviour deleting from an existing version without any upversioning of database but still upversioning metadata. This is where a user no longer wants any of the data available in any versions? I'm just wondering how if we have this and we have a `remove a schema` endpoint how a user would have to careful in choosing which one to use. Whereas `remove a schema` would remove it from say a v2 onwards.

Later we're creating a new major version metadata but not a new version of the data. Having reread #1662 it says both to delete raw and curated which is happening but also to not delete raw.

I'm wondering if it was expected to have a separate delete schema/table with versioning endpoint or if this is to be the only delete

LavMatt (commented):
This mostly looks good Mitch.

It will delete data and major upversion metadata but (and it may be me not following the intention of the ticket) i'm not 100% convinced we should be deleting data from within versions, e.g. data_product_v1 has tbl1 and tbl2 and then after calling this endpoint to delete tbl2, the same data_product_v1 only has tbl1.

And not sure having data product data version and metadata versions misalign is correct.

Another observation i made about getting all versions to delete data using versions from metadata, which will return all minor versions when data are versioned at major level only.

containers/daap-python-base/src/var/task/data_platform_paths.py
mitchdawson1982: Agreed that this will be revisited once data versioning has been implemented.

containers/daap-python-base/src/var/task/data_platform_paths.py
mitchdawson1982: Agreed that this will be revisited once data versioning has been implemented.

containers/daap-python-base/src/var/task/data_platform_paths.py
mitchdawson1982: Agreed that this will be revisited once data and database versioning has been implemented.

containers/daap-python-base/src/var/task/data_platform_paths.py
mitchdawson1982: Function now moved to the versioning module.

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -198,25 +199,51 @@ def extract_timestamp_from_curated_path(string: str):
...
+
+    for version in data_product_versions:
+        element_prefixes.append(
+            f"{path_prefix}/{data_product_name}/{version}/{table_name}/"
+        )
+
+    return element_prefixes
+
+
+def s3_recursive_delete(bucket_name: str, prefixes: list[str]) -> None:
```
LavMatt: You think this would be better in versioning as well?

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -8,10 +8,15 @@
 
 import boto3
 from data_platform_logging import DataPlatformLogger, s3_security_opts
-from data_platform_paths import DataProductConfig
+from data_platform_paths import (
+    DataProductConfig,
+    DataProductElement,
+    delete_all_element_version_data_files,
```
LavMatt: Don't want this import anymore as it's in versioning now

LavMatt (commented):
couple very minor/pedantic comments

containers/daap-python-base/src/var/task/versioning.py
mitchdawson1982: Removed

containers/daap-python-base/src/var/task/data_platform_paths.py
mitchdawson1982: Agreed, now moved to the versioning module.

LavMatt (approved):

PR #2156, opened 2023-10-31
Dp 1797 update athena_load to use existing schema for curated table

### LINKED ISSUE #1797: Use schema generated via api endpoint to create raw table in athena load

tom-webber:
resolves #1797:

* `athena_load` process uses type and comment/description (trimmed where necessary) data from the registered schema to create the curated parquet files. Be confident metadata registered in the glue catalog matches that in the s3 schema.
* data product database names are suffixed with their major version suffix.
* create tests for any additional functions required.


tom-webber:
Note that the truncating of the glue schema column comments and table description are currently **lossy** for large descriptions as these functions are manipulating the data that is saved. Future work will save a non-glue version of the metadata, so this implementation here will not be lossy in that case.

containers/daap-athena-load/src/var/task/athena_load_handler.py
```diff
@@ -26,12 +27,15 @@ def handler(
     extraction = RawDataExtraction.parse_from_uri(full_s3_path)
     data_product_element = extraction.element
 
+    data_product_name = data_product_element.curated_data_table.database
```
LavMatt: We now need to distinguish between the data_product_name and the database name that has a version suffix, otherwise you'll get a schema doesn;t exist error, e.g. looking for `{data_product_name}_v1`
```suggestion
    data_product_name = data_product_element.data_product.name
    database_name = data_product_element.curated_data_table.database

File: containers/daap-athena-load/src/var/task/athena_load_handler.py [hunk 2]
```diff
@@ -41,6 +45,23 @@ def handler(
         extraction.path, data_product_element, logger=logger
     )
 
+    try:
+        table_schema = (
```
LavMatt, line 49: what i've suggested does mean we might need to add another attribute to `DataProductSchema` or at least have it populate DatabaseName with the data_product_name+version suffix

could make sure that key's value in the schema includes the suffix in `DataProductSchema.convert_schema_to_glue_table_input_csv()`

LavMatt (commented):
Couple of minor suggestions re. the suffix as it will fail to find the schema as is. We don't save the data product with a version suffix, that's only for database name

containers/daap-athena-load/src/var/task/athena_load_handler.py
tom-webber: Good point.

I've added `database_name` to the `BaseJsonSchema` as it can be calculated there. It's then inherited into `DataProductMetadata` and `DataProductSchema` classes. 

This is used in the convert method for `DatabaseName`

LavMatt (approved):
Looks good now

PR #2176, opened 2023-11-02
add functionality to `ministryofjustice-data-platform-catalogue` for metadata/schema pushes to openmetadata

### LINKED ISSUE #104: Implement Data catalogue ingestion in the Lambdas  using the class created
### User Story
As a data producer / controller I want to see the metadata of a data product in the data catalogue and this should be automated at the point at which the data product is created.

### Value / Purpose
By having this functionality we will have metadata imported into the catalogue throughout the user journey and streamlined by using metadata api for ingestion.

Image

### Hypothesis
We will use the api for ingestion within our lambda functions to enrich the catalogue with metadata when the data product is created.

### Checklist 

- [ ] Run through user journey of registering a data product and call the functions for creating the dataproduct( Database) , Schema ( tables ) etc 
       - [ ] Test the output in open metadata 
- [ ] Run through the user journey of updating the data product  and call the specific functions needed
         - [ ] Test the output in open metadata . Explore more on the functionality and document it.  This is not been explored yet 

### Notes
- Dependent on existing ticket to create open metadata wrapper class
- Dependent on versioning work
- Dependent on defining the user journeys

LavMatt:
This PR makes the following additions or updates to `ministryofjustice-data-platform-catalogue`:

### entities.py
- `DataProductMetadata.from_data_product_metadata_dict()`  and `TableMetadata.from_data_product_schema_dict()` methods added to create respective objects being passed in data product metadata/schema

### client.py
- `CatalogueClient.get_user_id` added. It takes a user email and returns the user id saved in open metadata.
- `CatalogueClient.create_or_update_table` changed so column description is also passed to openmetadata

These updates will make integrating pushes of data product metadata and table schema from lambda handlers for creating metadata and schema easier

### DIFF
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
@@ -234,5 +234,8 @@ class CatalogueClient:
             for tag in tags
         ]
 
-    def _generate_duration(self, duration_in_days: int):
-        return Duration.parse_obj(f"P{duration_in_days}D")
+    def _generate_duration(self, duration_in_days: int | None):
+        if duration_in_days is None:
+            return None
+        else:
+            return Duration.parse_obj(f"P{duration_in_days}D")
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
@@ -154,7 +154,7 @@ class CatalogueClient:
         """
         Define a table.
         There can be many tables per data product.
-        columns are extpected to be a list of dicts in the format
+        columns are expected to be a list of dicts in the format
             {"name": "column1", "type": "string", "description": "just an example"}
         """
         columns = [
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
@@ -2,9 +2,6 @@ import json
 import logging
 from http import HTTPStatus
 
-from metadata.generated.schema.api.createEventPublisherJob import (
-    CreateEventPublisherJob,
-)
 from metadata.generated.schema.api.data.createDatabase import CreateDatabaseRequest
 from metadata.generated.schema.api.data.createDatabaseSchema import (
     CreateDatabaseSchemaRequest,
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
@@ -2,6 +2,9 @@ import json
 import logging
 from http import HTTPStatus
 
+from metadata.generated.schema.api.createEventPublisherJob import (
+    CreateEventPublisherJob,
+)
 from metadata.generated.schema.api.data.createDatabase import CreateDatabaseRequest
 from metadata.generated.schema.api.data.createDatabaseSchema import (
     CreateDatabaseSchemaRequest,
@@ -15,6 +18,7 @@ from metadata.generated.schema.entity.services.connections.metadata.openMetadata
 from metadata.generated.schema.entity.services.databaseService import (
     DatabaseServiceType,
 )
+from metadata.generated.schema.entity.teams.user import User
 from metadata.generated.schema.security.client.openMetadataJWTClientConfig import (
     OpenMetadataJWTClientConfig,
 )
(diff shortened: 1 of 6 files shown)


python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
```diff
@@ -183,6 +190,15 @@ def _create_or_update_entity(self, data) -> str:
 
         return response.dict()["fullyQualifiedName"]
 
+    def get_user_id(self, user_email: str):
+        """
+        returns the user id from openmetadata when given a user's email
+        """
+        username = user_email.split("@")[0]
+        user_id = self.metadata.get_by_name(entity=User, fqn=username).id
```
MatMoore, line 198: We'll need to make sure we set up the name in the same way when we do the control panel bit that provisions users the first time they log into the platform.

I can see this potentially going wrong if people mix up justice emails and other email domains that are in use across the department. Even if people have both they could be in a different format. I wonder if we should just enforce that a user email always has to end in justice.gov.uk?

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -20,11 +21,60 @@ class DataProductMetadata:
...
+        Expects a dict containing data product metatdata information as per the
+        required fields in the json schema at
+        https://github.com/ministryofjustice/modernisation-platform-environments/tree/main/terraform/environments/data-platform/data-product-metadata-json-schema
+
+        and should be pass version and owner id as in openmetadata.
+
+        Then populates a DataProductMetadata object with the given data.
+        """
+        new_metadata = DataProductMetadata(
+            name=metadata.get("name", "Not provided"),
```
MatMoore: I think it would be better to either pass `None` or an actual value here instead of introducing a special value. It'll clutter up the UI and make our lives more difficult if we need to fetch the metadata out of OpenMetadata at any point.

You can change the `DataProductMetadata` class to make everything optional if you prefer, I think I was assuming that the update metadata call would always have a complete dictionary, but now I'm doubting that assumption.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -20,11 +21,60 @@ class DataProductMetadata:
     dpia_required: bool
     tags: list[str] = field(default_factory=list)
 
+    @staticmethod
+    def from_data_product_metadata_dict(metadata: dict, version, owner_id: str):
```
MatMoore, line 25: 👍🏻 this looks useful

python-libraries/data-platform-catalogue/pyproject.toml
```diff
@@ -9,7 +9,9 @@ packages = [{ include = "data_platform_catalogue" }]
 
 [tool.poetry.dependencies]
 python = "^3.10.0"
-openmetadata-ingestion = "^1.1.7.0"
+# pinned to this version as was installing a later vesion and giving errors, seems a breaking
+# change introduced inadvertently
+openmetadata-ingestion = "1.1.7.0"
```
MatMoore, line 14: annoying... seems like the library is changing quite fast at the moment though so hopefully the problem goes away

MatMoore (commented):
LGTM, just would rather avoid the special values for missing fields. As far as I know we should be able to pass None through to the API instead

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
LavMatt: yeah that would be cleaner, i'll change

python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
LavMatt: The username in openmetadata is aways what ever comes before the `@` but we are using azure to authenticate so it'll always be justice.gov.uk (i think anyway), and we can't be sure peoples names will always follow the same spelling or format. Maybe we should enforce all data providers to use an @justice.gov.uk email, but one to raise for further discussion

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
LavMatt: in fact we would always expect there to be most of that data available, as we're enforcing it being supplied at collection. I'll only use `.get` for non required fields

python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
MatMoore: Yeah, let's run with this assumption for now 👍🏻

MatMoore (approved):
:shipit:

PR #2177, opened 2023-11-02
Update Data Product tech docs

tom-webber:
* add versioning documentation
* update spec and governance docs with optional implemented fields 
* add erb headers for previously added doc pages


docs/source/documentation/data-products/components/product-governance.html.md.erb
```diff
@@ -15,7 +22,18 @@ domain: HMPPS
...
+
+including optional fields:
+```yaml
+domain: HMPPS
+dataProductOwner: data.product.owner.name@justice.gov.uk
+dataProductOwnerDisplayName: Data Product Owner
+dataProductMaintainer: data.product.maintainer.name@justice.gov.uk
+dataProductMaintainerDisplayName: Data Product Maintainer
+dpiaRequired: false
+dpiaLocation: s3://data-platform-data/civil-courts-data/v1/
```
MatMoore, line 36: Did we already decide how the DPIA linkage will work? Just wondering what the expectations are around this, e.g. is it expected to be a public URL, or is it not guaranteed we'll be able to even access it? Are these documents in a standard format?

docs/source/documentation/data-products/components/product-specification.html.md.erb
```diff
@@ -15,9 +22,18 @@ You must also provide contact information for the product.
...
+email: data.product.contact@justice.gov.uk
 ```
 
+including optional fields:
+```yaml
+name: example_data_product
+description: Example Data Product contains published prison population from 2001 to present
+status: production
+email: data.product.contact@justice.gov.uk
+tags: {"Sandbox": "True"}
```
MatMoore, line 41: I have a question these docs should probably clarify: is there a predefined list of tags you can use, or does the data owner have free reign to come up with their own tagging conventions?

I ask because OpenMetadata's concept of a tag forces you to predefine "classifications" before you can use them, and there are a few that already exist, e.g. for specifying environments etc. If we don't have a fixed list of valid tag names, then data owners will diverge and the tags won't be very useful to consumers who are browsing via the catalogue and want to filter in some way.

We don't have to implement our tags using OpenMetadata tags, (we could use custom fields) but if we want to take advantage of it in the UI then implementing it their way would be simplest.

docs/source/documentation/data-products/components/product-specification.html.md.erb
```diff
@@ -96,7 +112,9 @@ The product `title` should be the human-readable version of the name. Keep it sh
 
 The product `description` can be longer, give more information and will help users find your data.
 
-> NOTE: the `product-version` attribute is not yet documented here. This will be automatically incremented when the Data Product is updated. Versioning logic will be clearly documented, but in brief breaking changes will constitute Major version bumps (`v1.0` -> `v2.0`), where backwards-compatible changes and updates will be Minor version bumps (`v1.0` -> `v1.1`).
+> NOTE: the `product-version` attribute is not yet documented here. This will be automatically incremented when the Data Product is updated. [Versioning logic is documented here](/documentation/data-products/index.html#altering-a-data-product), but in brief breaking changes will constitute Major version bumps (`v1.0` -> `v2.0`), where backwards-compatible changes and updates will be Minor version bumps (`v1.0` -> `v1.1`).
```
MatMoore: (minor copy suggestion)

```suggestion
> NOTE: the `product-version` attribute is not yet documented here. This will be automatically incremented when the Data Product is updated. [Versioning logic is documented here](/documentation/data-products/index.html#altering-a-data-product), but in brief, breaking changes will constitute Major version bumps (`v1.0` -> `v2.0`), whereas backwards-compatible changes and updates will be Minor version bumps (`v1.0` -> `v1.1`).
```

docs/source/documentation/data-products/components/product-specification.html.md.erb
```diff
@@ -140,6 +158,11 @@ Most Data Products will be tagged as `production` - the other statuses will flag
 <!--
 - `allow-derived-products` - typically `true` (and `true` if not supplied), but in some cases you may not wish derived products to be created. If the product status attribute is "production", you may be asked to justify disallowing derived products as we want to encourage data reuse. -->
 
+<!-- (Alpha) -->
+### Contact information
+
+You must supply an `email` address to indicate the point of contact between consumers and maintainers of the Data Product. It could be the owner or a distribution list, but must be reliable and responsive.
```
MatMoore: This is in contrast to the owner email, which we're assuming is the email that matches the one the data owner signs in with (@justice.gov.uk).

Should make sure this is clear as people might wonder why they have to provide two emails.

docs/source/documentation/data-products/index.html.md.erb
```diff
@@ -69,3 +69,17 @@ Sections of metadata that we plan to capture within a Data Product, but are not
...
 
 [Example Data Product](example_data_product/index.html)
+
+## Using a Data Product
+
+### Altering a Data Product
+
+Over time you may need to update your Data Product. This may come though adding data assets to the Product, removing depricated data assets, changes to the data assets themselves (such as column changes or additions in structured data), or altering the descriptive metadata of the Data Product itself.
+
+To communicate these changes in the Data Product to your users, and to maintain ongoing compatibility, changes to your Data Product (or the data assets associated with it) will automatically increment the version of your Data Product.
```
MatMoore: This is really clear I think 👍🏻

MatMoore (dismissed):
Nice 👍🏻

docs/source/documentation/data-products/components/product-governance.html.md.erb
tom-webber: This likely won't be relevant for a while, as DPIAs are currently stored/managed on a different platform that we don't have good access to. I don't know the format.

docs/source/documentation/data-products/components/product-specification.html.md.erb
tom-webber: I had this discussion with @jemnery - we settled on user-specified tags (free text), though this was not considering a link with OpenMetadata's tags. 

Can we link existing free-text tags with OpenMetadata and then have non-existing tags fall into a custom field? 
E.g. 
- if all user-specified tags exist in OpenMetadata, they all match to the OpenMetadata tags
- if no user-specified tags exist in OpenMetadata, they all go into a custom field
- if some user-specified tags exist in OpenMetadata, the matching tags become OpenMetadata tags, the others go into a custom field

MatMoore: I think something like that should be possible, let's capture in another ticket.

MatMoore (approved):

docs/source/documentation/data-products/components/product-specification.html.md.erb
MatMoore: I've moved this proposal to https://github.com/ministryofjustice/data-platform/issues/2249

PR #2205, opened 2023-11-03
Check csv for issues even if schema is not available

MatMoore:
If the schema is available, we can validate column types. However, even if we don't have a schema, we need to check that the csv looks parsable and does not contain embedded newlines.


LavMatt (approved):
looks good to me

PR #2211, opened 2023-11-03
Add data product name to the logger and update to latest base image in reload/resync lambdas

MatMoore:
https://github.com/ministryofjustice/data-platform/issues/2193

I've included the fix for landing-to-raw in https://github.com/ministryofjustice/data-platform/pull/2205


containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py
```diff
@@ -95,45 +93,23 @@ def get_data_product_pages(bucket, data_product_prefix, s3_client=s3) -> PageIte
...
-    example key: `key = "raw_data/data_product/table/extraction_timestamp=timestamp/file.csv`
+    example key: `key = "raw/data_product/v1.0/table/load_timestamp=timestamp/file.csv`
     size > 0 because sometimes empty directories get listed in contents
     """
     filtered_pages = pages.search("Contents[?Size > `0`][]")
     result_set = set("/".join(item["Key"].split("/")[1:-1]) for item in filtered_pages)
     return result_set
 
 
-def get_curated_unique_extraction_timestamps(curated_pages: PageIterator) -> set:
```
MatMoore, line 110: we don't need this anymore since the raw/curated path formats are the same

LavMatt (commented):
in `containers/daap-resync-unprocessed-files/src/var/task/resync_unprocessed_files.py` lines 36 and 37 
```
    raw_prefix = data_product.raw_data_prefix
    curated_prefix = data_product.curated_data_prefix
```
a `BucketPath` object is returned. Need the `key` attribute of these i believe. Not an issue introduced through this PR however.

Is also getting the latest version only, this might not matter though

MatMoore:
@LavMatt well spotted... just pushed up a fix. Regarding it only operating on the latest version, maybe that's fine for now, as if there's been a new version created, and the old version is missing data, then most likely nobody has ever used that data and nobody will use it in the future either.

LavMatt (approved):

LavMatt (approved):

PR #2252, opened 2023-11-07
Switch to structured logging with Structlog

MatMoore:
## Problem
Our lambdas log in a plain text format, and only send the full JSON to s3. This means that cloudwatch doesn't have the extra fields like `data_product_name`, so we can't filter very easily or count events for specific data products/tables.

For https://github.com/ministryofjustice/data-platform/issues/2034 I am investigating surfacing these logs in Grafana, so would like to send the full structured data to cloudwatch.

Note that it is possible to query the data in S3 via Athena, instead of visualising the data in cloudwatch but we don't have this configured yet, and I'm not sure if it adds additional cost.

**I also noticed that the process of writing logs to s3 may lose events**, because there is no way to append lines to objects in s3, so we're actually reading the whole file for the current (date, data product, lambda) combination, appending in python, and then writing back to s3. This assumes that the same lambda will never be called concurrently for the same data product, but this is not going to be the case when uploading multiple tables for the same product.

## Description
I've added [structlog](https://www.structlog.org/en/stable/index.html) as a drop in replacement for logging. This library was recommended in the lambda docs: https://docs.aws.amazon.com/lambda/latest/operatorguide/parse-logs.html

I've configured it to output logs as JSON instead of unstructured stdout. This means we now send the same structured data to cloudwatch as well as s3.

## Notes
- We could configure this differently in development - the library ships with a formatter that pretty prints the structured data. But for now I wanted to keep it simple.
- I've intentionally not made any changes to the s3 logging (except to wire it up to structlog's processor chain)

### DIFF
--- a/containers/daap-python-base/CHANGELOG.md
+++ b/containers/daap-python-base/CHANGELOG.md
@@ -9,6 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [6.1.0] - 2023-11-08
+
+### Changed
+
+- DataPlatformLogger now outputs JSON in the stdout logs, so that all the extra
+  data attached to a log record is preserved, and can be parsed and filtered
+  in CloudWatch.
+- The JSON written to S3 is unchanged.
+
 ## [6.0.0] - 2023-11-07
 
 ### Changed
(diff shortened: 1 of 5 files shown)


mitchdawson1982 (dismissed):
LGTM

mitchdawson1982 (approved):
LGTM

PR #2256, opened 2023-11-07
✨ Add `ONBUILD` to DaaP Python Base

jacobwoffenden:
This pull request:

- Adds an [`ONBUILD`](https://docs.docker.com/engine/reference/builder/#onbuild) instruction so child images run a `yum update --security` before proceeding, this will pull in any patches for system packages
- Fixes a typo in `LAMBDA_TASK_ROOT`
- Bumps pip to `23.3.1`


tom-webber (approved):
Looks like a good suggestion to me

PR #2263, opened 2023-11-08
Dpl delete data product

### LINKED ISSUE #2030: API end point : Delete Data product
## User story
As a data producer
I need to deleted unwanted data products
So that I'm not responsible for them forever

## Value/Purpose
- We want the data platform to be well-governed, so we don't want to accumulate old data products that nobody owns any more
- Data producers might experiment with new data products but not want to put them live. We don't want to keep data products hanging around that aren't useful.

## Description

When a data product is deleted all metadata and data should be gone. For alpha this means non-reversable, hard deletion.

- This should do everything that the delete schema endpoint does, and also delete the metadata.
- It should remove the data product from the catalogue
- It should remove all traces of prior versions from s3
- It will eventually notify users, but this is out of scope for this ticket

Image

## API design
The new endpoint is DELETE data-product/{data-product-name}

See https://dsdmoj.atlassian.net/wiki/spaces/DataPlatform/pages/4511991882/Initial+APIs

## Definition of done
See https://dsdmoj.atlassian.net/wiki/spaces/DataPlatform/pages/4511334405/API+endpoint+creation+checklist

- [ ] Added new lambda
- [ ] Added new endpoint to API gateway
- [ ] Add swagger docs

mitchdawson1982:


containers/daap-delete-data-product/src/var/task/delete_data_product.py
```diff
@@ -0,0 +1,61 @@
...
+from data_product_metadata import DataProductMetadata
+from versioning import VersionCreator
+
+logger = DataPlatformLogger(
+    extra={
+        "image_version": os.getenv("VERSION", "unknown"),
+        "base_image_version": os.getenv("BASE_VERSION", "unknown"),
+    }
+)
+
```
MatMoore:
```suggestion
logger = DataPlatformLogger()

```

As of 6.2.0 of the base, these versions are added automatically

containers/daap-delete-data-product/src/var/task/delete_data_product.py
```diff
@@ -0,0 +1,61 @@
...
+        request data.
+    :param context: The context in which the function is called.
+    :return: A response that is sent to Amazon API Gateway, to be wrapped into
+             an HTTP response. The 'statusCode' field is the HTTP status code
+             and the 'body' field is the body of the response.
+    """
+
+    data_product_name = event["pathParameters"].get("data-product-name")
+
+    logger.add_extras({"data_product_name": data_product_name})
```
MatMoore:
```suggestion
    logger.add_data_product(data_product_name)
```

containers/daap-delete-data-product/src/var/task/delete_data_product.py
```diff
@@ -0,0 +1,61 @@
...
+    if not data_product_metadata.exists:
+        error_message = (
+            f"Could not locate metadata for data product: {data_product_name}."
+        )
+        logger.error(error_message)
+        return format_error_response(HTTPStatus.BAD_REQUEST, event, error_message)
+
+    version_creator = VersionCreator(data_product_name=data_product_name, logger=logger)
+    try:
+        version_creator.remove_all_versions_of_data_product()
```
MatMoore, line 49: I like that this method is really clear about what it's going to do 🙂

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -88,6 +91,28 @@ def __init__(self, data_product_name, logger: DataPlatformLogger):
...
+    def remove_all_versions_of_data_product(self):
+        """Completely removes a data product, including tables, schemas, metadata
+        raw & curated data"""
+
+        # Delete the Glue Database
+        delete_glue_database(
+            data_product_name=self.data_product_config.name, logger=self.logger
+        )
+
+        # Delete raw files
```
MatMoore, line 108: This should also delete the failed uploads from fail/ in the data bucket.

I think we should keep the files in the log bucket though.

containers/daap-delete-data-product/tests/unit/mock_test.py
```diff
@@ -0,0 +1,161 @@
...
+        # Assert we have the correct number of metadata files to begin with
+        prefix = f"curated/{data_product_name}/"
+        response = s3_client.list_objects_v2(
+            Bucket=os.getenv("CURATED_DATA_BUCKET"),
+            Prefix=prefix,
+        )
+        assert response.get("KeyCount") == 30
+
+        ######################################################################
+        with patch("glue_utils.glue_client", glue_client):
```
MatMoore, line 112: Just an idea - another way of testing this would be to mock out the VersionCreator object. Then this test could verify that the `remove_all_versions_of_a_data_product` method is called, and the unit test for that method can verify that this method actually does the right thing in glue.

I think it's fine either way, it just depends how much of the integration you want to cover in this specific test

containers/daap-delete-data-product/tests/unit/mock_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+            with pytest.raises(glue_client.exceptions.EntityNotFoundException) as exc:
+                glue_client.get_database(Name=data_product_name)
+            assert (
+                exc.value.response["Error"]["Message"]
+                == f"Database {data_product_name} not found."
+            )
+            assert exc.value.response["Error"]["Code"] == "EntityNotFoundException"
+
+    def test_metadata_does_not_exist(self, fake_event, fake_context):
```
MatMoore: I would name this something like `test_400_error_when_metadata_does_not_exist` or something along those lines - just to be clear that it's testing the case when metadata doesn't exist *before* the call, rather than testing that metadata is removed *by* the call

MatMoore (changes requested):

containers/daap-delete-data-product/tests/unit/mock_test.py
mitchdawson1982: Thanks, ill incorporate this going forward in my testing.

containers/daap-delete-data-product/tests/unit/mock_test.py
mitchdawson1982: completed

containers/daap-python-base/src/var/task/versioning.py
mitchdawson1982: Completed

MatMoore (dismissed):

jacobwoffenden (approved):

MatMoore (approved):

PR #2347, opened 2023-11-14
[DP-2064] Tie data to a major version, instead of a minor version

MatMoore:
https://github.com/ministryofjustice/data-catalogue/issues/85

## Description
All data paths now contain the major version `/v1/` or `/v2/` rather than the full version `/v1.0/` or `/v2.1/`

## Consequences
- It is no longer possible to derive a full data product version from a file path. The version is the last minor version corresponding to the major version in the path.
- Data uploaded in an earlier minor version is still accessible via the athena table after a minor update. E.g. my_data_product_v1.some_table queries all data uploaded to v1.0, v1.1, v1.2 etc
- When columns have been added in a minor version, data from historical parquet files will be treated as NULL in the athena table ([source: stack overflow](https://stackoverflow.com/questions/67280896/add-columns-to-aws-athena-paquet-tables/67333744#67333744))

## Todo
- [x] Review landing-to-raw, presigned-url, athena-load - see if this breaks any assumptions
- [x] TBC: What happens when we add columns? In this case we will have historic parquet files without the column stored in the same location. Will this break athena, or set the columns to null?

### DIFF
--- a/containers/daap-python-base/CHANGELOG.md
+++ b/containers/daap-python-base/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [7.0.0] - 2023-11-14
 
-## Changed
+### Changed
 
 - Data paths now contain only the major version. A minor version change implies
   the schemas are all backwards compatable with existing data.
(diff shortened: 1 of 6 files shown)


containers/daap-python-base/CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [7.0.0] - 2023-11-14
+
+## Changed
```
tom-webber:
```suggestion
### Changed
```

mitchdawson1982 (approved):
LGTM

PR #2363, opened 2023-11-14
rename `VersionCreator` and add `create_schema` versioner

### LINKED ISSUE #1925: Use versioning module to handle versioning everywhere
### User Story

As a data producer
I want to be able to make changes to data and metadata in my data product and have the data platform handle versioning of that data product.
So that consumers of my data product can be effectively notified of any changes and the type of change and are best informed and should they need to change any down stream queries or code they will be able to do so easily.

And

As a developer
i want this logic to be in a central place as it's logic that will be called from multiple endpoints/lambdas.
Sp that code is concise and we are confident the same logic is always applied.

### Value / Purpose

- Traceable changes for any data product, giving the ability to easily rollback should that be desired.
- A way to notify consumers of data, what has changed and when between versions.
- A way to distinguish between versions of a data product where changes have been made that would be deemed breaking, i.e. code and queries ran against an updated data product would no longer work.

### Hypothesis

If we have a common module to handle versioning
Then we can be confident consistent versioning logic is being applied across different data product endpoints

### Proposal

We already have a `VersionCreator` class with `update_schema` and `update_metadata` methods.

This is called from update metadata and update schema.

We should also make sure it is called from **delete table** (creates a new major version) and **create schema** (may create a new minor version). The delete table use case should already be handled by the time this ticket is ready.

The create table use case is currently implemented in the lambda itself. We should refactor this to a `create_schema` method or similar to match `update_schema`. We decided the very first schema when no other schemas are registered in v1.0 of data product remains 1.0, then every subsequent schema created is a minor update

The versioning module should also handle the creation and upversioning of the actual data and databases, which it doesn't yet do (for major version updates).

### Additional Information

We can consider notifications out of scope for this ticket, but eventually `VersionCreator` could call some kind of `Notifier` class.

### Definition of Done

- [ ] Another team member has reviewed
- [ ] Tests are green

tom-webber:
- rename `VersionCreator` to `VersionManager`
- add `create_schema` versioning method
    - add `_minor_version_bump` to increment, copy files, update metadata (extracted from `update_metadata`)
    - add `_verify_input_metadata` to determine valid metadata update
- add `latest_version` as class attr from DataProductConfig
- update `schema_update_type` logic

- add tests for `create_schema` with and without existing schema at v1.0

- call v1.1.0 metadata spec function in conftest
- add logging for json_spec_path used


tom-webber:
closes #1925

containers/daap-create-schema/src/var/task/create_schema.py
```diff
@@ -148,73 +148,28 @@ def handler(event, context):
         )
 
     # Code below that handles verisoning of a data product will be moved to a central module eventually.
```
MatMoore: can remove this comment now I think

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -121,10 +122,14 @@ def remove_all_versions_of_data_product(self):
     def update_metadata_remove_schemas(self, schema_list: list[str]) -> str:
         """Handles removing schema(s) for a data product."""
         s3_client = boto3.client("s3")
+        # remove any list duplicates and preserve list order
+        schema_list = list({k: None for k in schema_list}.keys())
```
MatMoore, line 126: I assume this means the schema list is going to stay in creation order?

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -202,35 +203,85 @@ def update_metadata_remove_schemas(self, schema_list: list[str]) -> str:
...
         if not metadata.valid or not metadata.exists:
-            raise InvalidUpdate()
+            msg = "invalid metadata passed"
+            raise InvalidUpdate(msg)
+
+        changed_fields = metadata.changed_fields()
+        if changed_fields.difference(UPDATABLE_METADATA_FIELDS):
+            changed_fields = list(changed_fields)
+            num_fields = len(changed_fields)
+            msg = f"Non-updatable metadata field{('s'[:num_fields!=1])} changed:"
```
MatMoore, line 260: this looks useful for debugging 👍🏻

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -202,35 +203,85 @@ def update_metadata_remove_schemas(self, schema_list: list[str]) -> str:
...
-            latest_version=latest_version, new_version=new_version
-        )
-        new_version_key = self.data_product_config.metadata_path(new_version).key
-        metadata.write_json_to_s3(new_version_key)
+        new_version = self._minor_version_bump(metadata)
 
         return new_version
 
-    def update_schema(self, input_data: dict, table_name: str):
+    def update_schema(self, input_data: dict, table_name: str) -> tuple[str, dict]:
```
MatMoore, line 325: I noticed `update_schema` takes the data & table name and creates a `DataProductSchema` internally, but `create_schema` takes the `DataProductSchema` as an argument, so there is an inconsistency here.

update_metadata also takes the raw data and runs it through `_validate_input_metadata`. 

Wonder if it's worth doing something similar with schemas - have all the methods take the input data directly and then internally handle the validation and whether a previous version exists?

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -265,6 +315,45 @@ def update_schema(self, input_data: dict, table_name: str):
...
+        after a minor version bump.
+        """
+        data_product_name = self.data_product_config.name
+        metadata_dict = schema.parent_data_product_metadata
+
+        metadata = self._verify_input_metadata(metadata_dict)
+        schema.convert_schema_to_glue_table_input_csv()
+
+        if self.latest_version == "v1.0":
+            if not schema.parent_product_has_registered_schema:
```
MatMoore: This logic is going to mean that if you create a data product with multiple schemas, and then you upload multiple schemas to it, you get multiple minor versions created before you've even uploaded any data:
1.0 -> 1.1 -> 1.2 etc.

I think this is fine for alpha, but makes me wonder if we should instead freeze the version while the data product still has draft status?

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -265,6 +315,45 @@ def update_schema(self, input_data: dict, table_name: str):
...
+                self.logger.info(
+                    f"No existing schemas are associated with {data_product_name}; "
+                    f"setting version to 'v1.0' for {schema.table_name}"
+                )
+
+                schema.write_json_to_s3(
```
MatMoore: minor suggestion: there is some repetition between this if block and the outer scope, that you could avoid if you set `new_version` inside the if and added an else clause for the assignment below.

```python
if self.latest_version == "v1.0":
  # ...
  new_version = "1.0"
else:
  new_version = self._minor_version_bump(metadata)
```

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -288,10 +377,14 @@ def metadata_update_type(data_product_metadata: DataProductMetadata) -> UpdateTy
 
     changed_fields = data_product_metadata.changed_fields()
     if not changed_fields:
+        # NOTE: this does not capture all types of updates (e.g. schema-internal
```
MatMoore: What does schema internal changes mean? Like changes in data type?

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -288,10 +377,14 @@ def metadata_update_type(data_product_metadata: DataProductMetadata) -> UpdateTy
...
+        # changes)
         return UpdateType.Unchanged
     if changed_fields.difference(UPDATABLE_METADATA_FIELDS):
         return UpdateType.NotAllowed
     else:
+        # NOTE: this does not capture all types of updates (e.g. removal of schema ->
```
MatMoore, line 427: 👍🏻 will be looking at this in the follow up ticket

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -335,8 +428,10 @@ def schema_update_type(
         }
         if not changed_fields:
             update_type = UpdateType.Unchanged
-        elif changed_fields.intersection(MINOR_UPDATE_SCHEMA_FIELDS):
+        elif changed_fields.intersection(MINOR_UPDATE_SCHEMA_FIELDS) == changed_fields:
```
MatMoore, line 472: good catch - we definitely want to handle the case where we have a minor update + major update at the same time

MatMoore (dismissed):
Have made a few suggestions but overall looks like a good improvement 🎉

containers/daap-python-base/src/var/task/versioning.py
tom-webber: Yeah this felt a little uncomfortable: it was based on retaining the logic of there being a pre-existing `DataProductSchema` object in the lambda handler already.

For the `create_schema.handler` an existing `schema` object is used for:
- determining whether to commit to the `version_manager.create_schema` call
    - `schema.exists`, `schema.has_registered_data_product`, `schema.valid`
- pushing the schema to the catalogue

`version_manager.create_schema` could instead return a `DataProductSchema` object:
- this object gets passed to push to catalogue
- we could handle the 3 schema checks inside the `version_manager.create_schema` method

EDIT: just realised this is basically what you are proposing.

containers/daap-python-base/src/var/task/versioning.py
tom-webber: Yes, in later iterations we should offer an endpoint that can receive multiple schemas and process them in a single version bump.

containers/daap-python-base/src/var/task/versioning.py
tom-webber: Yes. I've clarified the comment.

containers/daap-python-base/src/var/task/versioning.py
MatMoore: Yep, either that or the other way round - we could change all the other methods to accept objects instead of dictionaries. But I think I prefer the above.

tom-webber: Done

containers/daap-create-schema/src/var/task/create_schema.py
```diff
@@ -118,105 +117,32 @@ def handler(event, context):
...
-        logger.error("schema has no associated registered data product metadata.")
-        return format_error_response(
-            HTTPStatus.FORBIDDEN, event=event, message=error_msg
+    try:
+        new_version, schema = version_manager.create_schema(
+            table_name=table_name, input_data=input_schema
         )
+    except InvalidUpdate as e:
+        error_message = str(e)
+        if error_message.startswith(f"schema for {table_name} has failed validation:"):
```
MatMoore, line 128: (Non blocking) If we need to distinguish these two cases it would be better to split `InvalidUpdate` into two exceptions, then we can just have something like

```python
except InvalidSchema:
  ...
except InvalidUpdate:
  ...
```

Don't think it needs to block the merge though

MatMoore (approved):

PR #2375, opened 2023-11-15
Dropping Cal's permissions

julialawrence:
Calum is now part of another team and doesn't need anything above data engineering levels of access.


Emterry (approved):
LGTM

PR #2378, opened 2023-11-15
:wrench: Update minimum standard number of nodes (dev)

Gary-H9:
As per this mornings troubleshooting we believe this is required configuration.


jhpyke (approved):
lgtm

PR #2388, opened 2023-11-16
:recycle: Move delete product code to handler

mitchdawson1982:
This PR does the following:

- Removes the VersionCreator.remove_all_versions() method
- Places all logic into the delete_data_product.handler function
- Refactors the code so that all steps are attempted regardless of whether metadata, database, or files are present ensuring that a data product is removed in its entirety ( minus Openmetadata at this stage)


containers/daap-delete-data-product/src/var/task/delete_data_product.py
```diff
@@ -24,26 +31,24 @@ def handler(event, context):
...
+        logger.info(f"Glue database '{data_product_name}' not found.")
+
+    # Delete fail files
+    s3_recursive_delete(get_fail_data_bucket(), [f"fail/{data_product_name}/"])
+    # Delete raw files
+    s3_recursive_delete(get_raw_data_bucket(), [f"raw/{data_product_name}/"])
+    # Delete curated files
+    s3_recursive_delete(get_curated_data_bucket(), [f"curated/{data_product_name}/"])
+    # Delete Metadata & Schema files
+    s3_recursive_delete(get_metadata_bucket(), [f"{data_product_name}/"])
```
tom-webber, line 49: Can you handle any exceptions raised if there are errors during delete?
Perhaps a loop over a list of tuples of the method and the path 
```
error_msg = ""
for get_bucket_method, product_path in
[
    (get_fail_data_bucket(), [f"fail/{data_product_name}/"]),
    (get_raw_data_bucket(), [f"raw/{data_product_name}/"]),
    (get_curated_data_bucket(), [f"curated/{data_product_name}/"]),
    (get_metadata_bucket(), [f"{data_product_name}/"]),
]:
    try:
        s3_recursive_delete(get_bucket_method, product_path)
    except Exception as e:
        error_msg.append(str(e))
```

containers/daap-delete-data-product/tests/unit/mock_test.py
tom-webber, line 1: Can you add test cases for if there are no files/keys matching a path? (is the lambda idempotent? Can it be rerun if there is a failure and 3/4 bucket paths are cleared, but one fails?)

containers/daap-delete-data-product/src/var/task/delete_data_product.py
mitchdawson1982: No need as the function wont throw an exception unless the bucket it not found.https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/bucket/objects.html#delete

containers/daap-delete-data-product/tests/unit/mock_test.py
mitchdawson1982: New test added

containers/daap-delete-data-product/src/var/task/delete_data_product.py
MatMoore: Do we need to check the response from delete though to make sure nothing failed?

> For each key, Amazon S3 performs a delete action and returns the result of that delete, success, or failure, in the response.

containers/daap-delete-data-product/tests/unit/conftest.py
```diff
@@ -205,3 +205,51 @@ def create_failed_raw_and_curated_data(
...
+    data_product_name,
+    data_product_versions,
+):
+    for version in data_product_versions:
+        for i in range(10):
+            s3_client.put_object(
+                Bucket=os.getenv("CURATED_DATA_BUCKET"),
+                Key=f"curated/{data_product_name}/{version}/schema0/curated-file-{str(i)}.json",
+                Body=json.dumps({"content": f"{i}"}),
+            )
```
tom-webber: DRY suggestion:
Pull out this loop to a `populate_bucket` function that takes `bucket_name` and `bucket_type` as additional args then call that function for these three `create_x_data` functions

(This could also be shrunk into a single fixture `populate_buckets` that loops over the three sets of bucket args)

containers/daap-delete-data-product/tests/unit/mock_test.py
```diff
@@ -37,15 +37,25 @@
 }
 
 
+def count_files(client, bucket, prefix):
+    response = client.list_objects_v2(
+        Bucket=bucket,
+        Prefix=prefix,
+    )
+    return response.get("KeyCount")
```
tom-webber:
```suggestion
    # TODO: pagination if >1000 objects expected in bucket
    return response.get("KeyCount")
```

NOTE: as I understand, this has a [limit up to 1000](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/list_objects_v2.html) (and so [requires pagination](https://stackoverflow.com/a/70716816/10478115) if more than 1000 entries)

containers/daap-delete-data-product/tests/unit/mock_test.py
```diff
@@ -65,91 +75,144 @@ def test_success(
...
-            Bucket=os.getenv("METADATA_BUCKET"),
-            Prefix=prefix,
+        assert (
+            json.loads(result.get("body"))["message"]
+            == f"Successfully removed data product '{data_product_name}'."
         )
-        assert response.get("KeyCount") == 12
 
-        # Assert we have the correct number of fail files to begin with
+    def test_fail_files_are_deleted(
```
tom-webber:
```suggestion
     def test_files_in_fail_are_deleted(
```
suggest renaming as `test_fail_` can be misleading in this context

containers/daap-delete-data-product/tests/unit/conftest.py
mitchdawson1982: `@pytest.fixture
def create_fail_data(
    s3_client,
    create_raw_bucket,
    data_product_name,
    data_product_versions,
):
    populate_buckets(
        data_product_versions,
        s3_client,
        data_product_name,
        "RAW_DATA_BUCKET",
        "fail",
    )`

containers/daap-delete-data-product/tests/unit/mock_test.py
mitchdawson1982: Makes sense

containers/daap-delete-data-product/tests/unit/mock_test.py
mitchdawson1982: Agreed, now paginated.

containers/daap-delete-data-product/tests/unit/conftest.py
```diff
@@ -179,29 +179,62 @@ def create_glue_tables(create_glue_database, glue_client, data_product_name):
...
 
-@pytest.fixture
-def create_failed_raw_and_curated_data(
-    s3_client,
-    create_raw_bucket,
-    create_curated_bucket,
-    data_product_name,
-    table_name,
-    data_product_versions,
+def populate_buckets(
```
tom-webber:
```suggestion
def populate_bucket(
```

containers/daap-delete-data-product/tests/unit/conftest.py
```diff
@@ -179,29 +179,62 @@ def create_glue_tables(create_glue_database, glue_client, data_product_name):
...
+
+
+@pytest.fixture
+def create_curated_data(
+    s3_client,
+    create_curated_bucket,
+    data_product_name,
+    data_product_versions,
+):
+    populate_buckets(
```
tom-webber:
```suggestion
    populate_bucket(
```

containers/daap-delete-data-product/tests/unit/conftest.py
```diff
@@ -179,29 +179,62 @@ def create_glue_tables(create_glue_database, glue_client, data_product_name):
...
+
+
+@pytest.fixture
+def create_raw_data(
+    s3_client,
+    create_raw_bucket,
+    data_product_name,
+    data_product_versions,
+):
+    populate_buckets(
```
tom-webber:
```suggestion
    populate_bucket(
```

containers/daap-delete-data-product/tests/unit/conftest.py
```diff
@@ -179,29 +179,62 @@ def create_glue_tables(create_glue_database, glue_client, data_product_name):
...
+
+
+@pytest.fixture
+def create_fail_data(
+    s3_client,
+    create_raw_bucket,
+    data_product_name,
+    data_product_versions,
+):
+    populate_buckets(
```
tom-webber:
```suggestion
    populate_bucket(
```

tom-webber (approved):
Looks good to go

Issue #2379, opened 2023-11-15
🐛 Kyverno fails to properly exclude system namespaces in existing Airflow deployment


jhpyke:
### Describe the bug.

Kyverno is used as the cluster admission controller for all pods on the airflow clusters. It is intended to be given a `resource_map` of values that indicate which namespaces are immune to enforcement, however [the current deployment](https://github.com/ministryofjustice/data-engineering-airflow/blob/main/infra/eks/kyverno.py) is not applying this correctly. This is leading to behaviour where kyverno can wrongly prevent admission onto the cluster of core components.

### To Reproduce

All steps should be done in `analytical-platform-data-production` using `airflow-sandpit` eks cluster.
1. Drop the number of nodes in the 'standard' node group to 0
2. Once the node is successfully deleted, attempt to raise it back to a non-zero number
3. Node will show as `NotReady` status, and no pods will be able to schedule
4. Remove the [webhooks per this guide](https://kyverno.io/docs/troubleshooting/) and the nodes will recover.

These steps assume that kyverno is installed on the Cluster when reproducing.

### Expected Behaviour

We should be able to restart from 0 nodes without manual intervention.

### Additional context

We need to ensure the component is ported into our terraform code and can be deployed to `dev` without issues. This should be done before trying to tackle the bug with the exempting namespaces.

jhpyke:
Resolved by the below PRs:

[Initial Dev Implementation](https://github.com/ministryofjustice/data-platform/pull/2444)
[Fixes to initial implimentation](https://github.com/ministryofjustice/data-platform/pull/2512)
[Prod Implimentation (with fixes applied)](https://github.com/ministryofjustice/data-platform/pull/2651)

PR #2444, opened 2023-11-20
Adding kyverno configuration in Airflow-Dev

jhpyke:
This PR is half of the resolution to #2379 - This will bring kyverno into management in terraform for Airflow Dev, and deploy the required fixes to ensure we don't have further issues. `Prod` is out of scope of this PR, as this will require a slightly different process due to the existing Kyverno installation on said cluster. As such, the `Prod` half of this ticket will be handled in a separate PR, and this ticket DOES NOT close the overall issue.

Overall, the main fix is the application of the new `webhooks:` section of the config in [helm/kyverno/values.yml.tftpl](https://github.com/ministryofjustice/data-platform/pull/2444/files#diff-75058917d652abff143bd52ab51e980bed966aa1c77484e1abe779f11a1d2281). All other changes are merely resource definitions to allow for this change to be applied effectively.


julialawrence (approved):
LGTM

PR #2465, opened 2023-11-21
Delete table database versioning

mitchdawson1982:
- Adds support for deleting versioned databases
- No longer returns if schema.json is not found
- No longer returns if glue table not found


containers/daap-python-base/tests/unit/versioning_test.py
```diff
@@ -434,63 +434,36 @@ def test_success(
...
-        schema_list = ["schema3", "schema4"]
-
-        with pytest.raises(InvalidUpdate) as exc:
-            version_manager.update_metadata_remove_schemas(schema_list=schema_list)
-        assert (
-            str(exc.value)
-            == "Invalid schemas found in schema_list: ['schema3', 'schema4']"
-        )
-
-    def test_glue_table_not_found(
```
tom-webber, line 455: Is there a replacement for this test? (trying to remove invalid schemas)

mitchdawson1982: The code used to be strict on validating if a schema existed in metadata schemas and would return, however it was pointed out that the could be scenarios where data files or tables were present, but a schema was not so this was not an optimal approach. The code will log schemas that are not present in the metadata but will attempt to remove tables and files for any provided schema value. Hence there is no longer a need to validate the supplied schema.

tom-webber: I think I get it now - there was no way to delete an 'orphaned' table for a Data Product through the API.

I think it may be preferable to at least return the failure message to the user `'{schema_name} schema metadata not found, {schema_name} schema metadata not deleted'`, ideally together with a message of the glue table being deleted to give the full context of what's happened (as part of the response message):

```python
# schema not found
f"{schema} schema metadata not found, {schema} schema metadata not deleted"
f"{data_product}.{schema} table was successfully deleted, and can no longer be queried"

# schema found
f"{schema} schema metadata found, {schema} schema metadata successfully deleted"
f"{data_product}.{schema} table was successfully deleted, and can no longer be queried"
```

MatMoore:
I've merged the changes to glue_and_athena utils and versioning modules in this PR: https://github.com/ministryofjustice/data-platform/pull/2496

We now create new versions of the athena db when something is removed, but we still need to handle the data copying: https://github.com/ministryofjustice/data-catalogue/issues/65

PR #2468, opened 2023-11-21
:recycle: Add database versioning support

mitchdawson1982:
- Adds support for deleting versioned databases

### DIFF
--- a/containers/daap-delete-data-product/CHANGELOG.md
+++ b/containers/daap-delete-data-product/CHANGELOG.md
@@ -11,18 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [2.1.0] 2023-11-21
 
-### Added
-
-- `database_names` fixture to generate a list of database names
-    for all major versions
-
 ### Changed
 
-- updated `create_glue_database` to `create_glue_databases`
-- `create_glue_databases` now creates databases for all major versions
-- `create_glue_tables` now creates tables for all database names
-- `test_databases_are_deleted` now checks for databases
-    for all major versions
+- `delete_data_product.handler` now deletes databases
+    for all majore versions of the data product
 
 ## [2.0.0] 2023-11-16
 
--- a/containers/daap-delete-data-product/CHANGELOG.md
+++ b/containers/daap-delete-data-product/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `delete_data_product.handler` now deletes databases
-    for all majore versions of the data product
+    for all major versions of the data product
 
 ## [2.0.0] 2023-11-16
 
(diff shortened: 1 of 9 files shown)


containers/daap-delete-data-product/CHANGELOG.md
```diff
@@ -9,6 +9,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Added
+
+- `database_names` fixture to generate a list of database names
+    for all major versions
+
+### Changed
+
+- updated `create_glue_database` to `create_glue_databases`
+- `create_glue_databases` now creates databases for all major versions
```
MatMoore: I think this is a little confusing as it's referring to test code. I think this changelog should focus on the observable behaviour of the lambda.

containers/daap-delete-data-product/src/var/task/delete_data_product.py
```diff
@@ -34,10 +35,14 @@ def handler(event, context):
     logger.add_data_product(data_product_name)
     logger.info(f"event: {event}")
 
-    try:
-        glue_client.delete_database(Name=data_product_name)
-    except glue_client.exceptions.EntityNotFoundException:
-        logger.info(f"Glue database '{data_product_name}' not found.")
+    major_versions = get_all_major_versions(data_product_name=data_product_name)
```
MatMoore, line 39: (Non-blocking) might be nice to move this to glue_and_athena_utils so we've got a standard way to create/update/delete glue databases and tables.
https://github.com/ministryofjustice/data-platform/blob/main/containers/daap-python-base/src/var/task/glue_and_athena_utils.py

containers/daap-delete-data-product/tests/unit/conftest.py
```diff
@@ -167,16 +167,28 @@ def create_metadata(
...
+    )
 
 
 @pytest.fixture
-def create_glue_tables(create_glue_database, glue_client, data_product_name):
-    for i in range(3):
-        glue_client.create_table(
-            DatabaseName=data_product_name, TableInput={"Name": f"schema{i}"}
-        )
+def create_glue_databases(glue_client, data_product_name, database_names):
```
MatMoore: I think this could be replaced with calls to code in the base image https://github.com/ministryofjustice/data-platform/blob/main/containers/daap-python-base/src/var/task/glue_and_athena_utils.py#L12

Either way, it looks like the data_product_name arg is not needed here.

MatMoore (dismissed):

containers/daap-delete-data-product/CHANGELOG.md
mitchdawson1982: Removed

containers/daap-delete-data-product/src/var/task/delete_data_product.py
mitchdawson1982: Moved to glue & athena utils

containers/daap-delete-data-product/tests/unit/conftest.py
mitchdawson1982: Argument removed and now using the create database code from athena and glue utils

MatMoore (approved):

PR #2469, opened 2023-11-22
:bug: Fixing trivy workflow

julialawrence:


jacobwoffenden:
@julialawrence with Trivy being performed on pull request AND https://github.com/ministryofjustice/data-platform/issues/886 being worked on, do we still need this workflow?

Emterry:
Theres an unresolved bug in Trivy where sarif outputs for terraform module uris can have the form `git::https:/github.com`. As a workaround I have added a step to remove the scheme. I think this was not an issue during the initial workflow pr because I missed the checkout step.

jacobwoffenden:
> Theres an unresolved bug in Trivy where sarif outputs for terraform module uris can have the form `git::https:/github.com`. As a workaround I have added a step to remove the scheme. I think this was not an issue during the initial workflow pr because I missed the checkout step.

@Emterry could you drop a link to that bug please?

I still think we should remove this workflow as stated in my previous comment

jacobwoffenden (approved):

PR #2475, opened 2023-11-22
Dpl 2107 major version increment

LavMatt:
This implements the creation of a new database version via update schema, when a schema update has resulted in a major version increment.

It copies the latest snapshot of data from all tables (other than updated table) from the previous version to the new version database

For the table that has been updated it creates an empty table as per the updated schema and then applies logic as to whether data for the updated table can be copied over tot he new version data product or if it should remain empty until new data has been uploaded by a user.

said logic:
```
Columns removed -> OK (will be ignored by athena)
Columns added -> OK (will be left as null)
Columns reordered -> OK
Data types changed -> Not OK
```

It seems to be working but still in need of some refinement.

### Additions
- `CuratedDataCopier` class to `curated_data_loader.py`, handles creation of the new database and the copying of each table in the data product to it. It defines whether the updated table is valid for copying via the changes passed in from versioning and calls a new method of `CuratedDataLoader` for each table.
- `CuratedDataLoader.create_new_major_version_data_product_table()` method added, which is what the above calls. This runs new queries added to the `CuratedDataQueryBuilder` class.
- `sql_create_next_major_increment_table()` and `sql_unload_for_major_updated_table()` added to the `CuratedDataQueryBuilder` class for generating the sql to run to create tables in new version database for unchanged and changed tables respectively.

### DIFF
--- a/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
+++ b/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
@@ -138,9 +138,9 @@ class CuratedDataCopier:
         self.glue_client = glue_client
         self.athena_client = athena_client
         self.logger = logger
-        self.get_tables_to_copy()
+        self._get_tables_to_copy()
 
-    def validate(self):
+    def _is_updated_table_valid_for_copy(self):
         """
         Check whether it is possible to copy the data of changed schema
         between two major versions
@@ -156,12 +156,14 @@ class CuratedDataCopier:
         else:
             return True
 
-    def run(self):
+    def run(self) -> list[DataProductSchema]:
         """
         This runs the actual copy of each data product table into the new database version
 
         The table for where schema has been updated is created via boto3
         as it will not always have data loaded to it via this method
+
+        returns a list of updated DataProductSchema objects to caller
         """
         # create new version database and table for updated schema
         create_glue_database(self.glue_client, self.new_database_name, self.logger)
@@ -215,11 +217,11 @@ class CuratedDataCopier:
 
         return schemas_for_rewrite
 
-    def get_tables_to_copy(self):
+    def _get_tables_to_copy(self):
         """
         Creates a list of tables to copy into new version of data product database
         """
-        if self.validate():
+        if self._is_updated_table_valid_for_copy():
             self.tables_to_copy = self.schema.parent_data_product_metadata["schemas"]
             self.copy_updated_table = True
         else:
--- a/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
+++ b/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
@@ -167,7 +167,6 @@ class CuratedDataCopier:
         create_glue_database(self.glue_client, self.new_database_name, self.logger)
         parquet_table_input = format_parquet_glue_table_input(
             self.schema.data["TableInput"],
-            # os.path.join(self.new_curated_data_product_path, self.schema.table_name),
             self.new_curated_data_product_path,
         )
         self.glue_client.create_table(
@@ -210,7 +209,7 @@ class CuratedDataCopier:
                 glue_client=self.glue_client,
                 logger=self.logger,
             ).create_new_major_version_data_product_table(
-                is_updated_table=table == self.schema.table_name
+                is_updated_table=(table == self.schema.table_name)
             )
             new_version_key = self.element.data_product.schema_path(
                 schema_for_copy.table_name
--- a/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
+++ b/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
@@ -106,7 +106,6 @@ class CuratedDataLoader:
                 database_name=self.curated_data_table.database,
                 sql=self.query_builder.sql_unload_for_major_updated_table(
                     curated_table=self.curated_data_table,
-                    column_metadata=self.column_metadata,
                 ),
                 logger=self.logger,
             )
--- a/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
+++ b/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
@@ -96,7 +96,6 @@ class CuratedDataLoader:
                 database_name=self.curated_data_table.database,
                 sql=self.query_builder.sql_create_next_major_increment_table(
                     self.curated_data_table,
-                    self.column_metadata,
                 ),
                 logger=self.logger,
             )
--- a/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
+++ b/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
@@ -173,7 +173,7 @@ class CuratedDataCopier:
             DatabaseName=self.new_database_name,
             TableInput=parquet_table_input,
         )
-
+        schemas_for_rewrite = []
         for table in self.tables_to_copy:
             input_data = format_table_schema(
                 (
@@ -211,11 +211,9 @@ class CuratedDataCopier:
             ).create_new_major_version_data_product_table(
                 is_updated_table=(table == self.schema.table_name)
             )
-            new_version_key = self.element.data_product.schema_path(
-                schema_for_copy.table_name
-            ).key
-            # save schema with updated database name for newly created version
-            schema_for_copy.write_json_to_s3(new_version_key)
+            schemas_for_rewrite.append(schema_for_copy)
+
+        return schemas_for_rewrite
 
     def get_tables_to_copy(self):
         """
--- a/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
+++ b/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
@@ -1,5 +1,3 @@
-import os
-
 from data_platform_logging import DataPlatformLogger
 from data_platform_paths import DataProductElement, QueryTable
 from data_product_metadata import DataProductSchema, format_table_schema
--- a/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
+++ b/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
@@ -1,5 +1,8 @@
+import os
+
 from data_platform_logging import DataPlatformLogger
-from data_platform_paths import QueryTable
+from data_platform_paths import DataProductElement, QueryTable
+from data_product_metadata import DataProductSchema, format_table_schema
 from glue_and_athena_utils import (
     create_glue_database,
     refresh_table_partitions,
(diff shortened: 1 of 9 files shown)


containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
```diff
@@ -80,3 +81,182 @@ def ingest_raw_data(self, raw_data_table: QueryTable, extraction_timestamp: str)
...
+        Check whether it is possible to copy the data of changed schema
+        between two major versions
+
+        Columns removed -> OK (will be ignored by athena)
+        Columns added -> OK (will be left as null)
+        Columns reordered -> OK
+        Data types changed -> Not OK
+        """
+
+        if self.column_changes["types_changed"]:
```
MatMoore, line 154: Maybe this should throw an exception rather than return a boolean, so we can expose a bit of information about what is invalid. Either should work for now but not sure if the logic will get more complicated than this later.

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
```diff
@@ -80,3 +81,182 @@ def ingest_raw_data(self, raw_data_table: QueryTable, extraction_timestamp: str)
...
+        This runs the actual copy of each data product table into the new database version
+
+        The table for where schema has been updated is created via boto3
+        as it will not always have data loaded to it via this method
+        """
+        # create new version database and table for updated schema
+        create_glue_database(self.glue_client, self.new_database_name, self.logger)
+        parquet_table_input = format_parquet_glue_table_input(
+            self.schema.data["TableInput"],
+            # os.path.join(self.new_curated_data_product_path, self.schema.table_name),
```
MatMoore: think this line can be deleted now

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
```diff
@@ -80,3 +81,182 @@ def ingest_raw_data(self, raw_data_table: QueryTable, extraction_timestamp: str)
...
+                    name=schema_for_copy.table_name,
+                ),
+                table_path=self.new_curated_data_product_path.replace(
+                    self.schema.table_name, table
+                ),
+                athena_client=self.athena_client,
+                glue_client=self.glue_client,
+                logger=self.logger,
+            ).create_new_major_version_data_product_table(
+                is_updated_table=table == self.schema.table_name
```
MatMoore: minor suggestion for readability

```suggestion
                is_updated_table=(table == self.schema.table_name)
```

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
```diff
@@ -80,3 +81,182 @@ def ingest_raw_data(self, raw_data_table: QueryTable, extraction_timestamp: str)
...
+                        data_product_name=self.data_product_name,
+                        table_name=table,
+                        logger=self.logger,
+                        input_data=None,
+                    )
+                    .load()
+                    .latest_version_saved_data
+                )
+            )
+            schema_for_copy = DataProductSchema(
```
MatMoore, line 192: what's the difference between this object and the result of `load()` above?

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -314,6 +315,15 @@ def update_schema(self, input_data: dict, table_name: str) -> tuple[str, dict]:
             )
             schema.convert_schema_to_glue_table_input_csv()
             schema.write_json_to_s3(new_version_key)
+            # if major we need to create next major version data product data
```
MatMoore, line 324: I've been wondering about whether this version update operation needs to be treated as an atomic operation. I.e. if something fails part way through, should we attempt to roll back the metadata?

I think this is probably something that is OK for now, but we could make it more resilient later on (i.e. once we have users!)

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
```diff
@@ -80,3 +81,182 @@ def ingest_raw_data(self, raw_data_table: QueryTable, extraction_timestamp: str)
...
+        self.schema = new_schema
+        self.element = element
+        self.new_curated_data_product_path = element.curated_data_prefix.uri
+        self.new_database_name = element.database_name
+        self.glue_client = glue_client
+        self.athena_client = athena_client
+        self.logger = logger
+        self.get_tables_to_copy()
+
+    def validate(self):
```
MatMoore: should this be called by the version class? (before `run()`)

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
```diff
@@ -80,3 +81,182 @@ def ingest_raw_data(self, raw_data_table: QueryTable, extraction_timestamp: str)
             database_name=self.curated_data_table.database,
             table_name=self.curated_data_table.name,
         )
+
+    def create_new_major_version_data_product_table(self, is_updated_table):
```
MatMoore, line 85: I think this could be split into two functions, since there's nothing in common between the two branches

i.e. we could have something like

```python
def copy_table_from_previous_major_version(self)
#...

def create_table_and_unload_from_previous_major_version(self)
#...
```

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
```diff
@@ -80,3 +81,182 @@ def ingest_raw_data(self, raw_data_table: QueryTable, extraction_timestamp: str)
...
+                athena_client=self.athena_client,
+                glue_client=self.glue_client,
+                logger=self.logger,
+            ).create_new_major_version_data_product_table(
+                is_updated_table=table == self.schema.table_name
+            )
+            new_version_key = self.element.data_product.schema_path(
+                schema_for_copy.table_name
+            ).key
+            # save schema with updated database name for newly created version
```
MatMoore: As discussed earlier, if we are writing schemas, I think that firmly belongs in the versioning class.

Also, I suspect this line wouldn't be needed if we didn't store the schemas files in glue format within the metadata bucket. Were we planning to change that to do the glue conversion on demand?

MatMoore (changes requested):
Overall this looks really promising. The main things I think need doing before merge are
1. make sure the validate method gets called
2. move that schema update into the versioning class
3. unit tests

Last week I spoke with @tom-webber about the possibility of adding a flag to create_schema, so that the user gets to decide whether data should be copied or not. That way if they try and copy data, but they have done a change we can't handle with this copy code (e.g. changing data types) they will find out straight away from the validation error, rather than it silently skipping that step. What do you think? (Since we're at the end of the sprint now it might be better to handle this in a separate ticket if we think it's worth adding)

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
LavMatt: Yeah i agree, it would be useful to return in a response that either the change has resulted in the next version database's table (that was subject to an update) being created and populated with the latest snapshot of data or simply created as an empty table in anticipation of new data for that table being uploaded

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
LavMatt: So may have over complicated this step but it gets the latest saved version of a schema, for each table in the list to copy, to feed into the `DataProductSchema` input data to get that as the full `DataProductSchema` object

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
LavMatt: yeah potentially, could be called from there and then the result used to inform the requester of the copy behaviour

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
LavMatt: Yeah agree this would be better placed in versioning class. And i think you're right if we saved metadata in the format we collect it i think we could reduce some of the complexity here and likely elsewhere too. Hopefully we can add a converter for data platform metadata to mojap-metadata and have that handle all our conversion needs

LavMatt:
> Last week I spoke with @tom-webber about the possibility of adding a flag to create_schema, so that the user gets to decide whether data should be copied or not. That way if they try and copy data, but they have done a change we can't handle with this copy code (e.g. changing data types) they will find out straight away from the validation error, rather than it silently skipping that step. What do you think? (Since we're at the end of the sprint now it might be better to handle this in a separate ticket if we think it's worth adding)

I suppose this depends on the scenarios where a user might not want to copy their old data across. Maybe if they wanted a clean break as they'd identified errors in data and wanted to create a new version to separate out incorrect from corrected data. 

That also makes me think we'd want an option to be able to create a new version data product where there have been no changes at all. 

I think it's worth adding something like this, i.e. a flag to indicate copy behaviour and maybe even a further endpoint for ict a manual user triggered major version update. 

Think worth adding a couple of tickets to cover this?

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
LavMatt: Thinking about this i'd rather keep as one. Keeps the call more simple as otherwise we'd need to conditionally run either method from the `CuratedDataLoader` class

containers/daap-python-base/tests/unit/curated_data/test_curated_data_loader.py
```diff
@@ -0,0 +1,163 @@
...
+    "dataProductOwnerDisplayName": "matt laverty",
+    "email": "matthew.laverty@justice.gov.uk",
+    "status": "draft",
+    "retentionPeriod": 3000,
+    "dpiaRequired": False,
+    "schemas": ["test_table", "test_table1"],
+}
+
+
+changes1 = {
```
MatMoore: Suggestion: maybe these vars could have more descriptive names, like `remove_one_column` `remove_column_and_change_type` etc?

containers/daap-python-base/tests/unit/curated_data/test_curated_data_loader.py
```diff
@@ -0,0 +1,163 @@
...
+        self.bucket_name = metadata_bucket
+
+        s3_client.put_object(
+            Body=json.dumps(test_metadata),
+            Bucket=self.bucket_name,
+            Key=f"{test_metadata['name']}/v1.0/metadata.json",
+        )
+
+    @pytest.mark.parametrize("changes, valid_copy, expected_copy", test_params1)
+    def test_init(self, athena_client, glue_client, changes, valid_copy, expected_copy):
```
MatMoore: [minor] I think you could skip this test, unless you consider `copy_updated_table` `table_to_copy` attributes to be an important part of the public interface of this class. The object will need to be initialised in every test, so the code will be exercised either way.

MatMoore (dismissed):

MatMoore (approved):

PR #2496, opened 2023-11-23
Delete table database versioning (revised PR with unit tests)

MatMoore:
This takes the changes to versioning.py from https://github.com/ministryofjustice/data-platform/pull/2465, adds unit tests, and reworks the behaviour slightly.

This could do with more integration testing, but I have tested the happy path by running the container locally, and once this is rolled out to API gateway we will do some further testing in order to document the scenarios here: https://dsdmoj.atlassian.net/wiki/spaces/DataPlatform/pages/4557963328/user+journeys+via+API+WIP

## New behaviour for delete table
1. We decided Thursday morning that the delete table endpoint should return early if a schema doesn't exist, even if there may be data hanging around. I.e. we shouldn't attempt to delete data for which there is no metadata. In future we will ensure that any generated schemas are written to s3 as a prerequisite of loading data into a table.

2. For now, we will just delete the table & data from the new version. This deviates from our original plan of deleting from all historical versions (i.e. "delete means delete"), but we had some doubts about implementing contradictory behaviour (i.e. creating a new major version while at the same time making breaking changes to existing versions doesn't make sense). I've created https://github.com/ministryofjustice/data-platform/issues/2500 to revisit the delete means delete behaviour. For now the exact logic doesn't matter as long as the endpoint behaves consistently.

So the expected behaviour when deleting a table is now:

1. If the schema is not found, do nothing
2. Create a new version of the database with the new version number in
3. If the glue table is there, delete it **from the new version only**
4. Data will still be there in historical versions

## Copying tables between databases
Originally this had an exclude list of fields that should not be copied over when we copy tables. I've changed this to an include list so that the code is less likely to break completely if amazon starts returning additional fields in the get table response. (When testing this I ran into several fields that weren't handled by the previous version of this code)
 
## Refactor of glue_and_athena_utils
This now contains functions for creating, deleting, and fetching databases and tables. I've renamed some of the existing functions in this module to follow a common naming convention, but kept aliases with the old names.

### DIFF
--- a/containers/daap-python-base/src/var/task/data_platform_logging.py
+++ b/containers/daap-python-base/src/var/task/data_platform_logging.py
@@ -63,6 +63,22 @@ class DataPlatformLogger:
 
         self.log_list_dict: List[dict] = []
         self.format = format
+
+        is_local = (
+            self.extra["lambda_name"] is None
+            or self.extra["lambda_name"] == "test_function"
+        )
+
+        extra_processors = (
+            [structlog.dev.ConsoleRenderer(event_key="message")]
+            if is_local
+            else [
+                structlog.processors.dict_tracebacks,
+                self.write_log_dict_to_s3_json,
+                structlog.processors.JSONRenderer(),
+            ]
+        )
+
         structlog.configure(
             wrapper_class=structlog.make_filtering_bound_logger(logging_levels[level]),
             processors=[
@@ -71,13 +87,11 @@ class DataPlatformLogger:
                     fmt="%Y-%m-%d %H:%M:%S", key="date_time"
                 ),
                 structlog.processors.add_log_level,
-                structlog.processors.dict_tracebacks,
                 structlog.processors.CallsiteParameterAdder(
                     parameters={structlog.processors.CallsiteParameter.FUNC_NAME},
                     additional_ignores=["data_platform_logging"],
                 ),
-                self.write_log_dict_to_s3_json,
-                structlog.processors.JSONRenderer(),
+                *extra_processors,
             ],
         )
         self.logger = structlog.get_logger(**self.extra)
--- a/containers/daap-python-base/src/var/task/data_platform_paths.py
+++ b/containers/daap-python-base/src/var/task/data_platform_paths.py
@@ -264,7 +264,12 @@ def generate_all_element_version_prefixes(
 def generate_element_version_prefixes_for_version(
     path_prefix: str, data_product_name: str, table_name: str, version: str
 ) -> str:
-    """Generates element prefixes for all data product versions"""
+    """
+    Generates element prefixes for all data product versions
+
+    >>> generate_element_version_prefixes_for_version('curated', 'my-data-product', 'some-table', 'v1.0')
+    'curated/my-data-product/v1/some-table/'
+    """
     major_version = version.split(".")[0]
 
     return f"{path_prefix}/{data_product_name}/{major_version}/{table_name}/"
--- a/containers/daap-python-base/src/var/task/data_platform_paths.py
+++ b/containers/daap-python-base/src/var/task/data_platform_paths.py
@@ -93,6 +93,12 @@ def specification_path(
     )
 
 
+def get_database_name_for_version(
+    data_product_name: str, latest_major_version: str
+) -> str:
+    return data_product_name + "_" + latest_major_version
+
+
 def get_new_version(version, increment_type):
     if increment_type == "minor":
         new_version = version.split(".")[0] + "." + str(int(version.split(".")[-1]) + 1)
@@ -264,7 +270,9 @@ class DataProductElement:
     @property
     def database_name(self) -> str:
         latest_major_version = self.data_product.latest_major_version
-        return self.data_product.name + "_" + latest_major_version
+        return get_database_name_for_version(
+            self.data_product.name, latest_major_version
+        )
 
     @staticmethod
     def load(element_name, data_product_name):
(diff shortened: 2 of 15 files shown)


containers/daap-python-base/tests/unit/versioning_test.py
```diff
@@ -474,36 +483,23 @@ def test_glue_table_not_found(
...
-            table = glue_client.get_table(
-                DatabaseName=data_product_name, Name=f"{schema_list[0]}"
-            )
-            assert table["ResponseMetadata"]["HTTPStatusCode"] == 200
-            assert table["Table"]["Name"] == schema_list[0]
-
-            with pytest.raises(ClientError) as exc:
-                # Call the handler
-                version_manager.update_metadata_remove_schemas(schema_list=schema_list)
+            # Call the handler
```
MatMoore: ❗ this test is flaky, need to investigate

MatMoore:
Decision after speaking to @PriyaBasker23 

For now, we will just delete the table & data from the new version. This deviates from our original plan of deleting from all historical versions "delete means delete", but we had some doubts about implementing contradictory behaviour (i.e. creating a new major version while at the same time making breaking changes to existing versions doesn't make sense).

Created https://github.com/ministryofjustice/data-platform/issues/2500 to revisit the delete means delete behaviour. For now the exact logic doesn't matter as long as the endpoint behaves consistently.

MatMoore:
Need to rebase this monday due to conflicts.

Remaining stuff

- [x] Bump the version
- [x] Make sure glue errors are handled in the delete table lambda
- [x] Test the lambda

containers/daap-python-base/src/var/task/glue_and_athena_utils.py
```diff
@@ -9,67 +9,160 @@
 athena_client = boto3.client("athena")
 
 
-def create_glue_database(
-    glue_client: BaseClient, database_name: str, logger: DataPlatformLogger
-):
+def create_database(
```
MatMoore, line 12: I think this change might have broken some tests on main, as it no longer gets passed the glue client

containers/daap-python-base/src/var/task/data_platform_logging.py
```diff
@@ -63,6 +63,19 @@ def __init__(
 
         self.log_list_dict: List[dict] = []
         self.format = format
+
```
MatMoore, line 66: This change makes the logs more readable when testing locally - see https://www.structlog.org/en/stable/console-output.html

containers/daap-python-base/src/var/task/data_platform_paths.py
```diff
@@ -240,17 +246,30 @@ def generate_all_element_version_prefixes(
...
+                data_product_name=data_product_name,
+                table_name=table_name,
+                version=version,
+            )
         )
 
     return element_prefixes
 
 
+def generate_element_version_prefixes_for_version(
```
murdo-moj, line 264: rename or add an example in the docstring of inputs and outputs `generate_table_path_prefix`

containers/daap-python-base/src/var/task/glue_and_athena_utils.py
```diff
@@ -9,67 +9,168 @@
...
 
-def delete_glue_database(database_name: str, logger: DataPlatformLogger) -> None:
-    """Delete a glue database with the given database name"""
+def get_database(database_name: str, logger: DataPlatformLogger) -> dict | None:
+    """Get the database for the given database name"""
     try:
-        glue_client.delete_database(Name=database_name)
-    except glue_client.exceptions.EntityNotFoundException:
-        logger.info(f"Glue database '{database_name}' not found.")
+        datebase = glue_client.get_database(Name=database_name)
```
murdo-moj: database =/= datebase

murdo-moj:
```suggestion
        database = glue_client.get_database(Name=database_name)
```

containers/daap-python-base/src/var/task/glue_and_athena_utils.py
```diff
@@ -9,67 +9,168 @@
...
+    except ClientError as e:
+        if e.response["Error"]["Code"] == "EntityNotFoundException":
+            logger.error(f"Database name {database_name} not found.")
+            return None
+        else:
+            logger.error(f"Unexpected error: {e}")
+            raise
+    else:
+        return datebase
```
murdo-moj:
```suggestion
        return database
```

containers/daap-python-base/src/var/task/glue_and_athena_utils.py
```diff
@@ -9,67 +9,168 @@
...
 
 
-def refresh_table_partitions(
-    database_name: str, table_name: str, workgroup: str = "data_product_workgroup"
+def clone_database(
+    existing_database_name: str, new_database_name: str, logger: DataPlatformLogger
 ) -> None:
     """
-    Refreshes partitions for a given table
+    Make a copy of a database with a new name, copying all metadata except ids and timestamps.
```
murdo-moj, line 74:
```suggestion
    Make a copy of a database with a new name, copying all metadata except ids and timestamps. This function makes a new empty database with the same metadata definition, note it does NOT copy any of the tables or data within a database.
```

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -399,6 +394,47 @@ def _copy_from_previous_version(self, latest_version, new_version):
             logger=self.logger,
         )
 
+    def _create_database_for_new_version(
+        self, data_product_name, latest_version, new_version
+    ):
+        """
+        Copy the athena database from the old version to the new version
```
murdo-moj:
```suggestion
        Copy the athena database configuration from the old version to the new version
        Only the metadata is copied, none of the contents data
```

murdo-moj (dismissed):

murdo-moj (approved):

PR #2503, opened 2023-11-24
Dpl 2107 upversion daap update schema

LavMatt:
collect and return response from the copy functionality added, where a major version increment is inferred from update.

Also return more information re. increment type in response.

### DIFF
--- a/containers/daap-update-schema/src/var/task/update_schema.py
+++ b/containers/daap-update-schema/src/var/task/update_schema.py
@@ -19,12 +19,14 @@ def handler(event, context):
     logger.add_data_product(data_product_name=data_product_name, table_name=table_name)
     logger.info(f"event: {event}")
 
-    schema_exists = DataProductSchema(
+    schema = DataProductSchema(
         data_product_name=data_product_name,
         table_name=table_name,
         logger=logger,
         input_data=None,
-    ).exists
+    )
+
+    schema_exists = schema.exists
 
     if not schema_exists:
         logger.error("No previous version of schema exists")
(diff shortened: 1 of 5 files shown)


containers/daap-update-schema/src/var/task/update_schema.py
```diff
@@ -19,12 +19,14 @@ def handler(event, context):
...
         data_product_name=data_product_name,
         table_name=table_name,
         logger=logger,
         input_data=None,
-    ).exists
+    )
+
+    schema_exists = schema.exists
 
     if not schema_exists:
```
tom-webber:
```suggestion
    if not schema.exists:
```

containers/daap-update-schema/src/var/task/update_schema.py
```diff
@@ -56,6 +60,18 @@ def handler(event, context):
...
-    return format_response_json(
-        status_code=HTTPStatus.OK, body={"version": new_version, "changes": changes}
-    )
+    # check if major or minor
+    if not int(new_version.split(".")[0][-1]) == int(schema.version.split(".")[0][-1]):
+        body_dict = {
+            "version": new_version,
+            "increment_type": "major",
+            "changes": changes,
+            **copy_response,
```
tom-webber, line 67: Should we change `VersionManager.update_schema` to return this dictionary, given all of this information is handled as part of the `update_schema` function anyway?

LavMatt: Yeah possibly. My thoughts on it were that it should be done as a wider review/refactor of all responses. I think we'll need to ensure better consistency as we've been doing our own and not following a strict framework (i haven't at least...), other than to be informative/useful responses.

tom-webber (approved):
LGTM

PR #2651, opened 2023-12-04
Kyverno airflow prod assets

jhpyke:
Applying the same fixes to Prod as exist in dev. This requires defining and importing additional infra compared to dev to achieve.


Gary-H9 (approved):
LGTM

PR #2725, opened 2023-12-12
grant sean AP access

Emterry:
Allow Sean access to AP aws


julialawrence (changes requested):
Could you do me a favour and add Sean adhoc to the accounts rather than the team, since that grants additional access. He already should have access to the labs one, so could you add him to all ap accounts?

julialawrence (approved):
Looks good, thank you!

calumabarnett (approved):
Happy to approve. Should this be in the data engineering access file though?

PR #2785, opened 2023-12-18
Add MadeTech engineers

calumabarnett:
This will give data engineers from MadeTech access to the data engineer role in the data production AWS account.

I (@calumabarnett) have confirmed their SC clearance status and that they have signed the PNC user agreement.


jacobwoffenden (dismissed):

jhpyke (approved):
Looks good to go

PR #2831, opened 2024-01-08
Tried to use CuratedDataCopier to copy old table data

murdo-moj:


containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -169,6 +169,25 @@ def update_metadata_remove_schemas(self, schema_list: list[str]) -> str:
...
+            logger=self.logger,
+            input_data=None,
+        )
+
+        # Populate the new versions of the athena tables with the old data
+        CuratedDataCopier(
+            column_changes={
+                "types_changed": False
+            },  # This forces all tables to be copied
+            new_schema=self.schema_object,  # dont want this either
```
murdo-moj, line 184:
```suggestion
            new_schema=self.schema_object,
```

PR #2836, opened 2024-01-08
886 static analysis - analytical-platform-development-cluster

BrianEllwood:
I have worked through the checkov issues for analytical-platform-development-cluster applying actions as in the spread sheet.  Some checks have been skipped and others code has been amended to pass the check.

Checkov scan results:

Passed checks: 267, Failed checks: 0, Skipped checks: 30

The terraform plan Plan: 1 to add, 12 to change, 0 to destroy.

aws_kms_key.eks in kms-keys.tf was failing checks as the KMS key had no Policy defined, so i created one which is the 1 to create in the terraform plan. - please ensure I have coded this correctly.

CKV_AWS_26 - Ensure all data stored in the SNS topic is encrypted - for aws_sns_topic  in sns-topics.tf the fix i implemented was to add this to the resource 'kms_master_key_id = "alias/aws/sns" '  I am not sure this is correct 

The rest of the changes are adding descriptions


BrianEllwood:
to address the aws_sns_topic requiring encryption with CMK i have tried to create a aws_iam_policy_document in main.tf, this needs the following in data.tf

data "aws_caller_identity" "target" {
  provider = aws.target
}

Unfortunately when added it results in 

erraform validate
╷
│ Error: Provider configuration not present
│ 
│ To work with data.aws_caller_identity.target its original provider configuration at provider["registry.terraform.io/hashicorp/aws"].target is required, but it has been removed. This occurs
│ when a provider configuration is removed while objects created by that provider still exist in the state. Re-add the provider configuration to destroy data.aws_caller_identity.target, after
│ which you can remove the provider configuration again.
╵
@julialawrence

BrianEllwood:
Following discussion in 22/1/24 standup it was decided to ignore the trivy alert "SNS topic not encrypted with CMK"  until we  investigate the down stream implications of implementing CMK.

I have removed all code relating to CMK from the PR.

BrianEllwood:
I have also dismissed the 4 trivy alerts raised for "SNS topic not encrypted with CMK"

terraform/aws/analytical-platform-development/cluster/efs-access-points.tf
```diff
@@ -3,6 +3,8 @@
 ##################################################
 
 resource "aws_efs_access_point" "eks_user_homes" {
+  #checkov:skip=CKV_AWS_329: requires a destroy to implement
+  #checkov:skip=CKV_AWS_330: requires ClientRootAccess IAM permission to implement
```
jacobwoffenden, line 7: Where does this come from? I'm not saying we should address this, just curious if this was an error back from Terraform

terraform/aws/analytical-platform-development/cluster/eks-cluster.tf
jacobwoffenden, line 1: Unrelated to this work, but we should look at removing Karpenter

terraform/aws/analytical-platform-development/cluster/iam-policies.tf
```diff
@@ -76,6 +76,7 @@ resource "aws_iam_policy" "cluster_autoscaler" {
 ##################################################
 
 data "aws_iam_policy_document" "efs_csi_driver" {
+  #checkov:skip=CKV_AWS_356: skip requires access to multiple resources
```
jacobwoffenden: As we have only one EFS instance, we could scope this down

```
resources = [aws_efs_file_system.eks_user_homes.arn]
```

terraform/aws/analytical-platform-development/cluster/karpenter.tf
jacobwoffenden, line 1: Again, Unrelated to this work, but we should look at removing Karpenter

terraform/aws/analytical-platform-development/cluster/efs-access-points.tf
BrianEllwood: No error from Terraform - just alerts from checkov 

when i tried the fix for CKV_AWS_329 the plan came back with a destroy which we agreed that we would not implement any destructive fixes and skip

As for CKV_AWS_330 requires "Enforcing a user identity is subject to the ClientRootAccess IAM permission." which we agreed we do not want to do

terraform/aws/analytical-platform-development/cluster/iam-policies.tf
BrianEllwood: change implemented

BrianEllwood: fix as suggested implemented

jacobwoffenden (approved):

PR #2855, opened 2024-01-10
Dp 2549 move remaining data after deleting a schema

murdo-moj:
- After a schema is deleted, copy any remaining data to the tables in the new major version which is created.
- I've basically implemented a switch in `CuratedDataCopier` for the case of when a schema is deleted. A more thorough refactor of this object might be needed into `CuratedDataCopierForSchemaUpdates` and `CuratedDataCopier`
- Since the data copy involves calls to Athena which can't be tested easily, I've stubbed all the calls to s3 in the tests which test the deletion of schemas

### DIFF
--- a/containers/daap-python-base/CHANGELOG.md
+++ b/containers/daap-python-base/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Add logic to dopy data to new tables after a schema gets deleted.
+- Add logic to copy data to new tables after a schema gets deleted.
 
 ## [8.0.1] - 2023-01-08
 
--- a/containers/daap-python-base/CHANGELOG.md
+++ b/containers/daap-python-base/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [8.1.1] - 2023-01-10
+
+### Changed
+
+- Add logic to dopy data to new tables after a schema gets deleted.
+
 ## [8.0.1] - 2023-01-08
 
 ### Changed
--- a/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
+++ b/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
@@ -125,14 +125,14 @@ class CuratedDataCopier:
         logger=DataPlatformLogger,
         schemas_to_copy=None,
         column_changes=None,
-        schema_delete: bool=False,
+        schema_delete: bool = False,
     ):
         """
         Copy data from existing version of data product to new version of
         data product.
         """
         self.schema = new_schema if new_schema else None
-        self.data_product_name = self.schema.data_product_name
+        self.data_product_name = element.data_product.name
         self.column_changes = column_changes
         self.new_curated_data_product_path = element.curated_data_prefix.uri
         self.new_database_name = element.database_name
--- a/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
+++ b/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
@@ -1,6 +1,6 @@
 from data_platform_logging import DataPlatformLogger
 from data_platform_paths import DataProductElement, QueryTable
-from data_product_metadata import DataProductSchema, format_table_schema
+from data_product_metadata import DataProductSchema, DataProductMetadata, format_table_schema
 from glue_and_athena_utils import (
     create_glue_database,
     refresh_table_partitions,
@@ -118,26 +118,29 @@ class CuratedDataLoader:
 class CuratedDataCopier:
     def __init__(
         self,
-        column_changes,
-        new_schema: DataProductSchema,
         element: DataProductElement,
         athena_client,
         glue_client,
         logger=DataPlatformLogger,
+        schemas_to_copy=None,
+        column_changes = None,
+        new_schema: DataProductSchema = None,
+        schema_delete: bool = False,
     ):
         """
         Copy data from existing version of data product to new version of
         data product.
         """
-        self.data_product_name = new_schema.data_product_name
-        self.column_changes = column_changes
         self.schema = new_schema
-        self.element = element
+        self.data_product_name = self.schema.data_product_name
+        self.column_changes = column_changes
         self.new_curated_data_product_path = element.curated_data_prefix.uri
         self.new_database_name = element.database_name
+        self.schemas_to_copy = schemas_to_copy
         self.glue_client = glue_client
         self.athena_client = athena_client
         self.logger = logger
+        self.schema_delete = schema_delete
         self._get_tables_to_copy()
 
     def _is_updated_table_valid_for_copy(self):
@@ -151,7 +154,7 @@ class CuratedDataCopier:
         Data types changed -> Not OK
         """
 
-        if self.column_changes["types_changed"]:
+        if self.column_changes is not None and self.column_changes["types_changed"]:
             return False
         else:
             return True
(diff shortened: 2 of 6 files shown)


containers/daap-python-base/CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [8.1.1] - 2023-01-10
+
+### Changed
+
+- Add logic to dopy data to new tables after a schema gets deleted.
```
MatMoore: dopy -> copy

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -169,6 +173,33 @@ def update_metadata_remove_schemas(self, schema_list: list[str]) -> str:
...
+                    table_name=current_metadata["schemas"][0],
+                    logger=self.logger,
+                    input_data=None,
+                )
+                .load()
+                .latest_version_saved_data
+            )
+            schema_to_keep = DataProductSchema(
+                data_product_name=self.data_product_name,
+                table_name=updated_metadata.data["schemas"][0],
```
MatMoore, line 190: I think we need to check here that there are any schemas left after removing the schema. If there isn't, we can skip this whole step.

containers/daap-python-base/tests/unit/data_platform_api_responses_test.py
```diff
@@ -27,33 +27,33 @@ def test_format_error_response():
     }
 
 
-def test_response_status_400():
-    response = data_platform_api_responses.response_status_400("something went wrong")
+# def test_response_status_400():
+#     response = data_platform_api_responses.response_status_400("something went wrong")
```
MatMoore: how come these are commented?

containers/daap-python-base/tests/unit/versioning_test.py
```diff
@@ -544,14 +547,27 @@ def setup(
                 )
 
     @pytest.fixture(autouse=True)
-    def setup_subject(self, glue_client, data_product_name):
+    def setup_subject(self, glue_client, athena_client, data_product_name):
         with patch("glue_and_athena_utils.glue_client", glue_client):
-            self.version_manager = VersionManager(data_product_name, logger)
+            with patch("glue_and_athena_utils.athena_client", athena_client):
+                athena_client.create_work_group(Name="data_product_workgroup")
```
murdo-moj, line 553: I might not need the workgroup call now I'm stubbing calls to athena

containers/daap-python-base/CHANGELOG.md
murdo-moj:
```suggestion
- Add logic to copy data to new tables after a schema gets deleted.
```

containers/daap-python-base/tests/unit/versioning_test.py
```diff
@@ -595,14 +619,35 @@ def test_deleted_schema_files_removed_from_new_version(
...
+                )
+
         schema_prefix = f"{self.data_product_name}/v3.0/{schema_list[0]}/schema.json"
         self.assert_object_count(self.bucket_name, schema_prefix, 0)
 
     def test_deleted_table_removed_from_new_version(self):
         schema_list = ["schema0"]
 
-        self.version_manager.update_metadata_remove_schemas(schema_list=schema_list)
+        with patch(
```
MatMoore, line 639: [Testing suggestion]

I think in this test we the `VersionManager` is calling `CuratedDataCopier`, which is calling `CuratedDataLoader`, which in tern calls these functions you are mocking out. So we have quite a lot of code under test and we are mocking dependencies of dependencies.

The danger here is that the test is likely to break if there are any changes to that dependency chain.

It may be worth instead mocking out the direct CuratedDataCopier instead (providing that it's well tested itself) to make the test less sensitive to implementation details and remove these nested `patch` calls.

(It's been a while since I've looked at this so I'm not sure if there was a reason we wrote the tests like this to start with)

containers/daap-python-base/tests/unit/versioning_test.py
murdo-moj:
```suggestion
                
```

containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -169,6 +173,33 @@ def update_metadata_remove_schemas(self, schema_list: list[str]) -> str:
             new_version=new_version,
         )
 
+        # Move any data across
+        if current_metadata["schemas"]:
+            input_data = format_table_schema(
```
MatMoore, line 178: I was wondering if there was a way we could change the `DataProductCopier` interface in a way that would eliminate passing any mandatory schema arguments here (e.g. by passing in `current_metadata` instead)?

This jumps out at me because the choice of schema here feels arbitrary - it has nothing to do with the actual operation we just did.

🤔 I will take a closer look at this and see if I can suggest something more concrete...

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
```diff
@@ -166,15 +167,16 @@ def run(self) -> list[DataProductSchema]:
...
-        parquet_table_input = format_parquet_glue_table_input(
-            self.schema.data["TableInput"],
-            self.new_curated_data_product_path,
-        )
-        self.glue_client.create_table(
-            DatabaseName=self.new_database_name,
-            TableInput=parquet_table_input,
-        )
+        if self.column_changes is not None:
+            create_glue_database(self.glue_client, self.new_database_name, self.logger)
```
MatMoore, line 171: Is there a reason we can't have this class create the new database in all cases, and remove the database creation from L170 in versioning.py?

Otherwise, the behaviour is a bit hard to predict just based on the arguments the class is initialised with.

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
```diff
@@ -221,7 +223,9 @@ def _get_tables_to_copy(self):
         """
         Creates a list of tables to copy into new version of data product database
         """
-        if self._is_updated_table_valid_for_copy():
+        if self.schemas_to_copy:
+            self.tables_to_copy = self.schemas_to_copy
+        elif self._is_updated_table_valid_for_copy():
             self.tables_to_copy = self.schema.parent_data_product_metadata["schemas"]
```
MatMoore, line 229: I think the `schemas_to_copy` parameter is a good addition here 👍🏻 

[Refactoring suggestion]

If you make this param required then, you could update the code from delete schema to pass `tables_to_copy=new_schema.parent_data_product_metadata["schemas"]`, and then you would be able to remove the conditional logic here

containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py
```diff
@@ -151,7 +152,7 @@ def _is_updated_table_valid_for_copy(self):
         Data types changed -> Not OK
         """
 
-        if self.column_changes["types_changed"]:
+        if self.column_changes is not None and self.column_changes["types_changed"]:
```
MatMoore, line 155: This could be equivalently written

```python
if self.column_changes is None or not self.column_changes["types_changed"]:
  return True
else:
  return False
```

not sure which is clearer, but the logic seems sound 👍🏻

MatMoore (commented):
Nice - I think this will let us reuse that copy class, and we have separated the data copying from the metadata copying, which is good.

I haven't approved just yet as I'd like to dig into the code some more to see if I can come up with a simpler/more intuitive way to pass parameters between the VersionManager and DataCopier classes. Have left a few suggestions inline though.

I *think* that it's functionally correct, but it's probably worth getting another pair of eyes on this, as this versioning logic has been quite tricky to get right.

containers/daap-python-base/tests/unit/data_platform_api_responses_test.py
murdo-moj: These are tests for functions which are deprecated, so running them just causes a deprecationwarning

containers/daap-python-base/src/var/task/versioning.py
murdo-moj: L177 should have that logic, I did come across that case while testing. However in that case we still want a database to be created, so if the database creation is inside CuratedDataCopier, then it needs to be called even if there aren't any tables left to populate with previous data.

LavMatt (dismissed):

LavMatt (approved):

PR #2902, opened 2024-01-12
Implement DataHub Catalogue client for dp-catalogue library

### LINKED ISSUE #2779: Connect registration APIs to catalogue
registering data product
registering metadata
exclude update

Abstract base class - daughter OMD / DataHub exports

DOD 
Use the python  library and prove end to end working using cmd

tom-webber:
- Convert CatalogueClient class into ABC base, so OMD and DataHub client classes can be inherited
- implement `__init__` and `upsert_table` for DataHubCatalogueClient using DataHub gms
- add tests to check DataHub client is submitting requests as expected (in the format supported by DataHub python sdk gms emitter).
- add packages to pypoetry.toml
- refactoring

## Further changes - @MatMoore 
- get tests working with a snapshot testing approach (based on DataHub unit tests)
- add integration test for datahub server
- further refactoring to make the interface catalogue-agnostic
- allow the dataset to be created within a container for its database (e.g. my_data_product_v1)
- remove dummy data from platform schema
- update readme & changelog


python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
```diff
@@ -79,13 +178,13 @@ class CatalogueClient:
     def __init__(
         self,
         jwt_token,
-        api_uri: str = "https://catalogue.apps-tools.development.data-platform.service.justice.gov.uk/api",
+        api_url: str = "https://catalogue.apps-tools.development.data-platform.service.justice.gov.uk/api",
```
PriyaBasker23: Can this be moved to config file

python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
```diff
@@ -201,47 +307,125 @@ def get_user_id(self, user_email: str):
...
 
-        return user_id
+    If there is a problem communicating with the catalogue, methods will raise an instance of
+    CatalogueError.
+    """
+
+    def __init__(
+        self,
+        jwt_token,
+        api_url: str = "https://datahub.apps-tools.development.data-platform.service.justice.gov.uk/api/gms",
```
PriyaBasker23: Can this be moved to config file

python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
jemnery: Is this a config which sets env vars? Environment variables could make it easier to deploy a different `api_url` (etc) to different environments via GHA in future.

tom-webber: This is just a default argument. The actual value used for client setup would be from usage of this catalogue package, and could be set by an env value.

python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
```diff
@@ -382,17 +395,36 @@ def create_or_update_schema(self, metadata: DataProductMetadata, database_fqn: s
...
+            dataset_urn: the dataset URN
         """
         dataset_schema_properties = SchemaMetadataClass(
             schemaName=metadata.name,
-            platform=make_data_platform_urn("glue"),
-            version=1,
+            platform=make_data_platform_urn(platform=platform),
+            version=version,
             hash="",
             platformSchema=OtherSchemaClass(rawSchema="__insert raw schema here__"),
```
MatMoore: this field is still a placeholder so I think we should either remove it (if not needed) or populate it

PriyaBasker23: yes , was hoping to mention the same

PriyaBasker23: what is  hash="" - used for ?. Can we remove any parameters if its not useful

tom-webber: `platformSchema` is not an optional field, and needs to be populated. Unfortunately I'm not sure _what_ it needs to be populated _with_.

Same for `hash`.

python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
tom-webber: default string has been removed.

python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
tom-webber: default string has been removed.

python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
MatMoore: hash is "the SHA1 hash of the schema content" and platformSchema is "The native schema in the dataset's platform." according to https://datahubproject.io/docs/generated/metamodel/entities/dataset/#schemametadata

Don't know what we would actually pass though, or if it matters that much. We already have the native types passed through for each field, so maybe these can both be left as empty strings?

python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
```diff
@@ -201,47 +310,277 @@ def get_user_id(self, user_email: str):
...
+            platform=platform, name=f"{metadata.name}", env="PROD"
+        )
+
+        metadata_event = MetadataChangeProposalWrapper(
+            entityType="dataset",
+            changeType=ChangeTypeClass.UPSERT,
+            entityUrn=dataset_urn,
+            aspect=dataset_schema_properties,
+        )
+        self.graph.emit(metadata_event)
```
MatMoore: ⚠️ From integration testing this, it looks like the `SchemaPropertiesClass` we pass here is not enough for the dataset to show up in the UI. I *think* we need a DatasetPropertiesClass aspect as well, i.e. this snippet

```python
dataset_properties = DatasetPropertiesClass(
    description="Example table for testing",
    customProperties={"dpia_required": "yes"},
)

# Construct a MetadataChangeProposalWrapper object.
metadata_event = MetadataChangeProposalWrapper(
    entityUrn=builder.make_dataset_urn(
        "athena", "athena.catalog.hmpps_use_of_force_v2.statements"
    ),
    aspect=dataset_properties,
)
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client.py
```diff
@@ -65,9 +121,55 @@ class ReferencedEntityMissing(CatalogueError):
...
+    @abstractmethod
+    def upsert_database(self, *args: Any, **kwargs: Any):  # type: ignore[override]
+        pass
+
+    @abstractmethod
+    def upsert_schema(self, *args: Any, **kwargs: Any):  # type: ignore[override]
+        pass
+
+    @abstractmethod
+    def upsert_table(
```
MatMoore: I don't think the use of `*args` and `**kwargs` is doing what you expect here.

Since this is an abstract base class, we would expect to be able to type any value to `BaseCatalogueClient` and then be able to substitute any of the concrete classes without breaking things. But this is not the case, since the base class accepts *any* arguments, and the concrete classes do not.

(See [Liskov's substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle))

I think what you want to do instead is write out the full arglist here, and then use *args, **kwargs in the subclasses only when you want to ignore some of them.

MatMoore:
I have tested using this branch to create a dataset on the apps and tools instance. See https://datahub.apps-tools.development.data-platform.service.justice.gov.uk/dataset/urn:li:dataset:(urn:li:dataPlatform:glue,my_data_product2.my_table3,PROD)/Schema?is_lineage_mode=false&schemaFilter=

I think I'm reasonably confident it's working end to end now, although there are some things we might need to come back to.

Notes:

- Custom variables are not implemented yet - so expecting we'll need to add these and any other missing fields later
- We have to pass tags at the table level now, since DatahubCatalogueClient is not passing along any tags from the Data Product (and either way I don't believe the dataset would inherit this from the data product).
- <strike>With the DatahubCatalogueClient, you need to pass the name as $database_name.$table_name, if you want the database to show up as a container. Otherwise it sticks everything in a "Default" container.</strike> This is no longer necessary
- In datahub entities have multiple "aspects". We were creating an aspect for the schema, and tags, but we were missing the one for basic properties associated with the dataset, so I've added that in.

python-libraries/data-platform-catalogue/data_platform_catalogue/client/openmetadata.py
```diff
@@ -201,42 +210,22 @@ def get_user_id(self, user_email: str):
...
-        """
-        raise NotImplementedError
+        user = self.metadata.get_by_name(entity=User, fqn=username)
+        if user is not None:
+            return user.id
+        else:
+            raise ReferencedEntityMissing
 
     def _generate_tags(self, tags: list[str]):
+        # TODO? update using the sdk logic:
```
MatMoore, line 220: There is an unaddressed TODO here which I'm just going to leave since it's the OpenMetadata implementation, and we don't know if we'll be returning to this.

MatMoore:
Here is a table created with the latest code:

https://datahub.apps-tools.development.data-platform.service.justice.gov.uk/dataset/urn:li:dataset:(urn:li:dataPlatform:glue,test_data_product_v2.test_table,PROD)/Schema?is_lineage_mode=false&schemaFilter=

Note that I have fully qualified the database name here (database.table), which groups everything from the same database in a container that shows up in the breadcrumb, and in the datahub search interface.
Screenshot 2024-01-18 at 15 25 36

We can change this later if we don't like it, but this keeps it consistent with what we were doing before, in the OpenMetadata implementation.

PriyaBasker23 (approved):

PriyaBasker23:
> Here is a table created with the latest code:
> 
> https://datahub.apps-tools.development.data-platform.service.justice.gov.uk/dataset/urn:li:dataset:(urn:li:dataPlatform:glue,test_data_product_v2.test_table,PROD)/Schema?is_lineage_mode=false&schemaFilter=
> 
> Note that I have fully qualified the database name here (database.table), which groups everything from the same database in a container that shows up in the breadcrumb, and in the datahub search interface. Screenshot 2024-01-18 at 15 25 36
> 
> We can change this later if we don't like it, but this keeps it consistent with what we were doing before, in the OpenMetadata implementation.

It would be good to map tables into its own database

julialawrence (approved):
Priya approved.

PR #2995, opened 2024-01-18
Dp 2977

murdo-moj:
- Copy curated data in s3 between versions upon table deletion

### DIFF
--- a/containers/daap-python-base/src/var/task/versioning.py
+++ b/containers/daap-python-base/src/var/task/versioning.py
@@ -213,12 +213,14 @@ class VersionManager:
             # Curated data does not have the minor version in it's directory structure
             latest_major_version = self.latest_version.split(".")[0]
             new_major_version = new_version.split(".")[0]
-            self._copy_data_from_previous_version(
-                bucket=bucket,
-                source_folder=source_folder,
-                latest_version=latest_major_version,
-                new_version=new_major_version,
-            )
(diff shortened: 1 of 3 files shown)


containers/daap-python-base/src/var/task/versioning.py
```diff
@@ -200,6 +207,21 @@ def update_metadata_remove_schemas(self, schema_list: list[str]) -> str:
                 logger=self.logger,
             ).run()
 
+            # Copy data files in the curated bucket
```
MatMoore, line 210: I thought this wasn't needed because it was already handled by the athena query in CuratedDataCopier. 

Rather than copy the all the parquet files in s3 we have an unload query that creates the new parquet from the latest load timestamp in the previous version. Am I misunderstanding how this works?

```sql
           UNLOAD (
                SELECT
                    *
                FROM {previous_major_database}.{curated_table.name}
                WHERE load_timestamp = (
                    SELECT MAX(load_timestamp)
                    FROM {previous_major_database}.{curated_table.name}
                )
            )
            TO '{self.table_path}'
            WITH(
                format='parquet',
                compression = 'SNAPPY',
                partitioned_by=ARRAY['load_timestamp']
            )
```

trigger from https://github.com/ministryofjustice/data-platform/blob/main/containers/daap-python-base/src/var/task/curated_data/curated_data_loader.py#L215

PR #3051, opened 2024-01-23
[DP-2983] Add a method for search

MatMoore:
Part of https://github.com/ministryofjustice/data-platform/issues/2983

This method accepts a query and pagination variables, and returns the total number of results plus the results to display on the current page.

For each search result, return:

- the result type: Table or Data Product
- ID, name and description
- tags
- a dictionary of additional metadata fields (exactly what this will contain will depend on what tests well with users)
- information about the properties which matched
- the time the metadata was last updated

The raw responses from Datahub are logged at DEBUG level.

For now I've excluded filters, facets, and sorting but these are supported by the underlying API.

See
- https://datahubproject.io/docs/graphql/inputObjects#facetfilterinput
- https://datahubproject.io/docs/graphql/objects#facetmetadata
- https://datahubproject.io/docs/graphql/inputObjects#searchsortinput

### DIFF
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql/search.graphql
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql/search.graphql
@@ -1,11 +1,11 @@
-query Search($query: String!, $count: Int!, $start: Int!) {
+query Search(
+  $query: String!
+  $count: Int!
+  $start: Int!
+  $types: [EntityType!]
+) {
   searchAcrossEntities(
-    input: {
-      types: [DATASET, DATA_PRODUCT]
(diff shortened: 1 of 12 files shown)


python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql/search.graphql
```diff
@@ -0,0 +1,152 @@
...
+              key
+              value
+            }
+            created
+            lastModified
+          }
+          editableProperties {
+            description
+          }
+          browsePathV2 {
```
MatMoore: This is not being returned at the moment

murdo-moj (approved):

PR #3075, opened 2024-01-25
[DP-3074] Add filtering to search function

### LINKED ISSUE #3074: Add filters to the search function
# Context
We have a UI component for filters from https://design-patterns.service.justice.gov.uk/components/filter/

Currently we are just showing the example and it doesn't do anything. So we should make it possible to apply the filters when calling the search method of the catalogue client.

# In scope
- Decide what should be filterable
- Make it possible to pass in filters
- Include the filters in the GraphQL query
- (Optional) The search query can return possible facet information via https://datahubproject.io/docs/graphql/objects#facetmetadata - can we use this to extract the possible filter values and the number of results with each value?

# Out of scope for now
- Accessible autocomplete to narrow down the filter options

MatMoore:
Add a way to filter by domain, tags, glossaryTerms, or customProperties when running a search query.

Resolves #3074

This adds a filter argument to the search function, and a facets attribute to the search response, mirroring the filters/facets objects in the underlying GraphQL API, and restricting to a fixed set of fields.

Example usage:

```python
response = client.search()

# Total results across all pages
print(response.total_results)

# Iterate over search results
for item in response.page_results:
  print(item)

# Iterate over facet options
for option in response.facets['domains']:
  print(option)

# Filter by domain
client.search(filters=[MultiSelectFilter("domains", [response.facets['domains'][0].value])])
```

Facet information will be returned for everything except customProperties.

### DIFF
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client/base.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client/base.py
@@ -8,7 +8,7 @@ from ..entities import (
     DataProductMetadata,
     TableMetadata,
 )
-from ..search_types import ResultType, SearchResponse
+from ..search_types import MultiSelectFilter, ResultType, SearchResponse
 
 logger = logging.getLogger(__name__)
 
@@ -65,6 +65,7 @@ class BaseCatalogueClient(ABC):
             ResultType.DATA_PRODUCT,
             ResultType.TABLE,
         ),
+        filters: Sequence[MultiSelectFilter] = (),
     ) -> SearchResponse:
         """
         Wraps the catalogue's search function.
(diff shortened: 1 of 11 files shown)


python-libraries/data-platform-catalogue/data_platform_catalogue/client/base.py
```diff
@@ -8,7 +8,7 @@
     DataProductMetadata,
     TableMetadata,
 )
-from ..search_types import ResultType, SearchResponse
+from ..search_types import MultiSelectFilter, ResultType, SearchResponse
```
tom-webber, line 11: ⛏️ I appreciate it's an absolute (pardon the pun) minefield, but do we want to standardise relative vs. absolute imports within the package?

Here it's relative, [here it's absolute](https://github.com/ministryofjustice/data-platform/pull/3075/files#diff-f425ca2ec2ca477a29960cc7cde6b0d8aab55fa17eaae0d62de6ce17c0e21fefR11).

tom-webber (dismissed):
Looks good!

tom-webber (approved):
lgtm

PR #3132, opened 2024-01-30
add vpc flow logs as part of trivy remediation

Emterry:
this pr adds the vpc flow logs for the trivy remediation as identified in the analytical-platform-data-engineering-production/10ds component , this issue is linked to #886

PR #3140, opened 2024-01-31
kms-key-hotfix

Emterry:
This pr is a hot fix for #3132 - failed on merge to create log group so have amended to use arn instead of id


Gary-H9 (approved):
LGTM

PR #3141, opened 2024-01-31
DP-886/static analysis dev control panel message broker

AntFMoJ:
Resolved checkov failures by:
- replacing wildcard with ARNs for the SQS queues
- adding argument `sqs_managed_sse_enabled = true` explicitly for each queue as SSE-SQS in use.


michaeljcollinsuk (approved):
LGTM

PR #3142, opened 2024-01-31
Dp 3127 add sorting in search

murdo-moj:


python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
```diff
@@ -63,6 +65,9 @@ def search(
             "filters": formatted_filters,
         }
 
+        if sort:
+            variables.update({"sort": sort})
```
MatMoore: I think this is wrong actually... we need to map the `SortOption` to the format GraphQL wants, e.g. {"sortCriterion": {"field": field, "sortOrder": "ASCENDING"}}

MatMoore (approved):

PR #3144, opened 2024-01-31
DP 886/static analysis control panel message broker

AntFMoJ:
Removed resources from aws_iam_policy_document.source


julialawrence (approved):
Let's see how this goes. :)

PR #3145, opened 2024-01-31
DP-886 / static analysis control panel message broker dev

AntFMoJ:
Simplified SQS IAM policies down to one access policy used by the three SQS resource policies.


Gary-H9 (approved):
LGTM

PR #3151, opened 2024-02-01
3009 static analysis

BrianEllwood:
skips and remediation work for module analytical-platform-data-production-airflow

As this contains both dev and prod i have only remediated for dev and if the approach for dev is approved i will implement for prod.

I have made a number of changes to code to remediate the alerts please ensure I have used the correct method.


terraform/aws/analytical-platform-data-production/airflow/buckets.tf
```diff
@@ -1,6 +1,22 @@
...
+  server_side_encryption_configuration {
+    rule {
+      apply_server_side_encryption_by_default {
+        sse_algorithm     = "aws:kms"
+        kms_master_key_id = aws_kms_key.s3_mojap_airflow_dev.arn
+      }
+    }
+  }
+
 }
```
jhpyke: Had something like this come up with Ant on some of the SQS queues. The actual state of the asset is that it is already encrypted, but with `Amazon S3 managed keys (SSE-S3)`. As such I think the change we'd want is just to update the terraform to reflect that rather than re-encrypting with a customer managed key.

terraform/aws/analytical-platform-data-production/airflow/eks.tf
```diff
@@ -1,4 +1,11 @@
 resource "aws_eks_cluster" "airflow_dev_eks_cluster" {
+  encryption_config {
+    resources = ["secrets"]
+    provider {
+      key_arn = aws_kms_key.eks.arn
+    }
+  }
+
```
jhpyke, line 8: In this case, I agree that encryption is missing and should be added if feasible. Would like to test in sandpit ideally first? I know that's not in terraform though, so would be somewhat manual testing.

terraform/aws/analytical-platform-data-production/airflow/eks.tf
```diff
@@ -9,14 +16,18 @@ resource "aws_eks_cluster" "airflow_dev_eks_cluster" {
   ]
   version = "1.25"
 
+
   vpc_config {
-    subnet_ids          = aws_subnet.dev_private_subnet[*].id
-    public_access_cidrs = ["0.0.0.0/0"]
-    security_group_ids  = [var.dev_cluster_additional_sg_id]
+    subnet_ids             = aws_subnet.dev_private_subnet[*].id
```
jhpyke: Would be good to chat about this change - looking at the subnets live I'm seeing `/24` ranges on them currently.

terraform/aws/analytical-platform-data-production/airflow/network.tf
```diff
@@ -137,6 +139,7 @@ resource "aws_cloudwatch_log_group" "airflow_dev_vpc_flow_log" {
   name              = "airflow-dev-vpc-flow-log"
   retention_in_days = 400
   skip_destroy      = true
+  kms_key_id        = aws_kms_key.cloudwatch.arn
```
jhpyke, line 142: Don't currently encrypt these logs streams, agree we should, need to test whether this affects the ability of the Airflow UI to surface those logs to users.

jhpyke (changes requested):
Looks broadly what I'd expect, with a couple of comments about areas where I think we can avoid needing to make as many changes.

terraform/aws/analytical-platform-data-production/airflow/buckets.tf
BrianEllwood: yes - recommendation was to skip

terraform/aws/analytical-platform-data-production/airflow/eks.tf
BrianEllwood: yes lets test in sandpit

terraform/aws/analytical-platform-data-production/airflow/eks.tf
BrianEllwood: amended an will push up change

terraform/aws/analytical-platform-data-production/airflow/network.tf
BrianEllwood: test in sandpit

PR #3206, opened 2024-02-05
Dp-3048-add-metadata-items

murdo-moj:
* Added data product level and table level metadata items
* Integrated these metadata items to the datahub client
* I added rowCount just to the datahub client. This needs to be updated every time data is written to a data product, or a job could update it weekly


python-libraries/data-platform-catalogue/tests/test_client_openmetadata.py
```diff
@@ -85,10 +88,18 @@ def data_product(self):
             description="bla bla",
             version="v1.0.0",
             owner="2e1fa91a-c607-49e4-9be2-6f072ebe27c7",
+            owner_display_name="April Gonzalez",
+            maintainer="j.shelvey@digital.justice.gov.uk",
+            maintainer_display_name="Jonjo Shelvey",
             email="justice@justice.gov.uk",
+            status=DataProductStatus.DRAFT,
```
murdo-moj, line 95:
```suggestion
            status=DataProductStatus["DRAFT"],
```

python-libraries/data-platform-catalogue/tests/test_entities.py
```diff
@@ -9,10 +16,18 @@ def data_product():
         description="bla bla",
         version="v1.0.0",
         owner="2e1fa91a-c607-49e4-9be2-6f072ebe27c7",
+        owner_display_name="April Gonzalez",
+        maintainer="j.shelvey@digital.justice.gov.uk",
+        maintainer_display_name="Jonjo Shelvey",
         email="justice@justice.gov.uk",
+        status=DataProductStatus.DRAFT,
```
murdo-moj, line 23:
```suggestion
        status=DataProductStatus["DRAFT"],
```

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -50,22 +66,41 @@ def from_data_product_metadata_dict(metadata: dict, version, owner_id: str):
             description=metadata["description"],
             version=version,
             owner=owner_id,
+            owner_display_name=metadata["dataProductOwnerDisplayName"],
```
mitchdawson1982, line 71: `metadata.get("dataProductOwnerDisplayName")`

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -50,22 +66,41 @@ def from_data_product_metadata_dict(metadata: dict, version, owner_id: str):
...
+            maintainer=metadata.get("dataProductMaintainer", None),
```
mitchdawson1982: None is the default returned if the key is not found. `metadata.get("dataProductMaintainer")`

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -50,22 +66,41 @@ def from_data_product_metadata_dict(metadata: dict, version, owner_id: str):
...
+            maintainer_display_name=metadata.get(
+                "dataProductMaintainerDisplayName", None
+            ),
```
mitchdawson1982: None is the default returned if the key is not found. `metadata.get("dataProductMaintainerDisplayName"),`

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
murdo-moj, line 70:
```suggestion
            maintainer=metadata.get("dataProductMaintainer"),
```

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
murdo-moj: I haven't done a get call with mandatory fields. We could add verification that mandatory fields are filled?

MatMoore: I think this is probably fine for now, if we call it without the mandatory fields we will get a KeyError and it should be easy to debug.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -50,22 +66,39 @@ def from_data_product_metadata_dict(metadata: dict, version, owner_id: str):
...
+            owner_display_name=metadata["dataProductOwnerDisplayName"],
+            maintainer=metadata.get("dataProductMaintainer"),
+            maintainer_display_name=metadata.get("dataProductMaintainerDisplayName"),
             email=metadata["email"],
+            status=DataProductStatus[metadata["status"]],
             retention_period_in_days=metadata["retentionPeriod"],
             domain=metadata["domain"],
             dpia_required=metadata["dpiaRequired"],
+            dpia_location=metadata.get("dpiaLocation"),
+            last_updated=datetime.strptime(metadata["lastUpdated"], "%Y%m%d"),
```
MatMoore: Just to check, have you tested this is the format datahub expects for this field? Just wondering because I encountered another timestamp that expected an integer somewhere.

Also, can you extract the date format string to a constant? e.g. DATAHUB_DATE_FORMAT = '%Y%m%d'

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -50,22 +66,39 @@ def from_data_product_metadata_dict(metadata: dict, version, owner_id: str):
...
+            creation_date=datetime.strptime(metadata["creationDate"], "%Y%m%d"),
+            s3_location=metadata.get("s3Location"),
             tags=metadata.get("tags", []),
         )
 
         return new_metadata
 
 
+class DataSensitivityLevel(Enum):
```
MatMoore: Could we call this `SecurityClassification` or `GovernmentSecurityClassification` or something like that? Just to reuse existing terminology and avoid confusion with the "sensitive" marking which we might also want to add

https://www.gov.uk/government/publications/government-security-classifications/government-security-classifications-policy-html#additional-markings

MatMoore (dismissed):
Looks good to me, just some minor suggestions - take em or leave em

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
murdo-moj: These are being added as custom fields and not as part of the datahub model. I think it might come out a bit weird because of the datetimes but it should be accepted as a string like `datetime.datetime(2021, 1, 1, 0, 0)`

MatMoore (approved):

PR #3218, opened 2024-02-06
Bumping replica count to 2

### LINKED ISSUE #3208: 🐛 Create-A-Derived-Table Action Queue is unacceptably long
### Describe the bug.

In create-a-derived-table, action run times can [stretch to upwards of an hour](https://github.com/moj-analytical-services/create-a-derived-table/actions/runs/7757848739) despite only [taking a matter of minutes](https://github.com/moj-analytical-services/create-a-derived-table/actions/runs/7757848739/usage) to run. This is due to long queue times when multiple PRs are made, and if many pushes are made to said PR branches.

### To Reproduce

1. Go to the Create-A-Derived-Table repo
2. Open a number of PRs with changes
3. Cause a mashup for the `deploy-dev` action
4. Get stuck queuing for ages

### Expected Behaviour

Multiple PRs should be able to use actions simultaneously.

### Additional context

Recommend Scaling runner count from 1 to 2. This should help alleviate the worst of the queuing at peak times. Could autoscale but this may be overkill.

jhpyke:
Reserving the right to push to 3 later depending on user outcomes.

### DIFF
--- a/charts/actions-runner/CHANGELOG.md
+++ b/charts/actions-runner/CHANGELOG.md
@@ -8,12 +8,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [2.1.0] - 2024-02-06
-
-### Changed
-
-- Increases replica count to 2.
-
 ## [2.0.0] - 2023-12-18
 
 ### Changed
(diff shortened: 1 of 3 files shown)


jacobwoffenden (changes requested):
This sets the chart default, do we just want to override CADT's deployment instead?

jhpyke:
> This sets the chart default, do we just want to override CADT's deployment instead?

Will change to that now

jacobwoffenden (approved):
LGTM

PR #3292, opened 2024-02-12
Update pagerduty/teams.tf

tom-webber:
add Tom to Data platform pagerduty rota


jacobwoffenden (dismissed):
@tom-webber this will add you to "our" team in PagerDuty, so you'll be on our rota, is that what you want?

Now DPAT has renamed to AP, some work needs to happen to align across the systems

Also, the plan is failing because you need adding in `users.tf`

Gary-H9 (approved):
LGTM

PR #3304, opened 2024-02-13
:sparkles: Add subdomain field to dataproduct metadata

mitchdawson1982:
- Removes Open Metadata code
- Adds subdomain property
- Updates domain references to `LAA` and subdomain to `Legal Aid`


python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -232,7 +242,7 @@ def upsert_table(
             customProperties={
                 "sourceDatasetName": metadata.source_dataset_name,
                 "sourceDatasetLocation": metadata.source_dataset_location,
-                "sensitivityLevel": metadata.data_sensitivity_level,
+                "sensitivityLevel": str(metadata.data_sensitivity_level),
                 "rowCount": "1177",
```
LavMatt: We've renamed source dataset location so may as well add that in here
```suggestion
                "whereToAccessDataset": "Analytical Platform",
                "sensitivityLevel": metadata.data_sensitivity_level.name,
                "rowCount": "1177",

File: python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py [hunk 2]
```diff
@@ -104,7 +106,7 @@ class TableMetadata:
     retention_period_in_days: int | None
     source_dataset_name: str | None = None
     source_dataset_location: str | None = None
```
LavMatt: This will need changing to the new name on line 133 too
```suggestion
    where_to_access_dataset: str | None = None

File: python-libraries/data-platform-catalogue/tests/test_integration_with_datahub_server.py [hunk 3]
```diff
@@ -8,6 +8,7 @@
 """
```
LavMatt, line 9: Will need to swap out `source_dataset_location` for `where_to_find_dataset` in this test file too

python-libraries/data-platform-catalogue/tests/test_integration_with_datahub_server.py
```diff
@@ -37,7 +38,8 @@ def test_upsert_test_hierarchy():
         email="justice@justice.gov.uk",
         status=DataProductStatus.DRAFT,
```
LavMatt:
```suggestion
        status=DataProductStatus.DRAFT.name,

File: python-libraries/data-platform-catalogue/CHANGELOG.md [hunk 5]
```diff
@@ -7,6 +7,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.13.0] 2024-02-14
+
+### Changed
+
+- Added subdomain property to dataproduct
```
LavMatt, line 14: Might want to note removal of openmetadata too

LavMatt (commented):
LGTM - just a minor thing regarding renaming of a metadata property (`source_dataset_location` or `sourceDatasetLocation` to `where_to_access_dataset` or `whereToAccessDataset`)  we did yesterday because it was a bit misleading. Think i've indicated everywhere it's used.

Does the removal of openmetadata mean we can use later versions of python now too?

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -151,41 +151,51 @@ def upsert_data_product(self, metadata: DataProductMetadata):
 
         data_product_urn = "urn:li:dataProduct:" + "".join(name.split())
 
-        if metadata.domain:
```
tom-webber, line 154: This was originally here to prevent domain being assigned if it was an empty string. 

Agree that this isn't the place for that test. But we'll need some validation in the entities section below.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -75,6 +76,7 @@ def from_data_product_metadata_dict(metadata: dict, version, owner_id: str):
             status=DataProductStatus[metadata["status"]],
             retention_period_in_days=metadata["retentionPeriod"],
             domain=metadata["domain"],
```
tom-webber, line 76: Need to ensure ~domain isn't an empty string~ _**either domain or subdomain**_ isn't an empty string (a more difficult problem)

Maybe we can make use of [pydantic (or pydantic style) validators](https://docs.pydantic.dev/latest/concepts/validators/), e.g. `min_length=1` for string values.

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -151,41 +151,51 @@ def upsert_data_product(self, metadata: DataProductMetadata):
...
-            domain = metadata_dict.pop("domain")
-            domain_urn = self.graph.get_domain_urn_by_name(domain_name=domain)
+        domain = metadata_dict.pop("domain")
+        domain_urn = self.graph.get_domain_urn_by_name(domain_name=domain)
```
tom-webber, line 155: Do we want to sentence case or lowercase `domain`? Is `get_domain_urn_by_name` case sensitive?

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -151,41 +151,51 @@ def upsert_data_product(self, metadata: DataProductMetadata):
...
-            metadata_event = MetadataChangeProposalWrapper(
-                entityType="dataproduct",
-                changeType=ChangeTypeClass.UPSERT,
-                entityUrn=data_product_urn,
-                aspect=data_product_domain,
-            )
-            self.graph.emit(metadata_event)
-            logger.info(f"Data Product {name} associated with domain {domain}")
+        subdomain_urn = None
+        if metadata.subdomain:
```
tom-webber, line 162: All of the `domain` logic should be an `elif`/`else` for if this conditional is false, as you're not using `domain`/`domain_urn` if `subdomain` exists.

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
mitchdawson1982: Included

python-libraries/data-platform-catalogue/tests/test_integration_with_datahub_server.py
mitchdawson1982: refactored

python-libraries/data-platform-catalogue/CHANGELOG.md
mitchdawson1982: Included

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
mitchdawson1982: Agreed, however Domain is mandatory so shouldn't be an empty value.

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
tom-webber: I'd keep `whereToAccessDataset` as `metadata.where_to_access_dataset` as you've specified in `entities.py`.

I'd provide "Analytical Platform" as a default suggestion in a UI dropdown, but I'm wary of adding it here as we don't know what other data people may add

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
mitchdawson1982: Domain is mandatory so should never be an provided as an empty string but this should be validated as you say. Perhaps one for another ticket.

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
mitchdawson1982: Domains are currently hard coded in the front end, users dont have the ability to enter them.

python-libraries/data-platform-catalogue/tests/test_integration_with_datahub_server.py
mitchdawson1982: Done. should be `where_to_access_dataset`

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
mitchdawson1982: Done

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
tom-webber: Nope this is wrong. Please ignore. I missed that you're creating a parent-child relationship for any new subdomain, which requires the `domain_urn`

mitchdawson1982: domain & domain_urn is needed when subdomain is present in the metadata but not yet created, so there would still be duplication of the calls. Happy to chat about moving to another ticket.

MatMoore: @tom-webber I think if subdomain is passed in we may still create the domain if it doesn't exist - otherwise we can't use it as the parent of the subdomain...

I think the scenarios are something like this:

| | domain passed in without subdomain | domain & subdomain passed in | 
| -- | -- | -- | 
| subdomain exists | n/a | do nothing |
| domain exists | do nothing  | create subdomain |
| nothing exists | create domain | create domain and subdomain |

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -231,8 +241,8 @@ def upsert_table(
             description=metadata.description,
             customProperties={
                 "sourceDatasetName": metadata.source_dataset_name,
-                "sourceDatasetLocation": metadata.source_dataset_location,
-                "sensitivityLevel": metadata.data_sensitivity_level,
+                "whereToAccessDataset": metadata.where_to_access_dataset,
+                "sensitivityLevel": str(metadata.data_sensitivity_level),
```
tom-webber:
```suggestion
                "sensitivityLevel": metadata.data_sensitivity_level.name,
```

MatMoore: Can do the same thing for the data product custom properties as well - I think we're just stringifying them atm

tom-webber: There's an existing type incompatibility here as these are optional arguments (can be `None`), but `customProperties` is expecting `dict[str, str]`. 

We should change the metadata fields for use in the `customProperties` dict in `entities.py` from `str | None = None` to `str = ""`

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -103,8 +103,8 @@ class TableMetadata:
     column_details: list
     retention_period_in_days: int | None
     source_dataset_name: str | None = None
-    source_dataset_location: str | None = None
-    data_sensitivity_level: SecurityClassification = SecurityClassification["OFFICIAL"]
+    where_to_access_dataset: str | None = None
```
tom-webber:
```suggestion
    source_dataset_name: str = ""
    where_to_access_dataset: str = ""
```
type error fix from above

mitchdawson1982: Done

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
mitchdawson1982: Done

mitchdawson1982: Done

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -128,7 +128,7 @@ def from_data_product_schema_dict(
             column_details=metadata["columns"],
             retention_period_in_days=retention_period,
             source_dataset_name=metadata.get("sourceDatasetName"),
-            source_dataset_location=metadata.get("sourceDatasetLocation"),
+            where_to_access_dataset=metadata.get("sourceDatasetLocation"),
```
tom-webber:
```suggestion
            source_dataset_name=metadata["sourceDatasetName"],
            where_to_access_dataset=metadata["sourceDatasetLocation"],
```
to fix linting

mitchdawson1982: Have passed an empty string to dict.get(str_param, "")

MatMoore (approved):

PR #3381, opened 2024-02-16
Bug fix catalogue name and page search

### LINKED ISSUE #3328: 🐛 Our python datahub search client does not return correct page search results
### Describe the bug.

The [start](https://github.com/ministryofjustice/data-platform/blob/main/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py#L63) parameter in the graphql query of our search client's search method, had until now assumed to be start page and passing a page number would return the results of size `count` from that page.

start actually refers to the index of individial serach results so to get page would be 
```
"start": count * page,
```
need to implement the fix in the catalogue package.

This bug is presenting itself in the paginated results of `find-moj-data`

### To Reproduce

1. make a search with more than 20 results
2. got to page 2
3. see top result is 2nd from top result from previous page

### Expected Behaviour

subsequent pages of search results display unique results. Should match the page results as seen in datahub UI

### Additional context

_No response_

LavMatt:
fixes a few minor bugs (fixes #3328 and others without ticket)

- search page corrected as `start` refers to individual result index rather than result page
- dataset name was not added to datahub - corrected
- data product assets were duplicated when appended to existing assets

### DIFF
--- a/python-libraries/data-platform-catalogue/CHANGELOG.md
+++ b/python-libraries/data-platform-catalogue/CHANGELOG.md
@@ -7,6 +7,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.14.0] 2024-02-19
+
+### Changed
+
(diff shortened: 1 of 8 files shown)


tom-webber (approved):
LGTM

MatMoore (approved):
Nice 👍🏻

PR #3396, opened 2024-02-20
Add list data product to catalogue client

LavMatt:
This adds a method to call the [`listDataProductAssets`](https://datahubproject.io/docs/graphql/queries/#listdataproductassets) graphql query from our client. 

Returns all data product children in a single call. Required for better performance of the data product details page in `find-moj-data`


python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql/listDataProductAssets.graphql
```diff
@@ -0,0 +1,148 @@
+query search($urn: String!, $query: String!, $count: Int!, $start: Int!) {
```
MatMoore: The `search` here is just a name. I'm not sure where this would actually be surfaced but we should change it to reflect what the query does, just to avoid confusion with the other query

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql/listDataProductAssets.graphql
```diff
@@ -0,0 +1,148 @@
...
+              urn
+              id
+              properties {
+                name
+                description
+              }
+            }
+          }
+        }
+        ... on DataProduct {
```
MatMoore: Don't think we need this block as a DataProduct will never be returned by this query

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
```diff
@@ -127,6 +133,41 @@ def search_facets(
...
+            "query": "*",
+            "start": start,
+            "count": count,
+        }
+        try:
+            response = self.graph.execute_graphql(
+                self.data_product_asset_query, variables
+            )
+        except GraphError as e:
+            raise Exception("Unable to execute listDataProductAssets query") from e
```
MatMoore, line 153: Just a thought - do you think it's worth introducing a custom exception type for these errors?

They can be thrown any time the catalogue is unavailable

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
```diff
@@ -127,6 +133,41 @@ def search_facets(
...
+        page_results = []
+        for result in response["listDataProductAssets"]["searchResults"]:
+            entity = result["entity"]
+            entity_type = entity["type"]
+            matched_fields = {
```
MatMoore: this isn't returned by this query is it? If not we can drop it

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
```diff
@@ -127,6 +133,41 @@ def search_facets(
...
+                i["name"]: i["value"] for i in result.get("matchedFields", [])
+            }
+            if entity_type == "DATASET":
+                page_results.append(self._parse_dataset(entity, matched_fields))
+            else:
+                raise ValueError(f"Unexpected entity type: {entity_type}")
```
MatMoore, line 162: Will we get different entity types coming back when we start adding charts to data products?

If so we might want to filter these out for now, rather than throw an error

MatMoore (changes requested):
Looks good, just have some minor suggestions

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql/listDataProductAssets.graphql
LavMatt: yeah good point, will change

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
LavMatt: Will double check, and drop if not returned

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
LavMatt: Yeah was wondering here as similar to search method. As we're not sure how to parse other result types yet i thought it be difficult to catch and could add other types in as and when we add new entity types to datahub.

Could be another ticket to make a more comprehensive `ResultType` enum and ensure our search client methods handle each type?

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
LavMatt: is used in `_parse_dataset` so rather than refactor that i'll just pass it an empty dict

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
LavMatt: Yeah quite probably, but another separate ticket

MatMoore (approved):

PR #3411, opened 2024-02-22
Catalogue high level diagram

jemnery:
The aim is to create something which strikes a balance between being immediately understandable to a newcomer, while also capturing enough detail to be useful in understanding how it all fits together and is deployed.

Data product registration and data uploading are currently out of scope for the diagram, and intentionally not included.

Image can be viewed in the repo here: https://github.com/ministryofjustice/data-platform/blob/catalogue-diagram/docs/source/documentation/diagrams/catalogue.excalidraw.png

Or if checking out the branch, use VS Code ([Excalidraw extension](https://marketplace.visualstudio.com/items?itemName=pomdtr.excalidraw-editor) required)

Note the PR also includes a couple of PNG diagrams lifted straight from the DataHub website. for extra context & detail.


MatMoore:
Looks good to me. Only thing I would suggest adding is an arrow from user -> find-moj-frontend.

We could optionally show github on here as it's a repository for data product metadata, but I think that will just make the diagram more confusing.

MatMoore (approved):

murdo-moj:
I'm unsure except intuitively what the arrows mean in the catalogue diagram (mainly for the AP/Datahub connection)

is the directionality of the arrow from AP/Datahub metadata ingestion from Datahub and to the AP as it's a job which pulls in data from AP?
or is it from AP to Datahub as that's where the data flows?

PR #3449, opened 2024-02-26
Find-moj-data-83/glossary

murdo-moj:
- a get_glossary method in the datahub client and SearchClient
- There's less flexibility with the query to get glossary terms, and I've set the default terms to return at 1000. I think this should be fine for our use case?
- I haven't created a custom class for `GlossaryTermMetadata` since those classes eg `TableMetadata` are used on set methods, which we don't have any for glossary terms. Perhaps it would be more extensible to introduce that class now though?

### DIFF
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
@@ -358,7 +358,7 @@ class SearchClient:
             last_updated=None,
         )
 
-    def get_glossary_terms(self, count: int = 1000)-> SearchResult:
+    def get_glossary_terms(self, count: int=1000)-> SearchResponse:
         "Get some number of glossary terms from DataHub"
         variables = {"count": count}
         try:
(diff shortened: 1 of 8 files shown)


python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
```diff
@@ -335,3 +342,39 @@ def _parse_facets(self, facets: list[dict[str, Any]]) -> SearchFacets:
...
+            result_type=ResultType.GLOSSARY_TERM,
+            matches={},
+            name=properties["name"],
+            description=properties.get("description", ""),
+            metadata=metadata,
+            tags=[],
+            last_updated=None,
+        )
+
+    def get_glossary_terms(self, count: int = 1000) -> SearchResponse:
```
LavMatt, line 361: Wondering if it might be worth including a page parameter here as glossary terms could possibly be something that wants to be paginated as it grows? Same as search method where start = page*count. 

It could also wait until pagination developed in the front end. You've done it intentionally on a single page for now anyway

LavMatt (approved):
LGTM - One minor comment but can be picked up another time if you prefer, as there are other methods that don't implement paginated results either

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
murdo-moj: I'll wait, just because it's not clear how a larger glossary would be represented on the frontend eg it could be all on one page with a good nav bar/hyperlinks

Issue #2502, opened 2023-11-24
🚸 Process & Code for Separate Image Build Repos


julialawrence:
### User Story

As the outcome of [this discussion](https://github.com/ministryofjustice/data-platform/discussions/2163) Data Platform engineers would like to implement a process, define a  template and provide documentation for anyone who wishes to manage their image build via a separate repo rather than using the data platform monorepo.

### Value / Purpose

The plusses and minuses of managing image builds outside the monorepo are outlined in the discussion linked in the previous section, and there is at least a minimal interest in taking advantage of this approach, 

### Useful Contacts

@julialawrence

### User Types

Data Platform Engineers

### Hypothesis

If we introduce a process as well as tools to manage repos outside DP monorepo we will minimise the issues that commonly arise out of proliferation of project repositories. 

### Proposal

1. Define a template for the image build repo
2. Define a process for creating a repo and managing it in code
3. Raise an example PR
4. Produce documentation on best end state and required controls

### Definition of Done

- [x] Docker Image build repo template defined
- [x] Process defined and documented
- [x] Example PR raised and example repo created.
- [x] Documentation updated.

Emterry:
update intial repo added https://github.com/ministryofjustice/analytical-platform-image-build-template

Emterry:
example repo manually created and test image built

PR #3451, opened 2024-02-26
added analytical-platform-image-build-template repo

Emterry:
- added terraform code
- create analytical-platform-image-build-template repo
- as per issue: #2502

### DIFF
--- a/terraform/github/analytical-platform-repositories.tf
+++ b/terraform/github/analytical-platform-repositories.tf
@@ -453,8 +453,8 @@ locals {
         pushers = [module.data_platform_team.id]
       }
     },
-    "analytical-platform-image-builder" = {
-      name                                   = "analytical-platform-image-builder"
+    "analytical-platform-image-build-template" = {
+      name                                   = "analytical-platform-image-build-template"
       description                            = "Analytical Platform Image Build Template"
       is_template                            = true
       use_template                           = false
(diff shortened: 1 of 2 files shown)


terraform/github/analytical-platform-repositories.tf
```diff
@@ -452,6 +452,20 @@ locals {
         admins  = [module.data_platform_teams["data-platform-apps-and-tools"].id]
         pushers = [module.data_platform_team.id]
       }
+    },
+    "analytical-platform-image-build-template" = {
+      name                                   = "analytical-platform-image-build-template"
+      description                            = "Analytical Platform Image Build Template"
+      is_template                            = true
+      use_template                           = false
```
julialawrence: We should use the default MOJ repo template for this.

terraform/github/analytical-platform-repositories.tf
```diff
@@ -452,6 +452,20 @@ locals {
...
+      visibility                             = "public"
+      advanced_security_status               = "disabled"
```
julialawrence: Should we be turning this on or not needed for a template?

julialawrence (changes requested):
Suggest using the moj repo template as a basis.

terraform/github/analytical-platform-repositories.tf
Emterry: ive removed it as the docs says If a repository's visibility is public, advanced security is always enabled and cannot be changed

julialawrence (approved):
Good start. :)

PR #3468, opened 2024-02-28
use fully qualified name in upsert_table

LavMatt:
Fully qualified name is now loaded to datahub alongside the name property which is just the name of a table/dataset entity

This will allow us to better display fully qualified name on search results and just the simple table name on the data product details page


MatMoore:
Can we also pass the displayName property here?

murdo-moj (dismissed):

MatMoore (approved):

PR #3542, opened 2024-03-04
Add a new repo for spiking a custom ingestion source

MatMoore:
We have a spike to investigate ingesting dashboards into the data catalogue.

As part of this I want to try and write a custom ingestion source that pulls chart metadata from the Justice Data API.

### DIFF
--- a/terraform/github/data-catalogue.tf
+++ b/terraform/github/data-catalogue.tf
@@ -33,6 +33,23 @@ module "data_catalogue_repository" {
   }
 }
 
+module "datahub_custom_api_source" {
+  source = "./modules/repository"
+
+  name        = "data-catalogue"
+  description = "Custom ingestion source for Datahub to consume from https://data.justice.gov.uk/api"
+  topics      = ["data-catalogue"]
+
+  use_template        = true
+  template_repository = "data-platform-app-template"
+  has_projects        = "true"
+  homepage_url        = null
+
+  access = {
+    admins = [module.data_catalogue_team.id]
+  }
+}
+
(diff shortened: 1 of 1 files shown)


terraform/github/data-catalogue.tf
```diff
@@ -33,6 +33,23 @@ module "data_catalogue_repository" {
...
 
+module "datahub_custom_api_source" {
+  source = "./modules/repository"
+
+  name        = "data-catalogue"
+  description = "Custom ingestion source for Datahub to consume from https://data.justice.gov.uk/api"
+  topics      = ["data-catalogue"]
+
+  use_template        = true
+  template_repository = "data-platform-app-template"
```
jacobwoffenden: instead of using the app template, you can drop this off and it will use the MoJ default template

```suggestion
```

terraform/github/data-catalogue.tf
```diff
@@ -33,6 +33,23 @@ module "data_catalogue_repository" {
   }
 }
 
+module "datahub_custom_api_source" {
+  source = "./modules/repository"
+
+  name        = "data-catalogue"
```
jacobwoffenden: This needs updating to the name you want to use, and then you'll need to update the module name to be `new_name_repository`

jacobwoffenden (changes requested):

terraform/github/data-catalogue.tf
MatMoore: 🤦🏻  ah of course

jacobwoffenden (approved):

PR #3546, opened 2024-03-04
Replace data-catalogue with a new repo

MatMoore:
The goal of this PR is to:
- Rename `data-catalogue` back to `find-moj-data`
- Create a *new* `data-catalogue` repo

Since find-moj-data creates a build artefact, we've decided to manage it in its own separate repo. We are expecting this will make deployment easier because
- we won't need workflow logic to conditionally execute stuff depending on subdirectory
- deployments of the same artefact type will follow the same basic template
- we can diff versions of an artefact in source without seeing unrelated changes from other components

However we also want a "top level" repo that links out to the various artefacts. It is this repo that would take over the repository terraforming if we completely decouple our work from the analytical platform. Right now, we will use it for the helm deployment of datahub itself, and any documentation or tickets related to the overall data catalogue project.

## What we are aiming for
Screenshot 2024-03-04 at 11 54 06

## Consequences
- After merging this, we should partially revert https://github.com/ministryofjustice/cloud-platform-environments/pull/21391

### DIFF
--- a/terraform/github/data-catalogue.tf
+++ b/terraform/github/data-catalogue.tf
@@ -16,21 +16,21 @@ module "data_catalogue_team" {
   ]
 }
 
-# module "data_catalogue_parent_repository" {
-#   source = "./modules/repository"
+module "data_catalogue_repository" {
+  source = "./modules/repository"
 
-#   name        = "data-catalogue"
-#   description = "Data catalogue"
-#   topics      = ["data-catalogue"]
+  name        = "data-catalogue"
+  description = "Data catalogue"
+  topics      = ["data-catalogue"]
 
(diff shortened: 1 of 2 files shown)


terraform/github/data-catalogue.tf
```diff
@@ -16,23 +16,43 @@ module "data_catalogue_team" {
   ]
 }
 
-module "data_catalogue_repository" {
+# module "data_catalogue_parent_repository" {
```
MatMoore: Will add this in a follow up PR - not sure how the provider will handle it if I apply the move & recreate at the same time

Gary-H9 (approved):
LGTM

PR #3548, opened 2024-03-04
set branch_protection_require_signed_commits default to true

tom-webber:
This pull request:

- Defaults `branch_protection_require_signed_commits` to true
- Removes unused `branch_protection_push_restrictions` and `branch_protection_blocks_creations`
  - These were removed in https://github.com/ministryofjustice/data-platform/pull/3383, but not picked up by Super-Linter as Dependabot pull requests are excluded from linting 🦝


jacobwoffenden (dismissed):
As I'd mentioned on Slack, you need to set this too true as well

https://github.com/ministryofjustice/data-platform/blob/f6a7ff86898471a971d8945f9e40ee288145cbbe/terraform/github/modules/repository/variables.tf#L190-L193

julialawrence (approved):
Roll tide!

PR #3604, opened 2024-03-05
Add method for fetching schema metadata

MatMoore:
Part of https://github.com/ministryofjustice/find-moj-data/issues/102

In order to display table schemas we need a way to lookup dataset metadata in Datahub.

## Implementation details    
This method fetches the schemaMetadata aspect and transforms it into a simpler structure for presentation.

It returns a `TableMetadata` object, which we previously defined for ingesting table metadata. However I've diverged from the original schema we were following for the column list:

1. Here I return datahub's types rather than glue types. Previously we based our types on glue because it made the ingestion easier. Now that we are focusing on cataloguing independently of ingestion into s3 buckets, we can revisit this decision and present whatever makes the most sense to catalogue users. Rather than deal with this right now I've just gone with the easiest option and present what datahub presents.

2. Our original metadata definitions didn't take into account nullability, primary keys and foreign keys, which we know we will need. Datahub can store this metadata so I've included it here so we can easily present it in Find MOJ data.

## Not included
This method does not return the complete set of custom properties and ownership information that we need to render the full details page. I'm gonna address this in a follow up.


python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql_helpers.py
```diff
@@ -0,0 +1,136 @@
...
+    primary_keys = set(schema_metadata.get("primaryKeys", ()))
+
+    foreign_keys = defaultdict(list)
+
+    # Attempt to match foreign keys to the main fields.
+    #
+    # Assumptions:
+    # - A given field may have multiple foreign keys to other datasets
+    # - Some foreign keys will not match on fieldPath, because fields
+    #   may be defined using STRUCT types and foreign keys can refrence
```
murdo-moj:
```suggestion
    #   may be defined using STRUCT types and foreign keys can reference
```

murdo-moj:
Fair enough about the data types returned from columns being datahub types, I agree that seems fine for now.

murdo-moj (approved):

Issue #3482, opened 2024-02-28
🔑  ITHC Support


julialawrence:
### User Story

Analytical Platform is undergoing an IT Health Check from Monday, March 4th to Monday, March 11th. 
In order to get the most out of the exercise, we need to provide support (on occasion time-critical) to the testers. 

### Value / Purpose

The quality of the support will correlate to the quality of the check output and will result in minimum time lost. 

### Useful Contacts

@julialawrence

### Proposal

1. Attend daily standups
2. Monitor for support requests and action them promptly
3. Resolve any access issues

### Additional Information

Testing will be carried out during business hours of 8:00 - 17:00
No out of hours support will be required, requested or provided.

Outside of issues relating to service outages, for the duration of the ITHC, this is our priority. 

### Definition of Done

- [ ] Required support provided

julialawrence:
4/3 Recap:

Work stream 3
9:30 Got github accounts for today's activity
9:54 Onboarded github accounts and provisioned access
9:56 Jacob asked for clarification on what superuser accounts can do
9:57 Explained what Superuser accounts are able to do
9:58 Asked what regular accounts can do
**11:01** Explained what regular accounts can do
13:54 Jacob asked about working Grafana link
14:01 explained that grafana has been retired (Asked @jacobwoffenden to raise a ticket to roll those logs into Obs Plat)

julialawrence:
4/3 Recap Continued:

Work stream 1 (EKS)
10:54 Felix unable to find EKS cluster
11:04 We passed over an aws config file and the assume role urls for all accounts but didn't make clear on the naming which caused confusion. Naming clarified, issue resolved. 

No further issues raised.

julialawrence:
5/3 Recap:

Workstream 2 (AWS)
9:15 Standup: Lewis confirmed he was able to log in and asked for clarification on what account/region the services he is looking at reside in. Resoved in standup: confirmed account and region
14:09pm Lewis asked for confirmation about which account he is to look at
14:12pm Confirmed it's the above account (double-checked account via last four digits of account id)

julialawrence:
6/3 AM Recap:
Standup: No issues from NCC team, MOJ requested look at QS. NCC agreed if time allows, asked for QS onboarding help as getting permission error when trying to sign up
12:24 Let Lewis know he should be able to sign up for QS. 
12:40: Ack

julialawrence:
6/3 PM Recap (ongoing)
[...]

PR #3610, opened 2024-03-06
Adding permission to create QS Admin user to PenTester Role in APDP account.

julialawrence:
#3482 

Enabling ITHC testers to create users in order to assess our current QS offering.

Overriding static analysis because will be deleted after March 11th

Destruction of IAM keys is fine, they're being manually managed by the testers.


Gary-H9 (approved):
LGTM

PR #3612, opened 2024-03-06
Bug fix: Creating policy in the right account.

julialawrence:
Fixes the apply error in https://github.com/ministryofjustice/data-platform/pull/3610
Didnt use a provider for policy resource so it created in the landing account rather than APDP.

This is a fix.

Override static analysis because it's due to be retired on March 11th.

#3482


BrianEllwood (approved):
LGTM

PR #3639, opened 2024-03-07
add static method to return match keys dict with custom property logic

mitchdawson1982:
Add's a new static method to return matched fields with logic to handle fields within the custom property value.

### DIFF
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
@@ -99,9 +99,7 @@ class SearchClient:
         for result in response["searchResults"]:
             entity = result["entity"]
             entity_type = entity["type"]
-            matched_fields = {
-                i["name"]: i["value"] for i in result.get("matchedFields", [])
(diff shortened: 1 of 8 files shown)


python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
```diff
@@ -114,6 +112,18 @@ def search(
...
         )
 
+    @staticmethod
+    def _get_matched_fields(result: dict) -> dict:
+        fields = result.get("matchedFields", [])
+        matched_fields = {}
+        for field in fields:
+            name = field.get("name")
+            value = field.get("value")
+            if name == "customProperties" and value != "":
```
MatMoore, line 122: Can you add a test for this case? I think the rest of this function is tests already but not the special case for customProperties

mitchdawson1982: Updated the query test case with a custom property match field.

MatMoore (dismissed):

MatMoore (approved):

PR #3640, opened 2024-03-07
Add James to pager duty support rota

michaeljcollinsuk:
Add James to pager duty support rota


jacobwoffenden:
There is an issue in the dependency logic when adding new users, to workaround you need to apply locally

```bash
terraform apply -target='module.users'

terraform apply -target='module.teams'

terraform apply
```

jacobwoffenden (approved):

PR #3721, opened 2024-03-14
Amend QuickSight ADR to accepted

BrianEllwood:
As the decision has been made to use QuickSight this change amends the  ADR to reflect the accepted status


julialawrence (approved):
So it begins!

PR #3722, opened 2024-03-14
Add missing metadata when fetching datasets

MatMoore:
I noticed a few bits of metadata are not being passed through to the frontend correctly, so we render some missing values on the details page. This PR adds those in.

We are still missing some attributes we've previously defined like `source_dataset`, `where_to_access_dataset`, but we don't have that in the catalogue yet, so we'll have to handle that in another ticket (either populating or removing these fields).

### Added

- Return `qualifiedName` and `tags` for datasets if available

### Changed

- Use timezone aware timestamps when returning last updated values, and source
  the information from `lastModified` in Datahub, not `lastIngested`.
- Prefer native data types to Datahub's type when returning schemas for datasets

### DIFF
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
@@ -665,7 +665,7 @@ class DataHubCatalogueClient(BaseCatalogueClient):
             domain = parse_domain(response)
             owner, owner_email = parse_owner(response)
             tags = parse_tags(response)
-            name = properties["name"]
+            name = properties.get("name", response.get("name"))
             created, modified = parse_created_and_modified(properties)
 
             # A dataset can't have both a container and data product parent, but if we did
@@ -678,6 +678,8 @@ class DataHubCatalogueClient(BaseCatalogueClient):
                 relations = parse_relations(
                     RelationshipType.PARENT, response["data_product_relations"]
                 )
+            else:
+                relations = {}
             return TableMetadata(
                 name=name,
                 fully_qualified_name=properties.get("qualifiedName") or name,
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
@@ -664,6 +664,7 @@ class DataHubCatalogueClient(BaseCatalogueClient):
             domain = parse_domain(response)
             owner, owner_email = parse_owner(response)
             tags = parse_tags(response)
+            name = properties["name"]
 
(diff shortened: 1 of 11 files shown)


python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -121,13 +121,15 @@ class TableMetadata:
...
+    fully_qualified_name: str | None = None
     data_sensitivity_level: SecurityClassification = SecurityClassification.OFFICIAL
     tags: list[str] = field(default_factory=list)
     major_version: int = 1
     row_count: int | None = None
-    last_updated: datetime | None = None
     owner: str = ""
     owner_email: str = ""
+    last_updated: datetime | None = None
+    first_created: datetime | None = None
```
LavMatt, line 132: Not necessarily for this PR (could be a separate ticket) but i think we should start adding descriptions, using pydantic's `Field` possibly, to our entity metadata properties. They are growing quickly and some are a little bit open to interpretation, e.g. `first_created` could mean first created in the catalogue or first created as a table on the AP, or even when the table first was created at source

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -664,6 +665,8 @@ def get_table_details(self, urn) -> TableMetadata:
             domain = parse_domain(response)
             owner, owner_email = parse_owner(response)
             tags = parse_tags(response)
+            name = properties["name"]
```
LavMatt: This is throwing an `KeyError`when run for dbt dataset urn. It's something that has slipped through,`name` is not set within the `properties` key of the response for dbt ingestions. They have a `name` key/value at the top level of the response, however this is equivalent to a fully qualified name e.g. `awsdatacatalog.dbt_database.dbt_table`

LavMatt (commented):
Looks good. I've picked up one issue that is unrelated to the changes here and a comment on descriptions for our entity metadata properties. 

Could be handled in fresh tickets if preferable

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
MatMoore: Yeah 💯 I would be keen to redo all these model classes.

I'd be in favour of adopting pydantic for this, as then we can serialise/deserialise to JSON as well without additional code.

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
MatMoore: Ah, annoying. I've added a fallback to top level `name` to match the search response.

LavMatt (approved):

Issue #3700, opened 2024-03-13
:lock: ITHC Offboarding


julialawrence:
### User Story

Now that ITHC activities have concluded we need to offboard NCC users from the system

### Useful Contacts

Julia Lawrene, Tevin Jemide

### Additional Information

NCC Users in `moj-analytical-services`

CP Prod

* ncctest1
* ncctest2
* ncctest3
* ncctest5

CP Dev

* at-test-account-1
* at-test-account-2
* at-test-account-3
* at-test-account-4

### Definition of Done

- [x] ITHC [terraform component](https://github.com/ministryofjustice/data-platform/tree/main/terraform/aws/analytical-platform-landing-production/ithc) destroyed
- [x] Github Users removed from `moj-analytical-services`
- [x] NCC Users deleted from Development Control Panel
- [x] NCC Users deleted from Production Control Panel
- [x] Resources created by ITHC NCC contractors identified and deleted in Analytical Platform Data Production
- [x] Resources created by ITHC NCC contractors identified and deleted in Analytical Platform Development
- [x] [This PR](https://github.com/ministryofjustice/data-platform/pull/3537) reverted
- [x] Contents of #analytical-platform-ithc-2024 exported and channel archived

julialawrence:
Completed. Channel archived.

PR #3763, opened 2024-03-15
Removing PenTester from cluster-masters.

julialawrence:
As part of this story #3700, removing PenTester role as a cluster-master for the dev cluster.


BrianEllwood (approved):
LGM

PR #3767, opened 2024-03-15
Dc 175 add upsert athena methods

LavMatt:
This PR adds two new upsert methods to the python datahub client to ingest metadata (in an expected format) as Athena databases and tables. 

Is part of the worked needed for https://github.com/ministryofjustice/find-moj-data/issues/175

### This PR adds
- `upsert_athena_database` and `upsert_athena_table` methods to `DataHubCatalogueClient` - to register/update databases or tables in datahub. These methods do not create a domain if it does not exist as we want doains to be a controlled entity.
- `DatabaseMetadata` class to `entities.py` - for defining metadata for an  athena database 
- `DatabaseStatus` enum to `entities.py`
- 2 custom exceptions in `datahub_client.py` for invalid domains and missing metadata.
- `parent_database_name`, `domain`, `row_count` optional properties to`entities.TableMetadata` class.

### What the PR doesn't do
This PR does not go as far as it potentially could go in the space of entities (as we have defined them).

I think there is some wider thought needed around what properties our metadata entities contain in light of our current focus and with what we know now. I'd suggest this could form another ticket as a task of itself and should be informed by discussion and agreement within the group after reflection on the initial user research results. 

Because of my thinking here i have not overhauled the metadata entity classes as much as i could've done

### DIFF
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
@@ -201,29 +201,17 @@ class DataHubCatalogueClient(BaseCatalogueClient):
                 f"{domain} does not exist in datahub - please align data to an existing domain"
             )
 
-        subdomain_urn = None
-        if metadata.subdomain:
-            subdomain = metadata_dict.pop("subdomain")
-            subdomain_urn = self.graph.get_domain_urn_by_name(domain_name=subdomain)
+        database_domain = DomainsClass(domains=[domain_urn])
 
-            if subdomain_urn is None:
-                logger.info(f"creating new subdomain {domain} for {name}")
-                domain_urn = self.create_domain(
-                    domain=subdomain, parentDomain=domain_urn
-                )
-
-        database_domain = DomainsClass(
-            domains=[subdomain_urn if subdomain_urn else domain_urn]
-        )
-
-        metadata_event = MetadataChangeProposalWrapper(
-            entityType="container",
-            changeType=ChangeTypeClass.UPSERT,
-            entityUrn=database_urn,
-            aspect=database_domain,
-        )
-        self.graph.emit(metadata_event)
-        logger.info(f"Database {name} associated with domain {domain}")
+        if domain_urn is not None:
+            metadata_event = MetadataChangeProposalWrapper(
+                entityType="container",
+                changeType=ChangeTypeClass.UPSERT,
+                entityUrn=database_urn,
+                aspect=database_domain,
+            )
+            self.graph.emit(metadata_event)
+            logger.info(f"Database {name} associated with domain {domain}")
 
         database_properties = ContainerPropertiesClass(
             customProperties={key: str(val) for key, val in metadata_dict.items()},
@@ -664,3 +652,6 @@ class DataHubCatalogueClient(BaseCatalogueClient):
             )
         except GraphError as e:
             raise Exception("Unable to execute getDataset query") from e
+
+    def _does_domain_exist(domain_name: str) -> bool:
+        pass
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
@@ -30,6 +30,7 @@ from datahub.metadata.schema_classes import (
 
 from ...entities import (
     CatalogueMetadata,
+    DatabaseMetadata,
     DataLocation,
     DataProductMetadata,
     TableMetadata,
@@ -174,7 +175,7 @@ class DataHubCatalogueClient(BaseCatalogueClient):
 
         return domain_urn
 
-    def upsert_athena_database(self, metadata: DataProductMetadata) -> str:
+    def upsert_athena_database(self, metadata: DatabaseMetadata) -> str:
         """
         Define a database. Must belong to a domain
         """
--- a/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
+++ b/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
@@ -260,7 +260,7 @@ class DataHubCatalogueClient(BaseCatalogueClient):
         dataset_urn = mce_builder.make_dataset_urn(
             platform="athena", name=fully_qualified_name, env="PROD"
         )
-
+        # jscpd:ignore-start
         dataset_properties = DatasetPropertiesClass(
             name=metadata.name,
             qualifiedName=fully_qualified_name,
@@ -329,6 +329,7 @@ class DataHubCatalogueClient(BaseCatalogueClient):
                 aspect=tags_to_add,
             )
             self.graph.emit(event)
+        # jscpd:ignore-end
 
         database_urn = "urn:li:container:" + "".join(
             metadata.parent_database_name.split()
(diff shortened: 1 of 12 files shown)


python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -146,6 +176,212 @@ def create_domain(
 
         return domain_urn
 
+    def upsert_athena_database(self, metadata: DatabaseMetadata) -> str:
+        """
+        Define a database. Must belong to a domain
+        """
+        metadata_dict = dict(metadata.__dict__)
+        metadata_dict.pop("version")
```
murdo-moj, line 184: why are you popping these?

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -146,6 +176,212 @@ def create_domain(
 
         return domain_urn
 
+    def upsert_athena_database(self, metadata: DatabaseMetadata) -> str:
```
MatMoore, line 179: I'm wondering if we can make this method support arbitrary containers, not just athena dbs? 

The only athena specific part seems to be the platform, which we could make an argument.

In a previous PR I added a [DataLocation class](https://github.com/ministryofjustice/data-platform/blob/3bcdf8dd49a6b0b6089b3df1ef81873f48289788/python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py#L18) thinking that this could represent all info about source platform and location. I don't know if this is still the best way of modelling it, but it might work here.

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -146,6 +176,212 @@ def create_domain(
...
+        """
+        Define a database. Must belong to a domain
+        """
+        metadata_dict = dict(metadata.__dict__)
```
MatMoore, line 183: How about adding a helper method to the `DatabaseMetadata` class to return the metadata dict? Then we wouldn't need to interact with the __dict__ directly and we could shorten this method

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -146,6 +176,212 @@ def create_domain(
...
+        name = metadata_dict.pop("name")
+        description = metadata_dict.pop("description")
+
+        database_urn = "urn:li:container:" + "".join(name.split())
+
+        domain = metadata_dict.pop("domain")
+        domain_urn = self.graph.get_domain_urn_by_name(domain_name=domain)
+
+        # domains are to be controlled and limited so we dont want to create one
+        # when it doesn't exist
```
MatMoore, line 197: makes sense - this will simplify things a bit I think

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
LavMatt: Similar to the data product upsert methods these are popped because they're not used. One for the wider review of metadata entities mentioned in the PR comment i think

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -146,6 +176,212 @@ def create_domain(
...
+        metadata_event = MetadataChangeProposalWrapper(
+            entityUrn=database_urn,
+            aspect=SubTypesClass(typeNames=[DatasetContainerSubTypes.DATABASE]),
+        )
+        self.graph.emit(metadata_event)
+        logger.info(f"Type updated for Database {name} ")
+
+        return database_urn
+
+    def upsert_athena_table(
```
MatMoore, line 250: This seems similar to `upsert_table`. Can we combine the two?

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -136,3 +145,34 @@ def from_data_product_schema_dict(
...
+    email: str
+    retention_period_in_days: int
+    domain: str
+    subdomain: str | None
+    dpia_required: bool
+    dpia_location: str | None
+    last_updated: datetime
+    creation_date: datetime
+    s3_location: str | None
+    status: DataProductStatus = DataProductStatus.DRAFT
```
MatMoore, line 172: Is this meant to be `DatabaseStatus` rather than `DataProductStatus`?

murdo-moj (approved):

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -146,6 +176,212 @@ def create_domain(
...
+            aspect=database_properties,
+        )
+        self.graph.emit(metadata_event)
+        logger.info(f"Properties updated for Database {name} ")
+
+        # set platform to athena
+        metadata_event = MetadataChangeProposalWrapper(
+            entityUrn=database_urn,
+            aspect=DataPlatformInstance(
+                platform="urn:li:dataPlatform:athena",
```
MatMoore, line 234: Have you looked into using platform instances here, rather than the predefined ones? https://datahubproject.io/docs/platform-instances/

We could potentially have a distinct platform specifically for AP

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -146,6 +176,212 @@ def create_domain(
...
+        """
+        metadata_dict = dict(metadata.__dict__)
+        metadata_dict.pop("version")
+        metadata_dict.pop("owner")
+        metadata_dict.pop("tags")
+
+        name = metadata_dict.pop("name")
+        description = metadata_dict.pop("description")
+
+        database_urn = "urn:li:container:" + "".join(name.split())
```
MatMoore, line 191: It might be safer to use Datahub's own methods for building URNs. Just makes it easier to keep up to date if they ever change the format. 

I'm also wondering if we need to pass the ENV component of the URN here - so far everything we've created goes in PROD, but I'm not sure if we are going to need to catalogue preprod metadata at any point.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -102,11 +108,14 @@ class TableMetadata:
     description: str
     column_details: list
     retention_period_in_days: int | None
+    domain: str | None = None
```
MatMoore, line 109: Do you think we still need the Data Product upsert method now? We wrote it with the assumption we were building a new data platform ingestion mechanism, and I think some of the assumptions are no longer valid. Datahub lets you manage data products as yaml out of the box without writing any python code, so I feel like this is not adding anything at the moment.

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -146,6 +176,212 @@ def create_domain(
...
+            ],
+        )
+
+        metadata_event = MetadataChangeProposalWrapper(
+            entityType="dataset",
+            changeType=ChangeTypeClass.UPSERT,
+            entityUrn=dataset_urn,
+            aspect=dataset_schema_properties,
+        )
+        self.graph.emit(metadata_event)
```
MatMoore, line 311: If we start building more ingestion libraries, we can perhaps lift and shift this code

LavMatt:
In the interest of time for the next round of user research we will pick up (resolved) comments from @MatMoore around structure and refactoring in subsequent ticket. https://github.com/ministryofjustice/find-moj-data/issues/187

murdo-moj (approved):

PR #3782, opened 2024-03-19
create datahub assumable role in ap prod

tom-webber:
Creates a role for DataHub to assume to allow metadata ingestion via dbt (CaDeT), Glue, and Athena ingestion sources from AP prod account.
- add minimal set of datahub glue + athena ingestion policies 
- athena datasets & query results access 
- CaDeT bucket read access 
- glue dataset & job read access 
- specify register-my-data buckets for athena locations 
- specify all environments irsa arns for role assumption

### Of note
The Glue permissions are broad, but read-only, so we can grab metadata for all items in the catalogue.

### DIFF
--- a/terraform/aws/analytical-platform-data-production/tooling-iam/terraform.tfvars
+++ b/terraform/aws/analytical-platform-data-production/tooling-iam/terraform.tfvars
@@ -9,17 +9,17 @@ account_ids = {
 }
 
 data_buckets = [
-    "moj-reg-dev-curated",
-    "moj-reg-preprod-curated",
-    "moj-reg-prod-curated"
-    ]
+  "moj-reg-dev-curated",
+  "moj-reg-preprod-curated",
+  "moj-reg-prod-curated"
+]
 
 athena_query_result_buckets = ["aws-athena-query-results-593291632749-eu-west-1"]
 
 datahub_cp_irsa_arns = {
-    dev = "arn:aws:iam::754256621582:role/cloud-platform-irsa-33e75989394c3a08-live"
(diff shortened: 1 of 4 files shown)


terraform/aws/analytical-platform-data-production/tooling-iam/terraform.tfvars
```diff
@@ -5,6 +5,21 @@
 account_ids = {
   analytical-platform-data-production       = "593291632749"
   analytical-platform-management-production = "042130406152"
+  cloud-platform                            = "754256621582"
+}
+
+data_buckets = [
```
julialawrence, line 11: Where are these managed?

tom-webber: The `data_buckets`? They're the register-my-data buckets

julialawrence (changes requested):
Dev, pre-prod, and prod policies need to have separate permissions allowing access to respective resources and allow assumptions from respective CP roles. They shouldn't rely on this all-in-one approac. Could I please ask you to rewrite?

terraform/aws/analytical-platform-data-production/tooling-iam/terraform.tfvars
tom-webber: They're managed here by the looks of it: https://github.com/ministryofjustice/register-my-data/blob/main/stacks

julialawrence (dismissed):

julialawrence (approved):
LGTM

PR #3790, opened 2024-03-20
Dc 156 add container entities

### LINKED ISSUE #156: update data product labels on front end to align with our new definition in DataHub
follows on from https://github.com/ministryofjustice/find-moj-data/issues/175

LavMatt:
## Issue

https://github.com/ministryofjustice/find-moj-data/issues/156

## What does this PR achieve?

This PR is doing quite a lot to the python datahub catalogue library, ultimately related to integrating another entity type - containers (or database), into `find-moj-data`:

- Adds new graphql query, `listContainerEntities.graphql` that lists all tables in a given container
- Adds ResultType.DATABASE enum to entities
- Adds the container entity to the existing search graphql query.
- Add new method to search.py, `SearchClient.list_database_tables` which uses `listContainerEntities.graphql`
- Update `SearchClient.search` to handle containers adding `_parse_container` method
- Expanded the properties contained within `entities.TableMetadata`
- Expanded the properties returned by `DataHubCatalogueClient.get_table_details` method and `getDatasetDetails.graphql` query to simply use in `find-moj-data` handling datasets of both data products and container (databases)
- Added `_parse_relations` function to `graphql_helpers`

I haven't made the search query return parent entities (relationships) of datasets where they belong to a container (currently returns data product parents only) - although this would be useful to populate the `Database Name` search result field in `find-moj-data` in the interest of time i didn't do this.


python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql_helpers.py
```diff
@@ -134,3 +134,17 @@ def parse_columns(entity: dict[str, Any]) -> list[dict[str, Any]]:
 
     # Sort primary keys first, then sort alphabetically
     return sorted(result, key=lambda c: (0 if c["isPrimaryKey"] else 1, c["name"]))
+
+
+def parse_relations(relations_dict) -> Tuple[int, list[dict]]:
```
MatMoore: I don't think this type is correct as it's a dict now.

But it might be simpler to just return the list here and ignore the total. I think the total would only be used if we need to page through a large set of relations?

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
```diff
@@ -176,19 +186,53 @@ def list_data_product_assets(
...
+            "count": count,
+        }
+
+        try:
+            response = self.graph.execute_graphql(
+                self.get_database_tables_query, variables
+            )
+        except GraphError as e:
+            raise Exception("Unable to execute listDatabaseEntities query") from e
+        page_results = self._get_data_collection_page_results(
```
MatMoore, line 215: whitespace suggestion

```suggestion

        page_results = self._get_data_collection_page_results(
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
```diff
@@ -223,19 +269,19 @@ def _parse_dataset(self, entity: dict[str, Any], matches) -> SearchResult:
...
+
+        relations_dict = parse_relations(entity.get("relationships", {}))
 
         metadata = {
             "owner": owner_name,
             "owner_email": owner_email,
-            "total_data_products": total_data_products,
-            "data_products": data_products,
+            "total_parents": relations_dict["total"],
+            "parents": relations_dict["entities"],
```
MatMoore: I think you are right we might have to change this if we end up using data products and containers together, but we might not need the distinction so keeping it simple makes sense for now

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
```diff
@@ -223,19 +269,19 @@ def _parse_dataset(self, entity: dict[str, Any], matches) -> SearchResult:
...
+            "entity_sub_type": (
+                entity.get("subTypes", {}).get("typeNames", ["Dataset"])
+                if entity.get("subTypes") is not None
+                else ["Dataset"]
```
MatMoore: Is it worth extracting this into a `parse_subtypes` or `parse_types_and_subtypes` function?

This is going to be very useful for labelling stuff consistently in the frontend

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
```diff
@@ -365,3 +411,43 @@ def get_glossary_terms(self, count: int = 1000) -> SearchResponse:
...
+            "owner": owner_name,
+            "owner_email": owner_email,
+            "entity_sub_type": (
+                entity.get("subTypes", {}).get("typeNames", ["Container"])
+                if entity.get("subTypes") is not None
+                else ["Container"]
+            ),
+        }
+
+        fqn = (
```
MatMoore: I think this logic is repeated in a few places, so could extract a common function here as well

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -107,13 +107,17 @@ class TableMetadata:
     column_details: list
     retention_period_in_days: int | None
     domain: str | None = None
-    parent_database_name: str | None = None
+    parent_entity_name: str | None = None
+    relationships: dict | None = None
```
MatMoore: Is `relationships` the right word here? I was thinking this is just `parents`? 

If we want this to be able to handle any kind of relationship I think there should be some way of inspecting the relationship type.

Maybe we could introduce another object for this, e.g.

```python

class RelationshipType(Enum):
  PARENT = auto()


class RelatedAsset:
  id: str
  name: str


class TableMetadata:
  relationships: dict[RelationshipType, list[RelatedAsset]]
```

`parent_entity_name` would then be redundant, and in the frontend we could iterate over any kind of relationship and render links to each one

MatMoore (dismissed):
This is looking good. I have a few suggestions but no major blockers

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
LavMatt: totally agree. Although `parent_entity_name` is also used for ingestion purposes, so i'll still keep that

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql_helpers.py
LavMatt: Yeah have fixed, maybe we could keep it in as if we use relations going the other direction we'd want it and not doing too much harm in there

LavMatt: or maybe it is complexity we don't need at the moment. Front end has no use for it so it's kind of redundant at the moment

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py
LavMatt: Have added this

MatMoore (approved):

PR #3884, opened 2024-03-27
Pydantic spike

MatMoore:
Our metadata model has evolved over time - we originally wrote it for the ingestion API but it has morphed into a way of ingesting metadata stored as JSON in https://github.com/ministryofjustice/data-catalogue-metadata into the DataHub catalogue.

So we're left with inconsistencies between read and write methods, and fields that don't align with how we are thinking about the catalogue. This branch explores what this could look like if we remodelled this from scratch, using pydantic as a way to serialise/deserialise the metadata and document what metadata fields are supposed to mean.

To do this I've combined the existing entities model with descriptions from https://github.com/ministryofjustice/data-platform-firebreak-python-api/blob/main/daap_api/models/api/metadata_api_models.py (firebreak project)

I've changed the table model to match how we are now using it (e.g. storing the source column types instead of mapping everything to a glue schema).

The goal is to arrive at an executable model that:
- maps cleanly to datahub's GraphQL API
- incorporates everything we are already representing in the frontend, (including custom properties we've introduced)
- is JSON serialisable
- uses consistent naming conventions
- is extendable

If we achieve this, then we can simplify the DatahubClient to use these models for reading and writing metadata without losing information.

## Removed metadata
- `retention_period` - I think this is too problematic as defined and we're not using it at the moment

## Metadata that is still unrefined
- How we represent platforms (currently part of the method name)
- Info about how to access the data (bearing in mind that if we catalogue athena tables, users may still access through s3)
- Summary data (we have a custom field for number of rows but datahub has its own way to represent this info)
- Data quality information (we haven't got to this yet)
- The actual governance model is more complicated than I've made it out to be, but I've dropped the older terminology in favour of DIP roles we are likely to surface in find-moj-data

## Databases
database

## Tables
table

## Charts
chart


python-libraries/data-platform-catalogue/data_platform_catalogue/entities_v2.py
```diff
@@ -0,0 +1,262 @@
...
+    )
+
+    # Needs refining
+    last_updated: datetime = Field(
+        description="When the metadata was last updated in the catalogue"
+    )
+    first_created: datetime = Field(description="When the data asset was first created")
+
+
+class Database(Asset):
```
LavMatt: Wondering for these child classes of Asset, if we could set the platform/entity type/sub-type here aswell. Might help moving towards having single methods to push and pull to/from datahub

python-libraries/data-platform-catalogue/data_platform_catalogue/entities_v2.py
```diff
@@ -0,0 +1,262 @@
...
+        default="",
+    )
+    source_dataset_name: str = Field(description="", default="")
+    s3_location: str | None = Field(
+        description="Location of the data in s3", default=None
+    )
+
+
+# Needs refining
+class DataSummary(BaseModel):
```
LavMatt, line 171: might be worth also adding metadata of when the data were last refreshed/updated - as in the actual data itself

python-libraries/data-platform-catalogue/data_platform_catalogue/entities_v2.py
```diff
@@ -0,0 +1,262 @@
...
+    )
+
+
+class Asset(BaseModel):
+    """
+    Any searchable data asset that is present in the metadata graph, which
+    may be related to other assets.
+    """
+
+    id: str | None = Field("Unique identifier for the asset")
```
LavMatt:
```suggestion
    id: str | None = Field("Unique identifier for the asset. Relates to Datahub's urn")

File: python-libraries/data-platform-catalogue/data_platform_catalogue/entities_v2.py [hunk 4]
```diff
@@ -0,0 +1,262 @@
...
+        default=DatabaseStatus.PROD,
+        description="Whether this database represents production data or not",
+    )
+
+
+class Table(Asset):
+    """
+    A table in a database or a tabular dataset
+    """
+
```
LavMatt, line 233: Do we also need to capture the actual name of a table as defined in database schema, i would imagine this might differ in a lot of cases to `display_name`. Probably same for `Database` too

LavMatt (commented):
Really like this Mat! Few comments/suggestions

python-libraries/data-platform-catalogue/data_platform_catalogue/entities_v2.py
MatMoore: Agreed!

python-libraries/data-platform-catalogue/data_platform_catalogue/entities_v2.py
MatMoore: Yeah - I think the way we talk about dates is a little confusing at the moment because we have dates the metadata was refreshed but we also have expectations about how frequently the dataset is updated that we want to communicate to the user.

I think this is tied to the data quality metrics - it's effectively a verifiable assertion about its timeliness

python-libraries/data-platform-catalogue/data_platform_catalogue/entities_v2.py
MatMoore: Yes, good shout. Maybe `name` / `display_name` / `fully_qualified_name` should be defined on the Asset class?

python-libraries/data-platform-catalogue/data_platform_catalogue/entities_v2.py
LavMatt: Think it's the more useful of the date information vs last updated in catalogue (from a consumer perspective), but not sure it's really data quality. E.g. You might have pipelines that are run daily from live operation systems but some might only be run monthly - but that schedule and the reality of the execution of the schedule are different. So as a user of the data you'd want to know if when go to use this data when are the records contained up to

python-libraries/data-platform-catalogue/data_platform_catalogue/entities_v2.py
LavMatt: Yeah i think that would make sense, would expect it of all assets

python-libraries/data-platform-catalogue/data_platform_catalogue/entities_v2.py
MatMoore: I've added this in as a new relationship type. Datahub's API seems to assume there is always one canonical platform, but I think the GMS is more flexible than that as when we ingest from DBT it gets associated with both a platform and a target platform.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities_v2.py
tom-webber: >that schedule and the reality of the execution of the schedule are different

Hopefully this will be [addressable in the future with Data Contracts](https://datahubproject.io/docs/next/managed-datahub/observe/data-contract/) / Metadata quality. We can ingest data [from dbt's `run_results`](https://datahubproject.io/docs/next/managed-datahub/observe/data-contract/#dbt-test) or [Great Expectations](https://datahubproject.io/docs/next/managed-datahub/observe/data-contract/#great-expectations) to let us know if the scheduled ingestion / transformation ran

python-libraries/data-platform-catalogue/data_platform_catalogue/entities_v2.py
```diff
@@ -0,0 +1,272 @@
...
+    )
+    access_information: AccessInformation = Field(
+        description="Metadata about how to access the data"
+    )
+    data_summary: DataSummary = Field(
+        description="Summary of data stored in this table"
+    )
+
+
+class Chart(Asset):
```
tom-webber: Suggest changing asset to entity, as it aligns more closely with the meaning, and [DataHub's](https://datahubproject.io/docs/metadata-modeling/metadata-model) [terminology](https://datahubproject.io/docs/metadata-modeling/metadata-model/#the-core-entities)
> For example, an instance of a Dataset or a CorpUser is an Entity
```suggestion
class Entity(BaseModel):
    """
    Any searchable data entity that is present in the metadata graph, which
    may be related to other entities.
    Examples include platforms, databases, tables, data products
    """

    id: str | None = Field("Unique identifier for the entity. Relates to Datahub's urn")
    display_name: str | None = Field("Display name of the entity")
    name: str = Field("Actual name of the entity in its source platform")
    fully_qualified_name: str | None = Field(
        "Fully qualified name of the entity in its source platform"
    )
    description: str = Field(
        description="Detailed description about what functional area this entity is representing, what purpose it has and business related information.",
    )
    relationships: dict[RelationshipType, list[EntityRef]] | None = Field(
        default=None,
        description="References to related entities in the metadata graph, such as platform or parent entities",
    )
    domain: DomainRef = Field(description="The domain this entity belongs to.")
    governance: Governance = Field(description="Information about governance")
    usage_restrictions: UsageRestrictions = Field(
        description="Limitations on how the data may be used and accessed"
    )
    tags: list[TagRef] = Field(
        default_factory=list,
        description="Additional tags to add.",
    )

    # Needs refining
    last_updated: datetime = Field(
        description="When the metadata was last updated in the catalogue"
    )
    first_created: datetime = Field(description="When the data entity was first created")


class Database(Entity):
    """
    For source system databases
    """

    status: DatabaseStatus = Field(
        default=DatabaseStatus.PROD,
        description="Whether this database represents production data or not",
    )


class Table(Entity):
    """
    A table in a database or a tabular dataset
    """

    column_details: list[Column] = Field(
        description="A list of objects which relate to columns in your data, each list item will contain, a name of the column, data type of the column and description of the column."
    )
    fully_qualified_name: str | None = Field(
        default=None,
        description="Fully qualified table name as it appears in the source platform, including the database",
    )
    access_information: AccessInformation = Field(
        description="Metadata about how to access the data"
    )
    data_summary: DataSummary = Field(
        description="Summary of data stored in this table"
    )


class Chart(Entity):
```

PR #3976, opened 2024-04-05
Remove tamsin from team as work has been completed

murad-ali-MoJ:
Take off Tamsin's name from the data-platform as the CaDT collaboration completed


michaeljcollinsuk (approved):
LGTM

PR #4106, opened 2024-04-17
🌐 Set BOLD egress KMS keys to multi region

jacobwoffenden:
This pull request:

- Sets `s3/mojap-data-production-bold-egress-*` to multi region, this is required because Airflow runs in `eu-west-1` but Analytical Platform new resources run in `eu-west-2`


michaeljcollinsuk (approved):
LGTM

PR #4115, opened 2024-04-18
Catalogue library refactor

murdo-moj:
https://github.com/ministryofjustice/find-moj-data/issues/265


python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql_helpers.py
```diff
@@ -2,24 +2,38 @@
...
+    OwnerRef,
+    RelationshipType,
+    TagRef,
+    UsageRestrictions,
+)
+
+
+def parse_owner(entity: dict[str, Any]) -> OwnerRef:
     """
     Parse ownership information, if it is set.
```
tom-webber, line 22:
```suggestion
    Parse ownership information, if it is set.
    If no owner information exists, return an OwnerRef populated with blank strings
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql_helpers.py
```diff
@@ -56,37 +70,55 @@ def parse_tags(entity: dict[str, Any]) -> list[str]:
     for tag in outer_tags.get("tags", []):
         properties = tag["tag"]["properties"]
         if properties:
-            tags.append(properties["name"])
+            tags.append(
+                TagRef(
+                    display_name=properties["name"],
+                    id=tag["tag"]["urn"],
```
tom-webber: ⛏️ using `.get` calls rather than key lookups
```suggestion
        properties = tag.get("tag", {}).get("properties", {})
        if properties:
            tags.append(
                TagRef(
                    display_name=properties.get("name", ""),
                    id=tag.get("tag", {}).get("urn", ""),
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -142,7 +143,7 @@ def check_entity_exists_by_urn(self, urn: str | None):
 
         return exists
 
-    def upsert_database_service(self, platform: str = "glue", *args, **kwargs) -> str:
+    def upsert_database_service(self, *args, platform: str = "glue", **kwargs) -> str:
```
tom-webber:
```suggestion
    def upsert_data_platform(self, *args, platform: str = "glue", **kwargs) -> str:
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -151,32 +152,16 @@ def upsert_database_service(self, platform: str = "glue", *args, **kwargs) -> st
...
-        raise NotImplementedError
-
     def create_domain(
-        self, domain: str, description: str = "", parentDomain: str | None = None
+        self, domain: str, description: str = "", parent_domain: str | None = None
     ) -> str:
         """Create a Domain, a logical collection of Data assets (Data Products).
 
         Args:
             metadata (DataProductMetadata): _description_
```
tom-webber:
```suggestion
        """Create a Domain, a logical collection of Data assets

        Args:
            domain (str): name of the new Domain
            description (str, optional): Description of the new Domain. Defaults to "".
            parent_domain (str | None, optional): Declared child relationship to existing Domains.
              Defaults to None.

        Returns:
            str: urn of the created Domain
        """
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/exceptions.py
```diff
@@ -0,0 +1,38 @@
+class CatalogueError(Exception):
+    """
+    Base class for all errors.
```
tom-webber:
```suggestion
    Base class for catalogue-related errors.
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/exceptions.py
```diff
@@ -0,0 +1,38 @@
...
+class ReferencedEntityMissing(CatalogueError):
+    """
+    A referenced entity (such as a user or tag) does not yet exist when
+    attempting to create a new metadata resource in the catalogue.
+    """
+
+
+class InvalidDomain(CatalogueError):
+    """
+    Exception thrown when a domain does not exist
```
tom-webber:
```suggestion
    Domain does not exist
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/exceptions.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+
+class MissingDatabaseMetadata(Exception):
+    """
+    Exception thrown when a entity is attempted to be ingested without
+    a given metadata specification
+    """
```
tom-webber: Exception name is sufficient
```suggestion
class MissingDatabaseMetadata(Exception):
    pass
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/exceptions.py
```diff
@@ -0,0 +1,38 @@
...
+    Exception thrown when an entity that does not exist is trying to be
+    accessed in DataHub
+    """
+
+
+class AspectDoesNotExist(CatalogueError):
+    """
+    Exception thrown when an Aspect that does not exist is trying to be
+    accessed in DataHub
+    """
```
tom-webber: Exception name is sufficient
```suggestion
class AspectDoesNotExist(CatalogueError):
    pass
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/exceptions.py
```diff
@@ -0,0 +1,38 @@
...
+    Exception thrown when a entity is attempted to be ingested without
+    a given metadata specification
+    """
+
+
+class EntityDoesNotExist(CatalogueError):
+    """
+    Exception thrown when an entity that does not exist is trying to be
+    accessed in DataHub
+    """
```
tom-webber: Exception name is sufficient
```suggestion
    pass
```

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
-    tags: list[str] = field(default_factory=list)
+class DatabaseStatus(Enum):
+    PROD = "PROD"
+    PREPROD = "PREPROD"
+    DEV = "DEV"
+
 
+class RelationshipType(Enum):
+    PARENT = "PARENT"
+    PLATFORM = "PLATFORM"
```
tom-webber: Why have these enum values changed from [Enum.auto()](https://docs.python.org/3/library/enum.html#enum.auto)?

murdo-moj: when `model_dump_json` is called on a pydantic model eg to encode classes to pass to datahub, enum classes are converted to their string representations. With auto(), this is an integer value which can't be read as it's actual value.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
-class DataProductStatus(Enum):
-    DRAFT = auto()
-    PUBLISHED = auto()
-    RETIRED = auto()
+class EntityRef(BaseModel):
+    """
+    A reference to another entity in the metadata graph.
+    """
 
+    id: str = Field(description="The identifier of the entity being linked to.")
```
tom-webber: 🔍 is the `id` always a urn?
```suggestion
    id: str = Field(description="The urn of the entity being linked to.")
```

MatMoore:
Since we all worked on this together and this is a big rewrite, I'm going to treat the review as a checklist and go through it bit by bit. We can improve stuff mid review and once we've got everything checked off we can merge.

## Review checklist

### Pydantic models
The models are
- [ ] no more complicated than needed
- [ ] intuitive (reusing existing terminology where possible, and using sensible names for anything custom to MOJ)
- [ ] documented
- [x] debuggable

### Overall API
The API
- [x] Is complete (with some exceptions: upsert method missing for chart, get method missing for database, list methods only for tables)
- [ ] Is self-consistent
- [x] Is at the right level of abstraction for use in Find MOJ data / registration API
- [x] Throws appropriate and consistent exceptions

### upsert_chart
N/A not implemented yet

### upsert_table
- [x] Test coverage
- [x] No obvious bugs or unhandled edge cases
- [x] Code is readable
- [x] Code uses Datahub API efficiently

### upsert_database
- [x] Test coverage
- [x] No obvious bugs or unhandled edge cases
- [x] Code is readable
- [x] Code uses Datahub API efficiently

### create_domain
- [ ] Rename to upsert_domain for consistency
- [x] Test coverage
- [x] No obvious bugs or unhandled edge cases
- [x] Code is readable
- [x] Code uses Datahub API efficiently

### list_database_tables
- [x] Test coverage
- [x] No obvious bugs or unhandled edge cases
- [x] Code is readable
- [x] Code uses Datahub API efficiently

### get_table_details
- [x] Test coverage
- [x] No obvious bugs or unhandled edge cases
- [x] Code is readable
- [x] Code uses Datahub API efficiently

### search
- [x] Good test coverage
- [ ] No obvious bugs or unhandled edge cases
- [ ] Code is readable
- [ ] Code uses Datahub API efficiently

### search_facets
- [x] Good test coverage
- [ ] No obvious bugs or unhandled edge cases
- [ ] Code is readable
- [ ] Code uses Datahub API efficiently

### get_glossary_terms
- [ ] Good test coverage
- [ ] No obvious bugs or unhandled edge cases
- [ ] Code is readable
- [ ] Code uses Datahub API efficiently

### Misc
- [ ] Nothing else missing from the changelog

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
tom-webber: If we're ~~above python 3.10~~ using python >=3.11, we can swap `(Enum)` to `(StrEnum)`, and we should get compatible behaviour for both Enum and Pydantic.
>Note Using [auto](https://docs.python.org/3/library/enum.html#enum.auto) with [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum) results in the lower-cased member name as the value.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
 DATAHUB_DATE_FORMAT = "%Y%m%d"
 
 
-@dataclass
-class CatalogueMetadata:
-    name: str
-    description: str
-    owner: str
-    tags: list[str] = field(default_factory=list)
+class DatabaseStatus(Enum):
```
MatMoore: Question: **Does this have a clear enough use case at this point?**

- If yes, we should add a clear docstring to communicate this.
- If no, we can drop it for now, and add it back in when we need it.

Some concerns with this concept, which are making me think we should drop it:

1. If we have both non-production and production copies of a database visible in the catalogue at the same time, this will confuse users unless we present in a special way
2. I don't understand what this actually represents and how this relates to the [fabric part of the URN](https://datahubproject.io/docs/what/urn/)?
  - Does a change of this status mean that a new entity has been created in a different environment OR does it mean that we are asserting something different about an already catalogued entity
  - Does the status reflect the source data (i.e. everything non-prod is likely to made up and comes with zero data quality assertions) OR does the status reflect level of support for the database (i.e. non-prod may contain high quality data but should not be used for anything business critical)
  - Since we expose URNs in the URLs, if this *is* a separate concept from fabric, then there is potential for users to confuse the two
3. The 3 options PROD/PREPROD/DEV is quite simple and engineering focused.
  - Is this classification understandable to a wide range of users?
  - Is this classification meaningful across the whole MOJ? Or would people struggle to classify as PREPROD vs DEV for example?

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
-    description: str
-    external_url: str
-    tags: list[str] = field(default_factory=list)
+class Column(BaseModel):
+    """
+    A column definition in a table
+    """
+
+    name: str = Field(
+        pattern=r"^[\w.]+$",
```
tom-webber, line 57: The dot matches any character, so should presumably be removed.
```suggestion
        pattern=r"^[\w]+$",
```

tom-webber, line 42: Do we accept numbers in column names? I feel like the answer is yes, but I'd like to be explicit.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
MatMoore: If we drop support for python 3.10, we could maybe use StrEnum with Auto https://docs.python.org/3/library/enum.html#enum.StrEnum

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+        description="The name of a column as it appears in the table.",
+        json_schema_extra={"pattern": r"^[\w.]+$"},
```
tom-webber: The dot matches any character, so should presumably be removed.
```suggestion
        pattern=r"^[\w]+$",
        description="The name of a column as it appears in the table.",
        json_schema_extra={"pattern": r"^[\w]+$"},
```

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+
 
+class RelationshipType(Enum):
+    PARENT = "PARENT"
+    PLATFORM = "PLATFORM"
 
-@dataclass
-class DataLocation:
+
+# Needs refining
```
MatMoore: Question: Do we use this classification anywhere at the moment?

If no, let's drop this entirely. 

If yes, we need to add in the missing values, possibly rename it.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
MatMoore: I believe so. The only reason we're using `id` at all is because originally, the models were independent of catalogue.

We could do a search and replace or id -> urn if nobody objects.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+    name: str = Field(
+        pattern=r"^[\w.]+$",
+        description="The name of a column as it appears in the table.",
+        json_schema_extra={"pattern": r"^[\w.]+$"},
+    )
+    display_name: str = Field(description="A user-friendly version of the name")
+    type: str = Field(
+        description="The data type of the column as it appears in the table",
+    )
+    description: str = Field(description="A description of the column")
```
tom-webber, line 49: We should come up with a way to clarify for all `description` properties whether we're referencing the `source` or `editable` schemaMetadataProperty, (see [python implementation for column description](https://datahubproject.io/docs/api/tutorials/descriptions/#add-description-on-dataset))

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+
+class OwnerRef(BaseModel):
+    """
+    A reference to a named individual that performs some kind of governance
+    """
+
+    display_name: str = Field(
+        description="The full name of the user as it should be displayed"
+    )
+    email: str = Field("Contact email for the user")
```
tom-webber, line 67:
```suggestion
    email: str = Field("Contact email for the user (@justice.gov.uk address)")
```

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+    id: str = Field("Unique identifier for the user")
+
+
+# Needs refining
+class Governance(BaseModel):
+    """
+    Governance model for an entity or domain
+    """
+
+    data_owner: OwnerRef = Field(description="")
```
MatMoore: Should give this descriptions

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+
+# Needs refining
+class UsageRestrictions(BaseModel):
+    """
+    Metadata about how entities may be used.
+    """
+
+    status: DatabaseStatus | None = Field(
+        description="this is an enum representing the status of this version of the Database. "
+        "Allowed values are: [draft|published|retired]. "
```
tom-webber: These current values don't match with the DatabaseStatus enum.
```suggestion
        "Allowed values are: [dev|preprod|prod]. "
```

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
LavMatt: Think i agree and this isn't adding much other than potential confusion. Is probably a bit of leftover from the previous iteration's schema where we were concerned re. the status of a piece of data that had been uploaded, and [this description](https://github.com/ministryofjustice/data-platform/blob/50a46dab56ecabc5e1f1ccae6c02e546fb6efbd7/python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py#L132) kind of backs that up (description not aligning with enum values). I'm not sure why anyone would need to catalogue data or a database that was draft status.

Guess it could play somehow into gated registration to the catalogue but that wouldn't be how it's currently being used and not database specific

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
MatMoore: I don't think we should be enforcing any pattern on column names at all in order to register to the catalogue. This name should be what the name actually is.

If we want to enforce a pattern we can do it on the `display_name` field.

Here is an example with loads of numbers https://datahub-catalogue-dev.apps.live.cloud-platform.service.justice.gov.uk/dataset/urn:li:dataset:(urn:li:dataPlatform:dbt,create_a_derived_table.awsdatacatalog.common_lookup.ns_postcode_lookup_latest_2011census,PROD)/Schema?is_lineage_mode=false&schemaFilter=

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
MatMoore: @tom-webber I don't think so, I checked and it is interpreted as literal inside a character class
Screenshot 2024-04-29 at 11 42 47

Datahub lets you pass in dotted field names so that must be where it's come from. It's a way of documenting nested structures like JSON fields in a database.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+    display_name: str = Field(
+        description="Human friendly tag name", json_schema_extra={"example": "PII"}
+    )
+    id: str = Field(
+        description="The identifier of the tag",
+        json_schema_extra={"example": "urn:li:tag:PII"},
+    )
+
+
+# Needs refining
```
MatMoore: I think we can drop these comments if we're not going to do this now
```suggestion
```

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+        description="Bool for if a data privacy impact assessment (dpia) is required to access this database",
+        default=None,
+    )
+    dpia_location: str | None = Field(description="", default=None)
+    data_sensitivity_level: SecurityClassification | None = Field(
+        description="", default=SecurityClassification.OFFICIAL
+    )
+
+
+# Needs refining
```
MatMoore:
```suggestion
```

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+class Governance(BaseModel):
+    """
+    Governance model for an entity or domain
+    """
+
+    data_owner: OwnerRef = Field(description="")
+    data_stewards: list[OwnerRef] = Field(description="")
+
+
+class DomainRef(BaseModel):
```
MatMoore, line 84: Question: do you think it's worth dropping `DomainRef` and `TagRef` and just using `EntityRef`

`OwnerRef` I deliberately made separate as there was some different fields, but in this case I'm not sure it's adding anything other than a bit of type safety.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+        "Allowed values are: [draft|published|retired]. "
+        "This is a metadata that communicates the overall status of the Database but is not reflected"
+        " to the actual deployment status.",
+        default=None,
+    )
+    dpia_required: bool | str | None = Field(
+        description="Bool for if a data privacy impact assessment (dpia) is required to access this database",
+        default=None,
+    )
+    dpia_location: str | None = Field(description="", default=None)
```
MatMoore: missing description

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+        description="User-friendly description of where the data can be accessed",
+        default="",
+    )
+    source_dataset_name: str = Field(description="", default="")
+    s3_location: str | None = Field(
+        description="Location of the data in s3", default=None
+    )
+
+
+# Needs refining
```
MatMoore:
```suggestion
```

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+
+class Entity(BaseModel):
+    """
+    Any searchable data entity that is present in the metadata graph, which
+    may be related to other entities.
+    Examples include platforms, databases, tables
+    """
+
+    id: str | None = Field("Unique identifier for the entity. Relates to Datahub's urn")
+    display_name: str | None = Field("Display name of the entity")
```
MatMoore, line 180: TODO: double check that we are using these 3 names consistently

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+        default={},
+        description="References to related entities in the metadata graph, such as platform or parent entities",
+    )
+    domain: DomainRef = Field(description="The domain this entity belongs to.")
+    governance: Governance = Field(description="Information about governance")
+    tags: list[TagRef] = Field(
+        default_factory=list,
+        description="Additional tags to add.",
+    )
+    # Needs refining
```
MatMoore:
```suggestion
```

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
LavMatt: Just a custom property. There was some talk of adding in subsets of official (which would be official sensitive too) see https://security-guidance.service.justice.gov.uk/government-classification-scheme/#government-classification-scheme

but this wasn't ever really agreed and not sure if it's useful content

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
```diff
@@ -1,200 +1,265 @@
...
+    last_updated: Optional[datetime] = Field(
```
MatMoore: I suggest we adopt Datahub's conventions for timestamps to avoid confusion

- `created`: A timestamp documenting when the asset was created in the source Data Platform (not on DataHub)
- `last_modified` instead of `last_updated`: A timestamp documenting when the asset was last modified in the source Data Platform (not on DataHub)

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
MatMoore: This is going to be important really soon as we develop the manual registration. Shall we create a separate ticket for this?

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql_helpers.py
```diff
@@ -2,24 +2,40 @@
 from datetime import datetime, timezone
 from typing import Any, Tuple
```
MatMoore, line 3: I think we agreed that if we changed catalogue we would create a different library rather than reuse this one? Is that right? If so we can drop the parent package of this module and just have `data_platform_catalogue.client`

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql_helpers.py
```diff
@@ -2,24 +2,40 @@
...
+    TagRef,
+    UsageRestrictions,
+)
+
+
+def parse_owner(entity: dict[str, Any]) -> OwnerRef:
     """
     Parse ownership information, if it is set.
+    If no owner information exists, return an OwnerRef populated with blank strings
+    If no owner information exists, return an OwnerRef populated with blank strings
```
MatMoore:
```suggestion
```

this line is duplicated

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/graphql_helpers.py
```diff
@@ -2,24 +2,40 @@
...
     """
     ownership = entity.get("ownership") or {}
     owners = [i["owner"] for i in ownership.get("owners", [])]
     if owners:
         properties = owners[0].get("properties") or {}
-        owner_email = properties.get("email", "")
-        owner_name = properties.get("fullName", properties.get("displayName", ""))
+        owner_details = OwnerRef(
+            display_name=properties.get("fullName", properties.get("displayName", "")),
```
MatMoore: should this be the other way round? Use displayName if set, otherwise fullName?

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
MatMoore: Yeah, that makes sense. I would feel better revisiting this once we are actually implementing gated registration - we are not there yet

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -1,23 +1,62 @@
...
+    SearchClient,
+)
+from data_platform_catalogue.client.exceptions import (  # pylint: disable=E0611
+    AspectDoesNotExist,
+    CatalogueError,
+    EntityDoesNotExist,
+    InvalidDomain,
+    ReferencedEntityMissing,
+)
+from data_platform_catalogue.entities import (  # pylint: disable=E0611
```
MatMoore: I don't think these pylint directives should be here

E0611 means the name can't be imported which would be a bug or environment issue https://pycodequ.al/docs/pylint-messages/e0611-no-name-in-module.html

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -142,7 +143,7 @@ def check_entity_exists_by_urn(self, urn: str | None):
 
         return exists
 
-    def upsert_database_service(self, platform: str = "glue", *args, **kwargs) -> str:
+    def upsert_data_platform(self, *args, platform: str = "glue", **kwargs) -> str:
```
MatMoore: TODO: drop this

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -151,32 +152,22 @@ def upsert_database_service(self, platform: str = "glue", *args, **kwargs) -> st
...
-        Define a database. Not implemented for DataHub, which uses Data Platforms + Datasets only.
-        """
-        raise NotImplementedError
-
     def create_domain(
-        self, domain: str, description: str = "", parentDomain: str | None = None
+        self, domain: str, description: str = "", parent_domain: str | None = None
     ) -> str:
-        """Create a Domain, a logical collection of Data assets (Data Products).
+        """Create a Domain, a logical collection of Data assets
```
MatMoore:
```suggestion
        """Create a Domain, a logical collection of entities
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -191,105 +182,159 @@ def create_domain(
...
-            entityUrn=database_urn,
-            aspect=DataPlatformInstance(
-                platform="urn:li:dataPlatform:athena",
-            ),
-        )
-        self.graph.emit(metadata_event)
-        logger.info(f"Platform updated for Database {name} ")
+    # to expand on and replace `list_database_tables` will need new graphql query i expect
+    # but will be more equivelent to the get chart and table details methods.
+    def get_database_details(self, urn: str):
```
MatMoore:
```suggestion
    def get_database_details(self, urn: str) -> Database:
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -191,105 +182,159 @@ def create_domain(
...
+        pass
 
-        # container type update
-        metadata_event = MetadataChangeProposalWrapper(
-            entityUrn=database_urn,
-            aspect=SubTypesClass(typeNames=[DatasetContainerSubTypes.DATABASE]),
-        )
-        self.graph.emit(metadata_event)
-        logger.info(f"Type updated for Database {name} ")
+    def upsert_table(self, metadata: Table):
```
MatMoore: todo - probably worth renaming metadata -> table here. Just adds confusion

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -191,105 +182,159 @@ def create_domain(
...
-        self,
-        metadata: TableMetadata,
-        database_metadata: DatabaseMetadata | None = None,
-    ) -> str:
-        if not metadata.parent_entity_name:
-            raise ValueError("parent_entity_name needs to be set in TableMetadata")
-
-        fully_qualified_name = f"{metadata.parent_entity_name}.{metadata.name}"
+        parent_name = metadata.relationships[RelationshipType.PARENT][0].display_name
+        fully_qualified_name = f"{parent_name}.{metadata.name}"
```
MatMoore: TODO: I think it's probably worth extracting out anything that constructs names from other names - need to make sure this is consistent

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -397,321 +442,108 @@ def upsert_athena_table(
 
         return dataset_urn
 
-    def upsert_data_product(self, metadata: DataProductMetadata):
+    def upsert_chart(self):
```
MatMoore:
```suggestion
    def upsert_chart(self, chart: Chart):
        """Define a chart. Must belong to a domain."""
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -397,321 +442,108 @@ def upsert_athena_table(
...
+        pass
```
MatMoore:
```suggestion
        raise NotImplementedError
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
MatMoore:
```suggestion
    def upsert_table(self, metadata: Table):
        """Define a table. Must belong to a domain."""
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -397,321 +442,108 @@ def upsert_athena_table(
...
 
-        data_product_properties = DataProductPropertiesClass(
-            customProperties={key: str(val) for key, val in metadata_dict.items()},
+        database_properties = ContainerPropertiesClass(
+            customProperties=self._get_custom_property_key_value_pairs(
+                metadata.custom_properties
+            ),
+            # customProperties=json.loads(
+            #     metadata.custom_properties.json(), parse_int=str
+            # ),
```
MatMoore:
```suggestion
```

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -397,321 +442,108 @@ def upsert_athena_table(
...
-        """
-        return self.search_client.search(
-            query=query,
-            count=count,
-            page=page,
-            result_types=result_types,
-            filters=filters,
-            sort=sort,
-        )
+    def list_database_tables(self, urn: str, count: int) -> SearchResponse:
```
MatMoore: Seems a bit weird that a `list` method returns a `SearchResponse` rather than a `list[Table]`. But I think we can leave this for now.

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
LavMatt: Wouldn't we always want to be referencing the editable property via this library? 

We still need to explore this and get a better understanding of how it works in practice, but the way i'd seen it working was that built-in native ingestions from datahub would use the source and anything we'd developed would use the editable. We also don't use editable in this library so that's some potential further work

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
LavMatt: might there be cases when an owner's email is not `@justice.gov.uk`?

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
LavMatt: Wondering if there might be additional properties we add to domain or tags, suppose both might benefit with having descriptions? And i think Datahub does allow descriptions for tags and domains

python-libraries/data-platform-catalogue/tests/client/datahub/test_datahub_client.py
```diff
@@ -481,40 +493,36 @@ def test_create_athena_database_and_table(
...
 
-        fqn_t_out = (
-            "urn:li:dataset:(urn:li:dataPlatform:athena,my_database.my_table,PROD)"
-        )
+        fqn_t_out = "urn:li:dataset:(urn:li:dataPlatform:athena,database.Dataset,PROD)"
         assert fqn_t == fqn_t_out
 
         output_file = Path(tmp_path / "datahub_create_athena_table.json")
         base_mock_graph.sink_to_file(output_file)
         check_snapshot("datahub_create_athena_table.json", output_file)
```
MatMoore: We should rename these snapshots since we renamed the test

python-libraries/data-platform-catalogue/tests/client/datahub/test_datahub_client.py
```diff
@@ -481,40 +493,36 @@ def test_create_athena_database_and_table(
...
 
-    def test_create_athena_table_with_metadata(
+    def test_create_table_with_metadata(
```
MatMoore: This name isn't reflecting what the test does anymore

I think this one is testing the case where you create a table independently of a database, whereas the one above tests creating a database and table that relate to each other

python-libraries/data-platform-catalogue/tests/snapshots/datahub_create_athena_table.json
```diff
@@ -7,18 +20,14 @@
...
                 "dpia_required": "False",
                 "dpia_location": "None",
-                "last_updated": "2020-05-17 00:00:00",
-                "creation_date": "2020-05-17 00:00:00",
+                "data_sensitivity_level": "OFFICIAL",
+                "where_to_access_dataset": "analytical_platform",
+                "source_dataset_name": "",
                 "s3_location": "s3://databucket/",
-                "status": "PROD"
+                "row_count": "None"
```
MatMoore: This seems like a bug - I would expect the row_count property to default to an empty string if not defined

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -191,105 +182,159 @@ def create_domain(
...
+    ) -> SearchFacets:
+        """
+        Returns facets that can be used to filter the search results.
+        """
+        return self.search_client.search_facets(
+            query=query, result_types=result_types, filters=filters
+        )
 
-        database_urn = "urn:li:container:" + "".join(name.split())
+    def get_glossary_terms(self, count: int = 1000) -> SearchResponse:
```
MatMoore: TODO: add tests

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
LavMatt: this should help inform https://github.com/ministryofjustice/find-moj-data/issues/287

python-libraries/data-platform-catalogue/data_platform_catalogue/entities.py
MatMoore: It's not enforced anywhere at the moment, except that only certain domains will be able to log in through EntraID. I don't think it matters too much as it's not something that users will be directly entering into a form

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -346,20 +391,13 @@ def upsert_athena_table(
             self.graph.emit(event)
         # jscpd:ignore-end
 
-        database_urn = "urn:li:container:" + "".join(
-            metadata.parent_entity_name.split()
-        )
+        database_urn = metadata.relationships[RelationshipType.PARENT][0].id
```
MatMoore: This is assuming that the relationships list has at least one element - think there's an unhandled IndexError here

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -300,18 +345,18 @@ def upsert_athena_table(
...
+            version=1,
             hash="",
             platformSchema=OtherSchemaClass(rawSchema=""),
             fields=[
                 SchemaFieldClass(
-                    fieldPath=f"{column['name']}",
+                    fieldPath=f"{column.name}",
                     type=SchemaFieldDataTypeClass(
-                        type=DATAHUB_DATA_TYPE_MAPPING[column["type"]]
+                        type=DATAHUB_DATA_TYPE_MAPPING[column.type]  # type: ignore
```
MatMoore: Note for later: we need to handle the case where `column.type` is missing from the dictionary if we expand the usage of this library

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -397,321 +442,108 @@ def upsert_athena_table(
...
+        custom_properties_unnested = self.unpack(custom_properties_dict)
+        custom_properties_unnested_all_string_values = {
+            key: str(value)
+            # if value is not None else value
+            for key, value in custom_properties_unnested.items()
+        }
+
+        return custom_properties_unnested_all_string_values
+
+    def unpack(self, d, custom_properties=None):
```
MatMoore: found this name a bit confusing

```suggestion
        custom_properties_unnested = self._flatten_dict(custom_properties_dict)
        custom_properties_unnested_all_string_values = {
            key: str(value)
            # if value is not None else value
            for key, value in custom_properties_unnested.items()
        }

        return custom_properties_unnested_all_string_values

    def _flatten_dict(self, d, custom_properties=None):
```

Might also be worth moving this to a method of `CustomProperties`

python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/datahub_client.py
```diff
@@ -191,105 +182,159 @@ def create_domain(
...
+            name = properties.get("name", response.get("name"))
+            created, modified = parse_created_and_modified(properties)
 
-        # domains are to be controlled and limited so we dont want to create one
-        # when it doesn't exist
-        domain_exists = self.check_entity_exists_by_urn(domain_urn)
-        if not domain_exists:
-            raise InvalidDomain(
-                f"{domain} does not exist in datahub - please align data to an existing domain"
+            # A dataset can't have both a container and data product parent, but if we did
```
MatMoore:
```suggestion
            # A dataset can't have multiple parents, but if we did
```

MatMoore (approved):

mitchdawson1982 (approved):
LGTM

jacobwoffenden (approved):

PR #4242, opened 2024-04-30
:dependabot: Add Dependabot configuration generator workflow

jacobwoffenden:
This pull request:

- Adds a new workflow that regenerates Dependabot configuration on every commit and pushes it back if changes, this means any new components should be picked up automatically if the guidance is followed
- It uses @moj-data-platform-robot's token so workflows can kick off again
- Removes step from docs (thanks @julialawrence)
- BONUS: Adds `concurrency` to Terraform workflow


julialawrence (commented):
I'm unclear. Does that mean we no longer need to run the config generation locally? And if not, does that mean guidance needs an update?

jacobwoffenden:
> I'm unclear. Does that mean we no longer need to run the config generation locally? And if not, does that mean guidance needs an update?

Yes and yes

julialawrence (approved):

PR #4243, opened 2024-04-30
😭 Remove noisy linters

jacobwoffenden:
This pull request:

- Removes Checkov, it loads everything even when `VALIDATE_ALL_CODEBASE: false`
- Removes JSCPD, since upgrading to Super-Linter v6, it processes everything, which makes sense given its looking for code duplication


Gary-H9 (approved):
LGTM

PR #4244, opened 2024-04-30
Consistently parse names from datahub

### LINKED ISSUE #294: 🐛 fully_qualified_name is returning none where name is populated
For the catalogue library, fully qualified name should always be populated if name is populated. There are some cases where this is not the case.

MatMoore:
Should resolve https://github.com/ministryofjustice/find-moj-data/issues/294 and https://github.com/ministryofjustice/find-moj-data/issues/295

Previously, we were parsing names inconsistently, and there were edge cases where we would get unexpected None values or KeyErrors.

Datahub has 4 possible names that might be set

- name: Unique guid for dataset No longer to be used as the Dataset display name. Use properties.name instead
- properties.name
- properties.displayName
- properties.qualifiedName

These are used inconsistently by different ingestion sources. If we have displayName / qualifiedName it's unambiguous, but the top level name can be either.

We expose 3 values:
- `displayName` and `qualifiedName` align with the corresponding datahub names, and fall back to the generic name if not set.
- `name` is probably redundant now, but I've kept it in to avoid breaking things further.

### DIFF
--- a/.github/linters/.python-lint
+++ b/.github/linters/.python-lint
@@ -1,0 +1,640 @@
+[MAIN]
+
+# Analyse import fallback blocks. This can be used to support both Python 2 and
+# 3 compatible code, which means that the block might have code that exists
+# only in one or another interpreter, leading to false positives when analysed.
+analyse-fallback-blocks=no
+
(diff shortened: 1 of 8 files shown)


Gary-H9 (approved):
LGTM

PR #4256, opened 2024-05-01
Coerce descriptions and custom properties to string/empty string

### LINKED ISSUE #296: 🐛  Some column descriptions are none in DataHub, which Pydantic rejects
https://datahub-catalogue-dev.apps.live.cloud-platform.service.justice.gov.uk/dataset/urn:li:dataset:(urn:li:dataPlatform:dbt,create_a_derived_table.awsdatacatalog.prisons_modelled_dim.fct_releases,PROD)/Schema?is_lineage_mode=false&schemaFilter=

the column descriptions are set to None

so when Pydantic tries to construct a Column class in `graphQL_helpers.py` L177, description is set to None. Description needs to be a string according to `entities.py`

suggestion 
```description = field.get("description") if field.get("description") else ""```

MatMoore:
The way we're handling descriptions and custom properties allows us to store the literal string "None" in datahub if values are not set. Instead, we should store empty string.

I also noticed that the type of dpia_required is `str | bool | None` which is very confusing - I've simplified to `bool | None` and added an extra line to convert from str to bool explicitly. This will avoid us treating "False" as True in the frontend.

Should resolve https://github.com/ministryofjustice/find-moj-data/issues/296


hjribeiro-moj (approved):
LGTM

jacobwoffenden (approved):

Issue #4120, opened 2024-04-19
✨ Short term accelerated compute instance


bcrawford-moj:
### Describe the feature request.

Short term provisioning of an accelerated compute instance. 

### Describe the context.

In the BOLD programme we are producing a publication on the number of prisoners with children. We have developed a methodology involving LLMs which checks whether prison case notes imply the prisoner has a child. The output will be an Official Statistics in Development report due for publication around end of May. 

We used the AP to run the LLM over the case notes but obviously it is quite slow (takes about a week to churn through them). The QA process has highlighted some changes we need to make. 

### Value / Purpose

This will allow us to meet our publication deadline.

We believe this will be the first time LLMs have been used in producing official statistics (and indeed one of our models was fine-tuned using generated labeled data, so we think it will also be the first time genAI has been used). 

We are happy to have associated costs journaled to the BOLD programme.

### User Types

Data scientists

### Proposed solution:

- [x] Get resource requirements from requestor
- [x] Select instance size
- [x] Add GPU-enabled node-group
- [x] Amend Control Panel code to allow specifying GPU requirements
- [x] Create a vscode limited release with GPU 
- [x] Docs updated
- [x] Tests Green
- [x] Follow-up stories raised
- [x] Code reviewed

AntFMoJ:
@bcrawford-moj do you have a sense of resource requirements at this time.

e.g. do you have an estimate on the file size of the input data being processed by the LLM.

yznlp:
Hi @AntFMoJ I'm the one working with the model. 

Not too familiar with how the infrastructure/resource provision works, but I imagine we just need something similar to what's available on the AP in terms of CPU and RAM, but with a GPU enabled with 16GB or even 8GB VRAM. 

Reasoning: the entire model runs on the AP by splitting the data into smaller chunks. For the LLM/transformer component it's a fairly small model (44M parameters) which should fit comfortably on 8GB of VRAM. The inputs are only 128 tokens (~words) long, and we will be able to scale the model to the amount of VRAM available. 

Thanks for your help and please let me know if something doesn't make sense :)

BrianEllwood:
Based on user resource requirements we would recommend a p3.2xlarge instance.

BrianEllwood:
Draft [PR](https://github.com/ministryofjustice/data-platform/pull/4253) created to build GPU node group

BrianEllwood:
Hi, @yznlp @bcrawford-moj 

[...]

PR #4371, opened 2024-05-21
Create and configure the Nvidia daemon set

BrianEllwood:
### Pull Request Objective

To put into code the changes needed to setup nvidia daemon set

The purpose is to creare the nvidia-device-plugin via yaml rather than installing via helm which proved not to work.

https://github.com/ministryofjustice/analytical-platform/issues/4120

**Note**: Static analysis has been overridden as the issues flagged in the development cluster are historic and not arising from the changes made in this PR.


jacobwoffenden (approved):
LGTM

Issue #4389, opened 2024-05-22
🚀 Manage Airflow EKS Launch Templates in Terraform


Gary-H9:
### User Story

As a… AP engineer
I need/want/expect to… the Airflow EKS Clusters to be properly managed in Terraform
As this is our expected practice and comes with all of the benefits of managing something in IaC

### Value / Purpose

We are regularly receiving support requests related to Airflow. In reconciling this anomalous state (in relation to our principles) it is hoped that we can better manage Airflow in future and provide a better user experience for our users. 

### Proposal

Manage Launch Templates 

### Additional Information
Re-importing the `aws_eks_node_group` resource sadly didn't have the desired effect.

### Definition of Done
- [x] Manage Launch Templates in Terraform for `airflow-dev` 
- [x] Manage Launch Templates in Terraform for `airflow-prod` 
- [x] Create comms with users informing them of any downtime required in Prod (if this is deemed necessary)

Gary-H9:
To ensure no downtime (when this is done in production) I am going to try the following approach in development: 
- create new node groups with launch templates aligned properly
- migrate workloads
- drain the old node group.
- update any references or configurations to point to the new node group.
- delete the old node group once the migration is complete.
- rename the 'new' node groups in Terraform so they don't include anything ugly in their names like `new` or `-2`

Gary-H9:
Created PRs to add 'new' nodes to [dev](https://github.com/ministryofjustice/analytical-platform/pull/4391) and [prod](https://github.com/ministryofjustice/analytical-platform/pull/4392).

Gary-H9:
Merged the above PRs, short these brought the `main` branch / state into agreement with changes that were made outside of CI/CD.

The process was:  
- Add 'new' nodes
- Move any running jobs from old to new nodes
- Remove old nodes

Tested `high_memory` nodes by running a DAG in both `airflow-dev` and `airflow-prod`. 

~DAGs work successfully in Production because this already has a node running currently*. They do not work in Development. Looking at the cluster-autoscaler logs we can see the following~

```
I0523 15:55:32.311615       1 static_autoscaler.go:290] Starting main loop
I0523 15:55:32.311696       1 auto_scaling_groups.go:396] Regenerating instance to ASG map for ASG names: []
[...]

PR #4392, opened 2024-05-22
🔧  Add 'new' `airflow-prod` Nodes

Gary-H9:
Raised as part of [this](https://github.com/ministryofjustice/analytical-platform/issues/4389) issue.


julialawrence:
The new node groups don't have a `launch template` block. Doesn't this mean that by default the nodes will be created with the metadata required set?

jacobwoffenden (approved):
LGTM

Issue #3555, opened 2024-03-04
🏗 Build Analytical Platform Compute EKS


jacobwoffenden:
### User Story

As an Analytical Platform Product Engineer
I need to host applications that can't live on Cloud Platform
So that we can run the service

### Value / Purpose

DPAT's EKS lives in DPAT account, and can't easily be renamed, my proposal it to provision a new Modernisation Platform (`analytical-platform-compute`), and rebuild our EKS cluster (without any Data Platform components)

### Useful Contacts

@jacobwoffenden

### User Types

Analytical Platform Product Engineering

### Hypothesis

If we rebuild our _new_ EKS cluster in Analytical Platform Compute
Then we have home for Analytical Platform Dashboard, Analytical Platform Tools, and Analytical Platform Airflow

### Proposal

* Request new Modernisation Platform accounts (`analytical-platform-compute`)
* Retire new Modernisation Platform accounts (`data-platform-compute`)
* Build out EKS cluster

### Additional Information

DPAT EKS currently hosts:

* Data Platform Assets (web service serving images)
* Data Platform's Open Metadata and DataHub instances
* GitHub Actions Self Hosted Runner capability

### Definition of Done

- [x] Request new Modernisation Platform accounts (`analytical-platform-compute`)
- [x] Retire new Modernisation Platform accounts (`data-platform-compute`)
  - https://github.com/ministryofjustice/modernisation-platform/issues/7071
- [ ] Build out EKS cluster

jacobwoffenden:
02/05/24 update:
- Build out of cluster in AP Dev continues in https://github.com/ministryofjustice/data-platform/pull/4259
  - Switched to EKS API for authentication management
  - Enabled CloudWatch Observability addon 
- VPCs have been been rearchitected to consume Airflow's VPC range with the intention that our networks will replaces theirs https://github.com/ministryofjustice/modernisation-platform-environments/pull/5927/commits/6577715a65e653c4d01f66d44dce676aef63147d

jacobwoffenden:
Cluster creation blocked as `MemberInfrastructureAccess` cannot create OIDC provider due to [org SCP](https://github.com/ministryofjustice/aws-root-account/blob/main/management-account/terraform/organizations-policy-service-control.tf#L377)

```
[...]

PR #4398, opened 2024-05-23
🌐 Import `analytical-platform.service.justice.gov.uk`

jacobwoffenden:
This pull request:

Is part of https://github.com/ministryofjustice/analytical-platform/issues/3555

- Imports `analytical-platform.service.justice.gov.uk` into code so we can create records programatically instead of ClickOps 🙅 I need to create some NSs from the compute environments


julialawrence (approved):
LGTM

PR #4435, opened 2024-05-31
Remove nvidia-device-plugin-ds from prod cluster

AntFMoJ:
### Pull Request Objective

This piece of work is being tracked in [this](https://github.com/ministryofjustice/data-platform/issues/4353) GitHub Issue.

PR to remove nvidia-device-plugin-ds from production cluster. This is to fix the tf state and will be added back in a follow up PR.


julialawrence (approved):
Tick

PR #4455, opened 2024-06-04
Add IRSA to airflow dev environment

AntFMoJ:
### Pull Request Objective

This piece of work is being tracked in [this](https://github.com/ministryofjustice/data-platform/issues/4319) GitHub Issue.

This PR adds IRSA to airflow dev environment. This is to initially test with the airflow monitoring DAG [here](https://github.com/moj-analytical-services/airflow/blob/main/environments/dev/dags/monitoring/daily_pass.py).

**Note**: SCA overridden due to historic issues leading to failures, not from changes in this PR.


terraform/aws/analytical-platform-data-production/airflow/iam-policies.tf
```diff
@@ -261,6 +261,32 @@ data "aws_iam_policy_document" "airflow_dev_eks_assume_role_policy" {
 
 }
 
+##### Airflow Dev IRSA
+data "aws_iam_policy_document" "airflow_dev_monitoring_inline_role_policy" {
+  statement {
+    sid    = ""
+    effect = "Allow"
+    resources = [
+      "arn:aws:iam::${var.account_ids["analytical-platform-data-production"]}:airflow-monitoring/airflow-scheduling-testing/*",
```
julialawrence: If these are s3 buckets, the arns of buckets are in the format of `arn:aws:s3:::<bucketname>`

terraform/aws/analytical-platform-data-production/airflow/kubernetes-service-accounts.tf
```diff
@@ -0,0 +1,9 @@
+resource "kubernetes_service_account" "airflow" {
+  metadata {
+    namespace = kubernetes_namespace.dev_airflow.metadata[0].name
+    name      = "airflow"
+    annotations = {
+      "eks.amazonaws.com/role-arn" = "arn:aws:iam::593291632749:role/airflow_monitoring_dev"
```
julialawrence: It is better to use the output of the role for service account module for the annotation rather than hard-coding it.

julialawrence (changes requested):
Good start! A few fixes needed.

terraform/aws/analytical-platform-data-production/airflow/iam-policies.tf
```diff
@@ -261,6 +261,32 @@ data "aws_iam_policy_document" "airflow_dev_eks_assume_role_policy" {
...
+data "aws_iam_policy_document" "airflow_dev_monitoring_inline_role_policy" {
+  statement {
+    sid    = ""
+    effect = "Allow"
+    resources = [
+      "arn:aws:s3:::airflow-monitoring/airflow-scheduling-testing/*",
+      "arn:aws:s3:::airflow-monitoring/"
+    ]
+    actions = ["s3:GetObject", "s3:ListBucket", "s3:PutObject", "s3:DeleteObject"]
+  }
```
julialawrence: Couple of things:
* Please split this policy definition into 2 statements. One which grants list bucket to the whole bucket and the other which grants the object-level perms to the folder
* Have a look in the console of the `airflow-role` policy attached to the `airflow_dev_monitoring` role which is what I believe you're looking to replicate. It's quite a lot bigger. Are you slimming it down intentionally?

julialawrence (changes requested):
A few more general thoughts/questions that might not lead to any changes:
* Namespaces are defined in eks.tf. Should service account definitions live in that file as well? Or the reverse. Should namespace definitions be split out into their own separate file? Would love to hear your thoughts on that.
* What's the plan on removing existing kube2iam annotations. In a future PR?

terraform/aws/analytical-platform-data-production/airflow/iam-policies.tf
AntFMoJ: We didn't intend to cut down on the policy definitions, we have updated to bring in-line with existing `airflow-role` policy.

julialawrence (approved):
Nicely done! Please hold off merging until discussing with @Gary-H9

jacobwoffenden (changes requested):
Looking at the plan:

```
# module.airflow_dev_monitoring_iam_role.aws_iam_role.this[0] will be created
  + resource "aws_iam_role" "this" {
    ...
      + name                  = "airflow-monitoring-dev"
    ...
```

The IAM role (`arn:aws:iam::593291632749:role/airflow_dev_monitoring`) already exists and will need importing

jacobwoffenden (approved):

PR #4750, opened 2024-07-23
S3 - Bucket Policy Missing SecureTransport Statement 

Emterry:
This pr is to add a Bucket Policy Missing SecureTransport Statement as part of the remediation work being carried out for this issue S3 - Bucket Policy Missing SecureTransport Statement (NCC-MOJU147-2H3)
[#14](https://github.com/ministryofjustice/analytical-platform-ithc-2024/issues/14) specifically for `mojap-raw-hist-dev`

PR #4763, opened 2024-07-25
Revert "S3 - Bucket Policy Missing SecureTransport Statement "

Emterry:
Reverts ministryofjustice/analytical-platform#4750


jhpyke (approved):
lgtm as a temp measure until we can figure out why our principals are bad

PR #5377, opened 2024-09-17
:wrench: Update RDS engine version to match live

Gary-H9:
Drift has occurred between our code and AWS RDS DB Engine. This code brings them into alignment.


julialawrence (approved):
LGTM

PR #5384, opened 2024-09-17
Reworking PagerDuty Component to Avoid Undefined Dependencies

julialawrence:
## Pull Request Objective

Refactoring PagerDuty component and fixing bug that requires a user to already exist before being added to the team.
Also adds Tom to the pagerduty rota.

## Context

Plan error when adding a new pagerduty user and a new team member at the same time:

```
Error: Invalid for_each argument

  on modules/team/main.tf line 14, in resource "pagerduty_team_membership" "responders":
  14:   for_each = var.responders
    ├────────────────
    │ var.responders is a map of object, known only after apply

The "for_each" map includes keys derived from resource attributes that cannot
be determined until apply, and so Terraform cannot determine the full set of
keys that will identify the instances of this resource.

When working with unknown values in for_each, it's better to define the map
keys statically in your configuration and place apply-time results only in
the map values.

Alternatively, you could use the -target planning option to first apply only
the resources that the for_each value depends on, and then apply a second
time to fully converge.
```

The existing implementation uses the IDs of the user resource (`var.responders`) in a `set` to drive the `for_each`. This ID isn't known until `apply`, so raises the error above. Instead by using the known and stated email address of the user in `users.tf` (referenced in `teams.tf` `locals`) to drive the `for_each`, we can iterate over all users within the teams.

## Checklist

- [ ] I have reviewed the [style guide](https://docs.analytical-platform.service.justice.gov.uk/documentation/platform/infrastructure/terraform.html#terraform)
- [ ] I have self-reviewed my code


AntFMoJ (approved):
I trust this is all working as expected 🙇‍♂️

jamesstottmoj (approved):
LGTM

PR #5441, opened 2024-09-23
fix: revert `http_tokens` required setting as DAG tooling needs updating to support `"required"` setting

tom-webber:
Reverting [earlier change](https://github.com/ministryofjustice/analytical-platform/pull/5437) as this is causing DAGs to fail in airflow dev. Airflow DAG templates / DE tooling may need updating before this change can be successfully rolled out. [See thread](https://mojdt.slack.com/archives/C04M8224WCV/p1727105019008249?thread_ts=1727092810.767799&cid=C04M8224WCV)


Gary-H9 (approved):
LGTM

PR #5869, opened 2024-10-29
fix: AP-5599 remove SSE-KMS encryption references for replication config

tom-webber:
fix for changes merged in #5810

removes [extraneous SSE-KMS encryption reference](https://docs.aws.amazon.com/AmazonS3/latest/API/API_SourceSelectionCriteria.html) within replication config destination block, as source bucket doesn't use SSE-KMS


AntFMoJ (approved):
LGTM

Issue #5666, opened 2024-10-15
✨ Non-spot runner deployment for create-a-derived-table


gwionap:
### Describe the feature request.

Workloads running in create-a-derived-table currently use spot instances. We would like these workflows to use non-spot instances to ensure they are not interrupted mid-run.

### Describe the context.

We have had instances of workflows failing due to instances being pulled. See [this](https://github.com/ministryofjustice/data-platform-support/issues/864) support ticked and [this](https://mojdt.slack.com/archives/C06TFT94JTC/p1728373129107929) thread for more details

### Value / Purpose

This will ensure critical workflow runs don't fail mid run.

### User Types

data engineers, analytics engineers, end users (analysts, data scientists)

simon-pope:
Take to Amigos

simon-pope:
Take to feature request forum

julialawrence:
DBT doesn't have ability to make the job preemptable so this is needed.

jacobwoffenden:
@gwionap the new runner is now available! you will need to switch from using `analytical-platform` to `analytical-platform-non-spot` in whatever workflows you need to run on our "on-demand" fleet

gwionap:
Amazing! Thanks @jacobwoffenden.

gwionap:
Updated our daily prison run [here](https://github.com/moj-analytical-services/create-a-derived-table/pull/2335). Will let you know how it goes overnight.

gwionap:
@jacobwoffenden - workflow [failed](https://github.com/moj-analytical-services/create-a-derived-table/actions/runs/11624630876/job/32373513473) this morning with following error:
```
botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the AssumeRoleWithWebIdentity operation: Not authorized to perform sts:AssumeRoleWithWebIdentity
```

jacobwoffenden:
Resolved by adding new k8s service account to role trust policy

PR #5908, opened 2024-11-01
🚑 Add non-spot service account to trust policy

jacobwoffenden:
This pull request:

- Adds `actions-runner-mojas-create-a-derived-table-non-spot`, which is the new role from https://github.com/ministryofjustice/analytical-platform/issues/5666


Emterry (approved):
LGTM

PR #5943, opened 2024-11-01
update exempt list

Emterry:
This pr is to remove the role `airflow_prod_civil` from the exempt list as [this role has been deleted in `moj-analytical-services` repo last week](https://github.com/moj-analytical-services/airflow/pull/4741).


AntFMoJ (approved):
LGTM

PR #5961, opened 2024-11-04
📛 (Re)Enable Amazon Prometheus Proxy

jacobwoffenden:
This pull request:

> [!IMPORTANT]
> This has deployed in a **failed** state. There are 16 "Pending" pods in the DaemonSet, presumably for **existing** nodes that have no more capacity. Maybe we should cycle a node and see what happens?

- Reenables the AMP proxy in AP Prod cluster


Gary-H9 (approved):
LGTM

Issue #5594, opened 2024-10-09
📖 Increase GPU node capacity (and make associated release) for AI/ML Day.


jhpyke:
### User Story

As a participant in the AI/ML day
I need/want/expect to have GPU provisioned access
So that I can run my models

### Value / Purpose

In response to a support issue, we're raising a ticket to capture the work required to support an AI/ML in-person day on the ~19th~ 13th of November. This work will allow them to run the day without being hit with capacity issues when doing GPU based work.

### Useful Contacts

@jhpyke

### Hypothesis

If we... [do a thing]
Then... [this will happen]

### Proposal

- [ ] Increase [GPU node size requirements](https://github.com/ministryofjustice/analytical-platform/blob/c2ec411b1db2eedcf60d5e3b2923e66f82e2bee7/terraform/aws/analytical-platform-production/cluster/terraform.tfvars#L189-L193) to match required capacity
- [ ] Create a new image release on the AP with access to the GPU node labelled `hackathon` or similar
- [ ] Make that image available publically on the day of the event
- [ ] Make it private again after
- [ ] Revert changes to GPU node capacity

### Definition of Done

- [ ] GPU Capacity has been increased in time for the 19th of November
- [ ] A 'Hackathon' VS code image exists, and GPU enabled features are tested working on it
- [ ] Image is made available on the day of the event
- [ ] Image is made private after event
- [ ] Capacity is reverted to normal on 20th of November

bcrawford-moj:
Hello @jhpyke . I've just noticed the date on this should be 13th November.  Could we please add this capability for the following teams and users please:

https://github.com/orgs/moj-analytical-services/teams/bold-ai-for-linked-data
https://github.com/orgs/moj-analytical-services/teams/pem-data-science
https://github.com/evincent-moj
https://github.com/monstermaker
https://github.com/zoewalkerMoJ

Thanks!

julialawrence:
Created a new release: https://controlpanel.services.analytical-platform.service.justice.gov.uk/release/119/

Currently restricted to Billy, will unrestrict tomorrow.

PR #6098, opened 2024-11-12
🎮 Increase GPU node capacity for AI/ML day

jacobwoffenden:
This pull request:

- Is part of https://github.com/ministryofjustice/analytical-platform/issues/5594
- Increases desired to 10
- Increases max to 15


julialawrence (approved):
I manually upped the desired number in the console otherwise this will fail on apply with the same error as the previous one where the min is greater than desired. Desired is being ignored somewhere. A quick look through the module code didn't turn up anything but a more thorough look is needed.

PR #6182, opened 2024-11-26
add role for ct tact list to be consumed by self hosted runner in apc

Emterry:


terraform/aws/analytical-platform-data-production/github-actions-roles/mojas-ct-tact-list.tf
jacobwoffenden, line 1: https://docs.analytical-platform.service.justice.gov.uk/documentation/platform/infrastructure/terraform.html#naming

PR #6188, opened 2024-11-27
Fixing deprecation of role_arn supplied via backend_config

Emterry:
Reference issue:
https://github.com/hashicorp/terraform/issues/27579

We are having issues running init due to role_arn being passed directly to backend config having been deprecated as a configuration option.

Applying a fix.


julialawrence (approved):
LGTM

Issue #6216, opened 2024-12-02
🌬 Airflow POC


YvanMOJdigital:
### User Story

As an Analytical Platform Engineer
I need a new experience for operating Airflow
So that we don't have to burn precious brain power on understanding the existing setup

### Value / Purpose

_TBC_

### Useful Contacts

@jacobwoffenden @jhpyke 

### User Types

Analytical Platform Engineering / Data Engineering

### Hypothesis

If we... _TBC_
Then... _TBC_

### Proposal

_TBC_

### Definition of Done

_TBC_

YvanMOJdigital:
draft tickets for next steps and agree priority

YvanMOJdigital:
also agree support strategy for airflow: what we do or do not support

PR #6268, opened 2024-12-11
🌬️ Zephyr PoC

jacobwoffenden:
This pull request:

- Is part of https://github.com/ministryofjustice/analytical-platform/issues/6216

TODO:
- [ ] MWAA is not running with CMK
- [ ] ALB target group uses hard coded IPs
- [ ] Airflow local settings don't seem to be taking effect, maybe because there are no DAGSs?
- [ ] Webserver complains it can't install requirements - is this a problem?


jacobwoffenden:
Need to delete these resources

Issue #4622, opened 2024-07-04
📖 Migrate jobs from `github-actions-moj-CoFBZ3xV` EKS Cluster


Gary-H9:
### Proposal

>[!NOTE]
> This ticket precedes [this](https://github.com/ministryofjustice/analytical-platform/issues/4623) ticket to decommission the cluster.

As per [this](https://mojdt.slack.com/archives/C04M8224WCV/p1720090731674809) thread. 

Migrate any workloads that run on the `github-actions-moj-CoFBZ3xV` cluster in `analytical-platform-management-production`.

### Value / Purpose

Consolidate compute, reduce cost. 

### Definition of Done

- [x] Establish what workloads are running on the cluster, what they are doing (e.g. is it building a container?) and what role they are using
- [x] Liaise with owner to gather any requirements (e.g. can repos that interact with CT data run on GitHub hosted?) and inform of our intention to migrate and sunset offering
- [x] Identify which jobs can move to GitHub Actions hosted runners (GHA) vs. self hosted runners (SHR)
- [x] Create a role for each job, GitHub assumable for GHA and IRSA for SHR
- [x] Create/update any shared GitHub Actions actions utilised by workloads
- [ ] Create any self hosted runners in APC
- [ ] Migrate workloads to their new location

julialawrence:
Split this ticket into 2:

1. Identify and profile workloads (see proposal)
2. Step to retire

jacobwoffenden:
jobs using `management-infrastructure`

https://github.com/search?q=org%3Aministryofjustice+OR+org%3Amoj-analytical-services+%22self-hosted%2C+management-infrastructure%22&type=code

jobs using `management-ecr`

https://github.com/search?q=org%3Aministryofjustice+OR+org%3Amoj-analytical-services+%22self-hosted%2C+management-ecr%22&type=code

Emterry:
Spreadsheet for jobs using ```management-infrastructure``` and  ``` management-ecr ``` https://docs.google.com/spreadsheets/d/1ZkTFF6MlbkZvk8sA7A9Cyd-JoJVcS6koE4FM2xuNZA4/edit?usp=sharing

Service accounts and roles on the cluster https://docs.google.com/document/d/10z0cGW0wTHP1rA0CPwKz5CcHobqceAD_s9FPdchKgBA/edit?usp=sharing

Emterry:
[...]

PR #6567, opened 2025-01-16
feat: add dynamic `s3Locations` statements to allow oidc roles access to specified s3 paths

tom-webber:
# Pull Request Objective

This piece of work is being tracked in #4622.

#4622 involves switching over workflow processes to run on GitHub Actions runners assuming roles by OIDC (roles specified in the files edited in this PR) rather than running on self-hosted runners (see [this PR](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/785/files) for an example). In order to do this, the new OIDC roles need permissions to be able to carry out all of the duties of the existing roles.

A change is required to allow the `analytics-platform-helm-charts` role to deposit artefacts (which [fails without this change](https://github.com/ministryofjustice/analytics-platform-helm-charts/actions/runs/12791904830/job/35661137176#step:8:109)) in s3. This is separate from the requirement for state bucket access, so a key:value pair was added to the oidc-roles json that creates a dynamic statement in the `github_oidc_role` for each role.

## Checklist

- [x] All checks have passed (or override label applied, if I've
- [x] I have reviewed the checks and can attest they're as expected

### DIFF
--- a/terraform/aws/analytical-platform/oidc/configuration/assumable-roles.json
+++ b/terraform/aws/analytical-platform/oidc/configuration/assumable-roles.json
@@ -7,6 +7,7 @@
         "stateBucketKey": ".pulumi/"
       }
     ],
+    "s3Locations": [],
     "repositories": ["ministryofjustice/moj-data-transfer-api"],
     "targets": [
       "analytical-platform-data-production",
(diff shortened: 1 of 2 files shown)


julialawrence (approved):
LGTM

PR #7211, opened 2025-03-13
Update permissions in Lake Formation configuration

matt-heery:
Refine permissions in the Lake Formation resource to only include "SELECT" instead of "DESCRIBE" and "SELECT."


Gary-H9 (approved):
LGTM

PR #7334, opened 2025-03-25
🚑 Fix CaDeT IRSA role

jacobwoffenden:
## Proposed Changes

- ARN was incorrectly constructed using management account ID 


jhpyke (approved):
Nothing but net 🏀

PR #7586, opened 2025-04-16
🔑  PowerBI EC2: Replace Public Key

Gary-H9:
This PR replaces a key which we can't seem to use to login. 

The key was generated with the following command: 
` ssh-keygen -m PEM -t rsa` 

As per [these](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/create-key-pairs.html) docs.


tamsinforbes (approved):
🌝🌖🚀LGTM🚀🌘🌚

PR #7621, opened 2025-04-22
♻️ Switch to sensitive credentials for GitHub application

jacobwoffenden:
## Proposed Changes

- I thought Terraform would be smart enough to mask this in a plan, but it wasn't and leaked it. I have rotated this client secret.


michaeljcollinsuk (approved):
LGTM

Issue #6639, opened 2025-01-31
🛫 CaDeT Airflow Deployments Trial


jhpyke:
### User Story

As a maintainer of CaDeT Deployments via Airflow
I need to run a trial of deploying via Airflow
So that I can be confident it is a suitable solution for our deployments.

### Value / Purpose

We're collecting signups for deployments via Airflow, with signups closing next Wednesday CoP. As those signups come in, we should be creating deployments for them so that they can see that Airflow can handle their deployments.

### Useful Contacts

@jhpyke

### Proposal

We should support two broad categories of deployment:
Dual Running - Avoid interfering with existing prod runs:
- Deploys against `Sandpit`
- Runs on same schedule as currently, but 12 hours delayed.

Solo Running:
- Deploys Against Prod
- Runs on the schedule of the existing github action
- Existing github action should be moved to `workflow_dispatch`.
- You may need to alter the [catalog protection policy](https://github.com/ministryofjustice/analytical-platform/blob/052e29e4ef87ccef733899d8f4cfd638aa38293c/terraform/aws/analytical-platform-data-production/data-engineering-pipelines/locals.tf) for specific databases in order to allow the deployer role to successfully deploy to them

### Additional Information

Workflows should be spun up with a particular attention to the `--target` - assume users don't want `prod` unless they either specify they do, or specify solo running.

### Definition of Done

- [ ] DAGs exist for all participating workflows.
- [ ] Dual Running DAGs are able to deploy on their shifted schedules
- [ ] DAGs are monitored for errors and those are resolved by the end of the test period.

jacobwoffenden:
Blocked by #6712

MattFunnell:
Hi - is there an expected completion date for this issue? We have airflow jobs in old DE airflow prod that have been migrated but can't be finalised until the trigger for the jobs is resolved (replicating this: https://github.com/moj-analytical-services/create-a-derived-table/blob/main/scripts/data/airflow-dag-trigger.yaml)

MattFunnell:
Just to add - I see Prod-DE Airflow is being retired in December, and this is a dependency for us moving our jobs out from the old prod

jhpyke:
Hi,

[...]

PR #8131, opened 2025-07-15
Add cadet airflow role for consumtion in AP airflow

jhpyke:
# Pull Request Objective

This piece of work is being tracked in
[this](https://github.com/ministryofjustice/analytical-platform/issues/6639
)
GitHub Issue.

This creates a role for use in Airflow by CaDeT deployments. A matching PR is to follow in AP Airflow repo.

## Checklist

> [!NOTE]
> Each items should be checked. Skipping below checks could delay your PR review!

### Additional Comments

This cuts the permission for assumption for the Quicksight stuff (because that's GH actions specific at least initially) as well as the `createCLItoken` permission as it's not relevant to the new platform (yet?)

### DIFF
--- a/terraform/aws/analytical-platform-data-production/airflow-service/iam-policies.tf
+++ b/terraform/aws/analytical-platform-data-production/airflow-service/iam-policies.tf
@@ -314,3 +314,112 @@ module "kms_iam_policy" {
   name   = "kms"
   policy = data.aws_iam_policy_document.kms.json
 }
+
+data "aws_iam_policy_document" "create_a_derived_table" {
+  statement {
+    sid    = "BucketAccess"
+    effect = "Allow"
+    actions = [
+      "s3:List*",
+      "s3:GetObject*",
+      "s3:GetBucket*",
+      "s3:DeleteObject*",
+      "s3:PutObject*"
+    ]
+    resources = [
+      "arn:aws:s3:::mojap-derived-tables/*",
+      "arn:aws:s3:::mojap-derived-tables",
+      "arn:aws:s3:::dbt-query-dump/*",
+      "arn:aws:s3:::dbt-query-dump",
+      "arn:aws:s3:::mojap-manage-offences/ho-offence-codes/*",
+      "arn:aws:s3:::mojap-manage-offences",
+      "arn:aws:s3:::mojap-hub-exports/probation_referrals_dump/*",
+      "arn:aws:s3:::mojap-hub-exports",
+      "arn:aws:s3:::alpha-app-opg-lpa-dashboard",
+      "arn:aws:s3:::alpha-app-opg-lpa-dashboard/dev/models/domain_name=opg/*",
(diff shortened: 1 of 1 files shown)


terraform/aws/analytical-platform-data-production/airflow-service/iam-policies.tf
```diff
@@ -314,3 +314,112 @@ module "kms_iam_policy" {
...
+    actions = [
+      "secretsmanager:GetSecretValue",
+      "secretsmanager:DescribeSecret",
+      "secretsmanager:ListSecrets",
+    ]
+    resources = [
+      "arn:aws:secretsmanager:*:*:secret:/alpha/airflow/airflow_dev_cadet_deployments/cadet-deploy-key/*",
+      "arn:aws:secretsmanager:*:*:secret:/alpha/airflow/airflow_dev_cadet_deployments/slack_bot_key/*"
+    ]
+  }
```
jacobwoffenden: We should use Airflow native secrets for this

terraform/aws/analytical-platform-data-production/airflow-service/iam-policies.tf
```diff
@@ -314,3 +314,112 @@ module "kms_iam_policy" {
...
+    actions = [
+      "s3:List*",
+      "s3:GetObject*",
+      "s3:GetBucket*"
+    ]
+    resources = [
+      "arn:aws:s3:::*",
+      "arn:aws:s3:::*/*"
+    ]
+  }
```
jacobwoffenden, line 357: Any chance we can review this?

jhpyke: I think potentially with buy in from DE we could get this down to a list of buckets, but it'll require a fundamental change of the current service model (I.E. users will be expected to add their data source location to this policy if they want it in CaDeT) as this is required currently because we're in a chicken and egg situation with the data sources - we need to be able to be able to access them to use them as a source in builds, but we don't know the location of a table until it's already available in the project as a source

jhpyke: I think best we could do unilaterally is limiting it to specific permissions rather than `List*` etc, but that doesn't really address the issue of the resources

terraform/aws/analytical-platform-data-production/airflow-service/iam-policies.tf
jhpyke: Agree, removed

terraform/aws/analytical-platform-data-production/airflow-service/iam-policies.tf
jacobwoffenden: Sure - Ultimately this just gives this role full access to data production, so wanted to check.

Appreciate legacy requirements are always ✨ fun ✨

jacobwoffenden (approved):

Issue #8135, opened 2025-07-15
🐞 JML Report is not running


jacobwoffenden:
### Describe the bug.

Don't ask me how I ended up on it's function page, but its currently idle and not running

### To Reproduce

1. n/a

### Expected Behaviour

It runs

Gary-H9:
[Repository](https://github.com/ministryofjustice/analytical-platform-jml-report) in question. Last [release](https://github.com/ministryofjustice/analytical-platform-jml-report/releases/tag/1.4.6) June 12th. Previous release in February. 

This lambda exists in APDP. Looking at the lambda I can see the message "The function analytical-platform-jml-report is idle. To reactivate your function, choose Restore." 

The [cron](https://github.com/ministryofjustice/analytical-platform/blob/d3cc1e5bcce489c85c27b6fb959674c5d6a364c6/terraform/aws/analytical-platform-data-production/joiners-movers-leavers/cloudwatch-event-rules.tf) for this lambda is set to the first day of every month at 2am. 

Clicked restore and got the following message. 

"Successfully restored the function analytical-platform-jml-report." 

In Configuration > Triggers I can see the error: 
`No target to arn:aws:lambda:eu-west-2:593291632749:function:analytical-platform-jml-report could be found on the rule jml-lambda-trigger.`

Gary-H9:
There is not a cloudwatch event target to tie the event rule to the lambda function. The above PR resolves this.

Gary-H9:
<img width="1385" height="400" alt="Image" src="https://github.com/user-attachments/assets/5bf4ae53-dcd0-4347-bc8c-ec5903a35e8b" />

Above PR merged. Moving to review.

Gary-H9:
Moving to blocked until we establish if this is still required. Slack thread on this [here](https://mojdt.slack.com/archives/C04M8224WCV/p1752759145714749).

jacobwoffenden:
Find email and we'll get in touch with them @Gary-H9

Gary-H9:
Greg has sent a query and I've set a reminder to retire this if we don't get a response in two weeks.

Gary-H9:
This will be picked up tomorrow if no response.

Gary-H9:
Closing this as it will be resolved in [here](https://github.com/orgs/ministryofjustice/projects/27?pane=issue&itemId=122769880&issue=ministryofjustice%7Canalytical-platform%7C8212).

PR #8137, opened 2025-07-15
:wrench: JML: Add target for CloudWatch Event Target

Gary-H9:
This PR adds an event target between the cloudwatch event rule and the lambda function - effectively tying them together. This PR is related to [this](https://github.com/ministryofjustice/analytical-platform/issues/8135) bug.


BrianEllwood (approved):
LGTM

PR #8504, opened 2025-08-19
:wrench: Add `.dependabot-terraform-ignore` to airflow directory

Gary-H9:
The purpose of this PR is to have Dependabot ignore the Airflow directory in APDP going forward as this is going to be retired.


BrianEllwood (approved):
LGTM

PR #8507, opened 2025-08-19
add and remove users from de support rota

guyvw1:
An update to add and remove users from the data engineering support rota


AntFMoJ (approved):
LGTM

PR #8605, opened 2025-08-27
🔥 Retire EM LF infrastructure

Gary-H9:
Supersedes [this](https://github.com/ministryofjustice/analytical-platform/pull/8604) PR.


Gary-H9:
Since this configuration had a provider set in each EM account I've manually deleted these resources. As such I will reopen the original PR which simply removes all of the code relating to these.

PR #9332, opened 2025-11-18
Rename coat athena workgroup

levgorbunov1:
The current name is too granular. This PR give a more general name to the COAT athena workgroup, which is more appropriate given how it is used.


Gary-H9 (approved):
LGTM

PR #9445, opened 2025-12-04
COAT Integration: Add S3 bucket for enriched CUR data

levgorbunov1:
**Summary**
Refactor coat-integration Terraform module to use for loops to create 2nd enriched CUR S3 bucket.

**Why**
We would like to replicate enriched CUR GreenOps data from a bucket in the coat-production MP account into APDP, so that the data is accessible to airflow, so we can begin to ingest GreenOps data for our dashboards.

**How**
- locals has been refactored to include a map of our S3 buckets.
- This map is used with for loops to create kms and s3 resources from the same configuration.
- A moved block has been added to remap the original cur hourly bucket address to the new module address, accounting for use of the for loop.

### DIFF
--- a/terraform/aws/analytical-platform-data-production/coat-integration/kms-keys.tf
+++ b/terraform/aws/analytical-platform-data-production/coat-integration/kms-keys.tf
@@ -26,7 +26,7 @@ module "coat_kms_keys" {
       principals = [
         {
           type        = "AWS"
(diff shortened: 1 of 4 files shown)


mamonu (approved):

jamesstottmoj (approved):
LGTM

PR #9473, opened 2025-12-11
CADET role perms for COAT greenops KMS key

levgorbunov1:
Error in CADET: https://github.com/moj-analytical-services/create-a-derived-table/actions/runs/20135323302/job/57786955969?pr=5424#step:19:50 

Where the KMS keys are defined in AP: https://github.com/ministryofjustice/analytical-platform/blob/main/terraform/aws/analytical-platform-data-production/coat-integration/kms-keys.tf

We need to give the CADET role assumed by GitHub Actions, permission to use the KMS key that our GreenOps data in APDP S3 is encrypted with, so that it can access data in that bucket.


AntFMoJ (approved):
LGTM

PR #9613, opened 2026-01-27
Remove Components To Be Retired

BrianEllwood:
This pull request makes a minor update to the scheduled issue checklist in the GitHub Actions workflow. Specifically, it removes two environments from the checklist, because they are no longer needed.

- Removed `analytical-platform-next-poc-hub` and `analytical-platform-next-poc-producer` from the environment checklist in `.github/workflows/schedule-issue-modernisation-platform-environments.yml` as these do not need to be maintained as soon to be retired.


Gary-H9 (approved):
LGTM

PR #9734, opened 2026-02-17
Update Trivy Ignore to New Format

BrianEllwood:
The trivy syntax has changed. The old format `#trivy:ignore:avd-aws-0006:Not encrypting... `(with colon immediately after rule ID) is no longer supported. The new format should use a space instead of a colon

This pull request makes changes to the Trivy ignore comments in the `athena-workgroups.tf` file. The comments are now consistent with the new formart.


tamsinforbes (approved):
🌝🌖🚀LGTM🚀🌘🌚

Issue #9751, opened 2026-02-20
🔱 Chainguard Ingress NGINX Replacement


jacobwoffenden:
### Context

[Ingress NGINX is being retired](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/)

We are running Ingress NGINX in all of our clusters:
- analytical-platform-development
- analytical-platform-production
- analytical-platform-compute-development
- analytical-platform-compute-test
- analytical-platform-compute-production

### Proposal

We are working with Chainguard to see if their commercial offering of Ingress NGINX Controller is a suitable replacement, until we either

A) Migrate to Data Platform
or
B) Migrate to a Gateway API based ingress solution

---

This work should be performed in `analytical-platform-development` first rather than `analytical-platform-compute-development` as we host more ingresses in the tooling clusters

### Spike requirements

Platform Engineering, 1 week

### Definition of Done

- [x] Upgrade Ingress NGINX to 1.14
- [x] @jacobwoffenden grants whoever is working on this access to Chainguard Console
- [x] [Create a pull token for Chainguard's registry](https://edu.chainguard.dev/chainguard/chainguard-images/chainguard-registry/authenticating/#managing-pull-tokens-in-the-chainguard-console)
- [x] Deploy pull token via [Terraform](https://github.com/ministryofjustice/analytical-platform/tree/main/terraform/aws/analytical-platform-development/cluster)
- [x] Deploy Chainguard Ingress NGINX via [Flux](https://github.com/moj-analytical-services/analytical-platform-flux)

Gary-H9:
Since we are running [v4.12.1](https://github.com/moj-analytical-services/analytical-platform-flux/blob/826907dc0cc6e8b1dd260f07b53d633e0a5e5ab7/overlays/production/core/ingress-nginx/release-patch.yaml#L9) of the helm chart, this corresponds to [v1.12.1](https://github.com/kubernetes/ingress-nginx#supported-versions-table). This applies to both development and production clusters.

Gary-H9:
I've investigated the above upgrade today. From my reading of things, there is no reason to not skip `v1.13.x` i.e. we can go straight from `v1.12.1` to `v1.14.3`. Some more details in the draft PR. 

[...]

PR #9771, opened 2026-02-24
🔒 Development: Create Chainguard Kubernetes Secret 

Gary-H9:
👷 🚧 This PR has already been applied 🚧 👷

This PR relates to [this](https://github.com/ministryofjustice/analytical-platform/issues/9751) piece of work. 

This change creates a Kubernetes secret of type `kubernetes.io/dockerconfigjson` in the `ingress-nginx` namespace, sourced from AWS Secrets Manager.

The secret stores Docker registry credentials for `cgr.dev` (Chainguard's container registry), enabling the cluster to authenticate when pulling Chainguard images.

We use the `kubernetes.io/dockerconfigjson` secret type because Kubernetes requires image pull credentials to be in the standard Docker config JSON format. This allows the secret to be referenced directly as an `imagePullSecrets` entry in pod specs or Helm chart values, which is the mechanism Kubernetes uses to authenticate against private container registries during image pulls.

The AWS Secrets Manager secret (`chainguard-image-pull-token` in `analytical-platform-management-production`) stores the credentials as JSON containing `username` and `password` fields, which are decoded and assembled into the expected Docker config structure at apply time.

### DIFF
--- a/terraform/aws/analytical-platform-development/cluster/data.tf
+++ b/terraform/aws/analytical-platform-development/cluster/data.tf
@@ -59,6 +59,11 @@ data "aws_secretsmanager_secret_version" "pagerduty_analytical_platform_storage_
   secret_id = "pagerduty/analytical-platform-storage/integration-keys/cloudwatch"
 }
 
+data "aws_secretsmanager_secret_version" "chainguard_image_pull_token" {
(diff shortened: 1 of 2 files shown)


BrianEllwood (approved):
LGTM

PR #10025, opened 2026-04-08
update de support rota, remove william add gw to g7

guyvw1:
remove william, add gw to g7 rota. Also updating SC's details, and tidying up the code


laurentb4 (approved):
LGTM

PR #10046, opened 2026-04-10
update id values for pagerduty users and rebuild schedule

guyvw1:
Accidentally deleted some users from PagerDuty after deleting the dev data engineer support rota in #10045 . Asked operations team to reinstate deleted users. This PR rebuilds the DE support rota, and updates the user id values to their new accounts.


laurentb4 (approved):
LGTM

PR #10212, opened 2026-05-13
Bugfix/remove ant

michaeljcollinsuk:
Previous attempt to remove Ant failed to apply when it was merged https://github.com/ministryofjustice/analytical-platform/pull/10210/changes

I have now removed Ant by applying locally in two steps:
1. Remove Ant from the schedule
2. Remove Ant from the users

This has no applied successfully so merging should have no further effect.


Gary-H9 (approved):
LGTM

PR #10218, opened 2026-05-13
first go at sandbox fabric module

Thomas-Hirsch:
POC for fabric iam module. Not sure if it should live here or elsewhere, but probably the latter if it is to be more genericised


moj-supratikc:
Couple comments so far:
1. Separate OIDC components into a separate module
2. Research how best to pass secrets (e.g. tenant id for the OIDC component, client id for the IAM role component) in terraform - exploring how it is currently done in other components in the ap repo as well. Options so far:
  2a. Create secretsmanager secret ahead of time and use data blocks to pull in - is there a clean way to also manage the secret creation in terraform or do we not need to?

PR #10386, opened 2026-07-01
remove issue-metrics workflow

Gary-H9:
Based on the last three of these that I've just closed, they have no update, I don't think this is used at all so removing the workflow to prevent these issues being created.


tamsinforbes (approved):
🔥

PR #10572, opened 2026-08-18
Externalise Splink S3 KMS key-user principals to a JSON access list

dami-moj:
# Pull Request Objective

Moves the KMS key-use principal identifiers for the Splink prod and test S3 keys into an auditable kms_key_users.json file, decoded in locals and referenced by both key policies, so access is managed in one reviewable place and new buckets can be added easily. Prod is scoped to the Airflow roles while test also includes the named users.

### DIFF
--- a/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms.tf
+++ b/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms.tf
@@ -64,6 +64,25 @@ data "aws_iam_policy_document" "s3_kms_policy" {
       values   = ["s3.${data.aws_region.current.region}.amazonaws.com"]
     }
   }
+
+  statement {
+    sid = "AllowAirflowKeyUse"
+
+    principals {
+      type        = "AWS"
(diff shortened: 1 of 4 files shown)


JamesS-MoJ (dismissed):
Externalising role names for easier maintenance

JamesABaker (dismissed):
Looks good.

I anticipate "splink_s3_bucket" will fail until the roles are set up in [`analytical-platform-airflow/environments/production/laa/`](https://github.com/ministryofjustice/analytical-platform-airflow/tree/main/environments/production/laa). I think we would need to set this up in prod similarly to `environments/development/laa/s3-ops/workflow.yml` `environments/development/laa/si-access-test/workflow.yml`.

This is not a blocking issue; the workflows can follow.

JamesABaker (approved):
```
{
  "splink_s3_bucket": {
    "key_users": ["alpha_user_jamess-moj"]
  },
```

In lieu of a prod airflow, giving @JamesS-MoJ permissions makes sense. These should be revoked once we have airflow prod workflows set up.

jacobwoffenden (approved):

PR #10579, opened 2026-08-20
created s3 modules and corresponding changes

arun-moj-laa:
# Pull Request Objective

This piece of work is being tracked in
[this](https://github.com/ministryofjustice/analytical-platform/issues/<your_issue_number_here>)
GitHub Issue.

## Checklist

> [!NOTE]
> Each items should be checked. Skipping below checks could delay your PR review!

- [ ] I have reviewed the [style guide](https://docs.analytical-platform.service.justice.gov.uk/documentation/platform/infrastructure/terraform.html#terraform)
- [ ] I have self-reviewed my code


terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/laa-splink-buckets.tf
```diff
@@ -0,0 +1,258 @@
...
+    target_prefix = "s3access/${each.value}/"
+  }
+
+  lifecycle_rule = [
+    {
+      id      = "expire-noncurrent-versions"
+      enabled = true
+
+      noncurrent_version_expiration = {
+        days = 90
```
JamesABaker: Test bucket was 3 days. How come we are raising to 90 days? I think this is a good idea, but I wonder if maybe even longer is worth doing for this project?

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/locals.tf
```diff
@@ -12,13 +12,77 @@ locals {
...
+  splink_test_bucket_name          = local.app.splink_test_bucket_name
+  splink_test_sns_topic_name       = "splink-s3-event-notification-topic-test"
+  splink_test_event_rule_name      = "splink-bucket-event-rule-test"
+  bucket_access_policy_statements = {
+    (local.splink_search_input_bucket_name) = [
+      {
+        Sid       = "ReadOnlyBucket"
+        Effect    = "Deny"
+        Principal = "*"
+        Action    = ["s3:PutObject", "s3:DeleteObject", "s3:DeleteObjectVersion"]
```
JamesABaker: I would want maybe @JamesS-MoJ to confirm this. Would `["s3:GetObject", "s3:GetObjectVersion"]`  with an additional statement denying  `s3:ListBucket ` on the bucket resource be more like the behaviour we want? That way AP users can drop search requests into the bucket without being able to read each other's submissions, while Airflow, the only principal with KMS decrypt rights, can read and process them.

If I have understood the permissions proposed, AP users are denied  PutObject , so no search requests can ever be submitted and the pipeline never triggers.

JamesABaker (changes requested):
Arun, thanks so much for submitting this!

This PR creates all five new buckets from the bucket map in one PR. All use a shared  for_each  pattern with standard security config (KMS, TLS, versioning, CloudWatch alarms, EventBridge).

- laa-splink-search-input
- laa-splink-search-output
- laa-splink-source-file-input
- laa-splink-source-file-output
- laa-search-audit

This completes the total buckets required for phase 1.

Before we merge and proceed, I have a few queries on the proposed bucket configs.

There is also a question which we can incorporate later about object locking these buckets. I don't think we need to raise this issue now unless @JamesS-MoJ does.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/laa-splink-buckets.tf
```diff
@@ -0,0 +1,258 @@
...
+          Resource = [
+            "arn:aws:s3:::${each.value}",
+            "arn:aws:s3:::${each.value}/*"
+          ]
+          Condition = {
+            NumericLessThan = {
+              "aws:TLSVersion" = "1.2"
+            }
+          }
+        },
```
dami-moj: I also think we are missing `DenyBucketDeletion` and `object_lock_enabled = true`, both are in the `splink_search_s3` pattern in splink_s3.tf. neither is here. Just wondering if its worth consideration.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/laa-splink-buckets.tf
JamesS-MoJ: I would make it something like 10 years, for now

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/locals.tf
JamesS-MoJ: I don't understand this, surely you can't deny PutObject or we can never put in the bucket

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms.tf
```diff
@@ -53,9 +53,16 @@ data "aws_iam_policy_document" "s3_kms_policy" {
       test     = "StringEquals"
       variable = "kms:EncryptionContext:aws:s3:arn"
 
-      values = [
-        "arn:aws:s3:::${local.splink_bucket_name}"
-      ]
+      values = concat(
+        ["arn:aws:s3:::${local.splink_bucket_name}"],
+        [
+          "arn:aws:s3:::${local.splink_search_input_bucket_name}",
```
dami-moj: The KMS encryption context has been extended to include all of the new bucket ARNs, which looks right. However, can we also confirm the intended KMS principals for each bucket? 

I think `kms:EncryptionContext:aws:s3:arn` constrains where the key can be used, but it does not itself grant a principal permission to use the key (correct me if I am wrong). We may need explicit KMS key users per bucket/workflow role so that only the required Airflow/service principals have `Encrypt` / `Decrypt` / `GenerateDataKey` access.

Could we define the expected KMS access matrix for each bucket (maybe with a placeholder) rather than granting the same key access across all of them?

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/laa-splink-buckets.tf
JamesABaker: I think we can address this on Monday - it might be for Jav to decide rather than us.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/locals.tf
JamesABaker: The  `ArnNotEquals`  condition for [def9cee](https://github.com/ministryofjustice/analytical-platform/pull/10579/commits/def9ceeeed562a07133f8c453b6e795aeccf3ab6) means named principals in  splink_s3_key_user_arns  (currently just you `alpha_user_jamess-moj`) can still  `PutObject`. The bucket isn't completely write-locked.
 
 I think there is an issue that this might get fiddly for production Airflow roles. These will be added to  `splink_s3_key_user_arns`, and they'll also be exempted from the write deny. Airflow gets write access to the input bucket, when it _should_ only be reading. 

I don't want this to be a hard blocker, but @arun-moj-laa is there another approach we could try that you can think of?

arun-moj-laa: I will review this

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/application_variables.json
```diff
@@ -3,6 +3,11 @@
   "logging_bucket_name": "moj-analytics-s3-logs",
   "splink_bucket_name": "laa-splink-search-s3",
   "splink_test_bucket_name": "laa-splink-search-s3-test",
+  "splink_search_input_bucket_name": "laa-splink-search-input",
+  "splink_search_output_bucket_name": "laa-splink-search-output",
+  "splink_source_input_bucket_name": "laa-splink-source-file-input",
+  "splink_source_output_bucket_name": "laa-splink-source-file-output",
+  "splink_audit_bucket_name": "laa-search-audit",
```
rajindersinghsandhu, line 17: Better to have env prefix with bucket names so that you can have same buckets in other envs

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/laa-splink-buckets.tf
JamesABaker: We have a steer from Marisa Baker in LAA legal to make this 14 years.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms.tf
```diff
@@ -53,9 +53,16 @@ data "aws_iam_policy_document" "s3_kms_policy" {
...
+          "arn:aws:s3:::${local.splink_search_output_bucket_name}",
+          "arn:aws:s3:::${local.splink_source_input_bucket_name}",
+          "arn:aws:s3:::${local.splink_source_output_bucket_name}",
+          "arn:aws:s3:::${local.splink_audit_bucket_name}"
+        ]
+      )
```
JamesABaker, line 67: This is the component I was thinking about over the weekend. My preference is to explode this out and explicitly and repetitively define the bucket permissions. This will reduce the complexity for a human to understand the inherited permissions, and allow us to control the nuances of read write access in different buckets.

I will open a PR explaining what I mean.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/application_variables.json
JamesABaker: I think this is a valid point and the correct way to go, but I would push this down the road a bit.

In practice there's no dev/test equivalent for the LAA data environment right now.  `laa-splink-search-s3-test`  is a test bucket inside production, not a separate account. Hence the `-test` suffix pattern rather than an env prefix.

My concern on changing bucket names at this stage is two fold (i) Other people are developing in parallel using these agreed bucket names, and we would have to carefully coordinate the change, and (ii) because there is no env for test or dev currently, we will end up in the strange and confusing naming convention of `prod_<bucket>_test`.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/locals.tf
JamesABaker: Sorry to unresolve the thread; my fault for not being more clear. 

I think there are two issues here.

1. Deny is on wrong actions [unresolved]
2. ARN list conflates KMS users with write principals [resolved by [7e6392f](https://github.com/ministryofjustice/analytical-platform/pull/10579/commits/7e6392f2572e0f5b41e95587f6b79f983c01391f)]

```yaml
Action = ["s3:PutObject", "s3:DeleteObject", "s3:DeleteObjectVersion"]
Effect = "Deny"
```

This looks like it is still denying writes. I think GetObject / GetObjectVersion / ListBucket  is what needs denying.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms.tf
JamesABaker: @arun-moj-laa what do you think of this approach?

- https://github.com/ministryofjustice/analytical-platform/pull/10585 

I branched off of main so there is some noise until you update branch. Broadly speaking this replaces the  `for_each`  +  `locals.tf`  policy map approach with five explicit bucket files, each following Kyle's  `splink_s3.tf`  pattern (one file per bucket, policy fully inline). 

The intent is that anyone reviewing a bucket's permissions can read a single file and understand exactly what it does without cross-referencing  `laa-splink-buckets.tf` ,  `locals.tf `,  `kms_key_users.json `, and  `application_variables.json`  simultaneously.

I've also adopted your read/write principal separation from your latest commits.  splink_s3_phase1_input_readers  and  splink_s3_phase1_output_writers  offer a good granularity improvement over my earlier single list.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/laa-splink-buckets.tf
```diff
@@ -0,0 +1,261 @@
...
+    local.splink_source_input_bucket_name,
+    local.splink_source_output_bucket_name,
+    local.splink_audit_bucket_name
+  ])
+
+  bucket = module.s3_buckets[each.key].s3_bucket_id
+  name   = "EntireBucket"
+}
+
+resource "aws_cloudwatch_metric_alarm" "additional_bucket_size" {
```
JamesS-MoJ: I don't know why this is here? I don't see it in Kyle's template.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/laa-splink-buckets.tf
```diff
@@ -0,0 +1,261 @@
...
+  alarm_actions       = [aws_sns_topic.splink_bucket_alerting_topic.arn]
+  treat_missing_data  = "notBreaching"
+
+  dimensions = {
+    BucketName  = module.s3_buckets[each.key].s3_bucket_id
+    StorageType = "StandardStorage"
+  }
+}
+
+resource "aws_cloudwatch_metric_alarm" "additional_object_count" {
```
JamesS-MoJ: Similar to above, I don't know why this is here?

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/locals.tf
arun-moj-laa: This block based on the assumption of kms_users role placeholders. But permission action act based on the condition

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/application_variables.json
arun-moj-laa: leave it as it is based on above comment.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms.tf
arun-moj-laa: The reason I designed in this way to make it have module design to eliminate redundancy and follow DRY principle but this approach add complexity of code. Agree if we want to eliminate cross-reference then obviously make it exclusive files for each bucket.

basically what we want to trade off, we can proceed in that way. Latter we can bring best practice

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/locals.tf
JamesS-MoJ: This looks good from a Terraform code point of view, but I think the config above in JSON doesn't give the correct permissions. Also, here we are only giving an explicit deny. I think that means that we still need an explicit Allow somewhere to actually give the users/airflow access.

- For AirFlow, I assume we will use the usual AirFlow yaml to give Allow access
- For AWS Console users, I think we will need to put something in this policy?

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -10,5 +10,43 @@
       "alpha_user_jamess-moj",
       "alpha_user_dami-moj"
     ]
+  },
+  "splink_s3_input_read_bucket": {
```
JamesS-MoJ, line 20: I don't think these are quite right. We will put the correct access required here: https://dsdmoj.atlassian.net/wiki/spaces/LSI/pages/6267667710/Release+1+-+Bucket+Security

JamesS-MoJ (changes requested):
See comments

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms.tf
JamesABaker: As discussed yesterday, readability > DRY for this project.

JamesABaker:
Just writing up an update so I can keep track of where we are at. 

[f43cb54](https://github.com/ministryofjustice/analytical-platform/pull/10579/commits/f43cb542ec2b41b3197d153b659dd449093be10c) and  [2806b3e](https://github.com/ministryofjustice/analytical-platform/pull/10579/commits/2806b3e795b3c1c20ed78beedfa5173b5c4c7ef5) introduce 5 explicit `.tf` files for readability, retention is now 5110 days (14 years), and `Read deny` now correct on input buckets (GetObject ,  GetObjectVersion ,  ListBucket), and a bunch of security hardening. ✅ 

What is still different from Kyle's template is:

- [x] `concat` is still in every bucket.  It's a smaller version of the same problem. We still can't read the full policy from one place. Policy statements are split across a  locals {}  block at the top of each file and the module block below. I really think here we can just have it in one place; there is no need for concat.
- [x] `DenyBucketDeletion` missing.

Final issue to address:

- [x] Permissions need aligning with the list on [confluence](https://dsdmoj.atlassian.net/wiki/spaces/LSI/pages/6267667710/Release+1+-+Bucket+Security) as highlighted by @JamesS-MoJ  here https://github.com/ministryofjustice/analytical-platform/pull/10579#discussion_r3844891946

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit.tf
```diff
@@ -0,0 +1,94 @@
...
+        Action    = "s3:PutObject"
+        Resource  = "arn:aws:s3:::${local.splink_audit_bucket_name}/*"
+        Condition = { StringNotEquals = { "s3:x-amz-server-side-encryption-aws-kms-key-id" = aws_kms_key.s3_kms_key.arn } }
+      }
+      ], [
+      {
+        Sid       = "DenyWritesForUnauthorisedPrincipals"
+        Effect    = "Deny"
+        Principal = "*"
+        Action    = ["s3:DenyBucketDeletion", "s3:PutObject", "s3:DeleteObject", "s3:DeleteObjectVersion"]
```
JamesABaker: `DenyBucketDeletion` isn't an s3 action as far as I am aware so will just be silently ignored. I was referring to the Sid used by Kyle in `analytical-platform/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3.tf`

```tf
Sid    = "DenyBucketDeletion"
Action = [
  "s3:DeleteBucket",
  "s3:PutBucketAcl",
  "s3:PutBucketPolicy",
  "s3:PutEncryptionConfiguration",
  "s3:PutBucketVersioning"
]
```

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms.tf
JamesABaker, line 262: `local.splink_s3_key_user_arns`  resolves to the main splink bucket users from  `kms_key_users.json`  (splink_s3_bucket) and currently only  `alpha_user_jamess-moj` (who will be removed after successful validation). 

The 5 new buckets each have their own entries in  `kms_key_users.json`  but there is no KMS grant statement granting  Encrypt/Decrypt/GenerateDataKey*  to those roles. My concern is that Airflow on the new buckets has no KMS data-plane access and writes will fail.

JamesABaker (changes requested):
We are getting closer. Explicit files are a real improvement and the retention, SSL/TLS, and encryption statements all look correct to me. I think three substantive issues remain before this is mergeable.

 @JamesS-MoJ correct me if I am wrong, I think we want object lock on output/audit, so now is the time to add it. Object lock cannot be added retroactively. Output and audit buckets need  `object_lock_enabled = true`  in the module block. Without it the 5110-day lifecycle rule only expires non-current versions and it would not prevent deletion of current object versions.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit.tf
arun-moj-laa: remove it.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms.tf
arun-moj-laa: updated the configuration

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit.tf
JamesABaker: @arun-moj-laa This is not resolved following [c9ef6f9](https://github.com/ministryofjustice/analytical-platform/pull/10579/commits/c9ef6f96085e7eb7c031a75e3de54ecf1d11386b).

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit.tf
```diff
@@ -0,0 +1,103 @@
...
+        sse_algorithm     = "aws:kms"
+      }
+      bucket_key_enabled = true
+    }
+  }
+  object_lock_configuration = {
+    rule = {
+      default_retention = {
+        mode = "COMPLIANCE"
+        days = 5110
```
JamesABaker: One for @JamesS-MoJ. Flagging for confirmation before this merges. `COMPLIANCE`  mode means no one, including root, can delete or shorten the retention for 5110 days. That's irreversible from the moment the bucket is created. Should this be  `GOVERNANCE`? My understanding was we wanted an escape hatch in case the retention period needs correcting. `GOVERNANCE` still prevents accidental deletion but allows an IAM-privileged principal to override if needed. 

Happy to be wrong, but worth explicitly confirming given the permanence.

JamesS-MoJ:
> We are getting closer. Explicit files are a real improvement and the retention, SSL/TLS, and encryption statements all look correct to me. I think three substantive issues remain before this is mergeable.
> 
> James Shepherd (@JamesS-MoJ) correct me if I am wrong, I think we want object lock on output/audit, so now is the time to add it. Object lock cannot be added retroactively. Output and audit buckets need  `object_lock_enabled = true`  in the module block. Without it the 5110-day lifecycle rule only expires non-current versions and it would not prevent deletion of current object versions.

I can't find any documentation that says that Object lock can't be set on a bucket after creation. As it is bucket wide, I would expect that it always applies to all objects, but I can't find anything in the docs.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit.tf
JamesS-MoJ: Are we sure we are able to change the bucket policy after creation - are we denying that anywhere? I would suggest that we go for Governance for now, assuming we can change it later, but you might want to check with Kyle.

Also, I would like to double check with the dev team what we are suggesting has object lock enabled. I would say the source, splink and audit are the obvious ones, but would like to check these and the others.

JamesABaker:
> > We are getting closer. Explicit files are a real improvement and the retention, SSL/TLS, and encryption statements all look correct to me. I think three substantive issues remain before this is mergeable.
> > James Shepherd (James Shepherd (@JamesS-MoJ)) correct me if I am wrong, I think we want object lock on output/audit, so now is the time to add it. Object lock cannot be added retroactively. Output and audit buckets need  `object_lock_enabled = true`  in the module block. Without it the 5110-day lifecycle rule only expires non-current versions and it would not prevent deletion of current object versions.
> 
> I can't find any documentation that says that Object lock can't be set on a bucket after creation. As it is bucket wide, I would expect that it always applies to all objects, but I can't find anything in the docs.

I am starting to think this is old docs. Reading more around I am getting the impression that if you set an object lock on a bucket after creation, you would have to handle how old objects in it would be locked.

Do you think it is sensible to introduce it now?

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit.tf
TawneeOwl: I believe that with compliance mode; we would run into difficulties from a data protection standpoint. Governance will give an administrator backdoor if it is discovered we need to remove this data sooner which i'm happy with. Governance will still protect against normal deletion. We will just need to ensure that administrator account is locked down.

JamesABaker: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock.html <- there is some mention of the legal type of object lock.

Please do check with dev team if we even want this from the off.

JamesABaker: Thanks @TawneeOwl . Your call @JamesS-MoJ . I can be convinced either way.

JamesABaker (changes requested):
- [x] `DenyBucketDeletion` still missing from `splink_s3_audit.tf` and `splink_s3_search_output.tf`. *
- [x] Three of us are leaning toward `GOVERNANCE`. Lets go with that.
- [x] @JamesS-MoJ to check if the concern about wrong KMS users has been resolved.

* The `s3:DenyBucketDeletion` fake action was removed from `source_input/source_output` correctly, but `splink_s3_audit.tf`  and  `splink_s3_search_output.tf ` still have no  `DenyBucketDeletion`  statement at all. Kyle's separate Sid with real actions is still absent from these two files.

JamesABaker:
I think this is the state of the PR.

- Policy structure is correct.
- The deny-unless-allowlisted pattern matches the design. 
- I think the ARN lists in `kms_key_users.json` are incomplete: 
    - Audit write missing Airflow
        - `ArnNotEquals` deny statements mean that Airflow roles need to be in those ARN lists or they'll be denied by the bucket policy itself.
    - output/audit read missing DE/DS named users (this can come later)
    - no LAA user role exists yet. 

Do we know how we will be managing "the LAA user role" (Adrian at first)?

JamesABaker:
I have fixed the airflow access and pushed, but GitHub is failing to update the pr. https://github.com/ministryofjustice/analytical-platform/commit/4680e8fc1a1edfca3a5c5bf49a05cd4b1cefc56b

JamesS-MoJ:
> > We are getting closer. Explicit files are a real improvement and the retention, SSL/TLS, and encryption statements all look correct to me. I think three substantive issues remain before this is mergeable.
> > James Shepherd (James Shepherd (@JamesS-MoJ)) correct me if I am wrong, I think we want object lock on output/audit, so now is the time to add it. Object lock cannot be added retroactively. Output and audit buckets need  `object_lock_enabled = true`  in the module block. Without it the 5110-day lifecycle rule only expires non-current versions and it would not prevent deletion of current object versions.
> 
> I can't find any documentation that says that Object lock can't be set on a bucket after creation. As it is bucket wide, I would expect that it always applies to all objects, but I can't find anything in the docs.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock-configure.html and Copilot seem to say that you can enable object lock after creation but it doesn't lock existing objects, which I think was what you were saying @JamesABaker . So we will need object lock on the output buckets and audit bucket. I would want to put object lock on the input buckets, but we can't - Airflow has to be able to move the files to a different path to avoid processing the same files again.

JamesABaker:
@JamesS-MoJ  Already done I think. `object_lock_enabled = true`  is present on all three output/audit buckets, with  `object_lock_configuration`  blocks (the 5110-day GOVERNANCE rule). Input buckets correctly have neither.

JamesABaker (dismissed):
The PR now delivers:

- Five new buckets with correct naming (laa-splink-search-input/output ,  laa-splink-source-file-input/output ,  laa-search-audit).
- Object lock (GOVERNANCE, 5110 days) on output and audit buckets. Protects results and audit trail against deletion; deliberately absent from input buckets where Airflow must move processed files.
- `DenyBucketDeletion` with real S3 actions on all five buckets, following the established  `splink_s3.tf`  pattern.
- Access control via `ArnNotEquals` deny statements scoped per bucket and operation. Read/write access segregated between LAA user roles, Airflow roles, and DE/DS roles
- Airflow roles granted audit write access. The three processing roles that write output also write to audit
- KMS grants for all new buckets with  `kms:ViaService`  scoped to S3, consistent with the existing key policy pattern

Approved. Outstanding item before production: LAA user role ARN.

Waiting for approvals from @JamesS-MoJ and the AP team once that is OK'd

JamesS-MoJ:
> James Shepherd (@JamesS-MoJ) Already done I think. `object_lock_enabled = true`  is present on all three output/audit buckets, with  `object_lock_configuration`  blocks (the 5110-day GOVERNANCE rule). Input buckets correctly have neither.

I think this:

s3:PutBucketPolicy

as a deny will mean that when we create the bucket we will not be able to change the policy afterwards? I would suggest that we don't do this until we have tested the buckets.

As James said above, I don't think the lists of permissions for each bucket is correct yet

JamesABaker:
# Bucket Permissions Analysis

@JamesS-MoJ I think I have corrected the access. Below is my review.

**A short overview:** Access control uses **Deny unless in allowlist** statements in the bucket policy. ARN lists are populated from [kms_key_users.json](https://github.com/ministryofjustice/analytical-platform/blob/f-ldb-314-splink-buckets/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json).

| Bucket | Required (spec) | Implemented | Notes | Defined in |
|---|---|---|---|---|
| **laa-splink-search-input** | Airflow: Read<br>DE/DS: Read | Read/List: `airflow-development-laa-search-index-source`, `airflow-development-laa-si-access-test`, `airflow-development-laa-s3-ops`, `alpha_user_jamess-moj`<br>Write/Delete: `alpha_user_jamess-moj` only (placeholder) | Airflow is read-only. `alpha_user_jamess-moj` holds the write slot as a testing placeholder until the LAA user role is created. | [splink_s3_search_input.tf](https://github.com/ministryofjustice/analytical-platform/blob/f-ldb-314-splink-buckets/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_search_input.tf)<br>[kms_key_users.json](https://github.com/ministryofjustice/analytical-platform/blob/f-ldb-314-splink-buckets/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json) keys: `splink_s3_input_read_bucket`, `splink_s3_input_write_bucket` |
| **laa-splink-source-file-input** | Airflow: Read<br>DE/DS: Read | Read/List: `airflow-development-laa-search-index-source`, `airflow-development-laa-si-access-test`, `airflow-development-laa-s3-ops`, `alpha_user_jamess-moj`<br>Write/Delete: `alpha_user_jamess-moj` only (placeholder) | Airflow is read-only. `alpha_user_jamess-moj` holds the write slot as a testing placeholder until the LAA user role is created. | [splink_s3_source_input.tf](https://github.com/ministryofjustice/analytical-platform/blob/f-ldb-314-splink-buckets/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_source_input.tf)<br>[kms_key_users.json](https://github.com/ministryofjustice/analytical-platform/blob/f-ldb-314-splink-buckets/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json) keys: `splink_s3_input_read_bucket`, `splink_s3_input_write_bucket` |
| **laa-splink-search-output** | Airflow: Write<br>DE/DS: Read | Write/Delete: `airflow-development-laa-search-index-source`, `airflow-development-laa-si-access-test`, `airflow-development-laa-s3-ops`<br>Read/List: `alpha_user_jamess-moj` only (placeholder) | Airflow write-only. `alpha_user_jamess-moj` holds the read slot as a testing placeholder; DE/DS named roles to be added after testing. | [splink_s3_search_output.tf](https://github.com/ministryofjustice/analytical-platform/blob/f-ldb-314-splink-buckets/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_search_output.tf)<br>[kms_key_users.json](https://github.com/ministryofjustice/analytical-platform/blob/f-ldb-314-splink-buckets/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json) keys: `splink_s3_output_write_bucket`, `splink_s3_output_read_bucket` |
| **laa-splink-source-file-output** | Airflow: Write<br>DE/DS: Read | Write/Delete: `airflow-development-laa-search-index-source`, `airflow-development-laa-si-access-test`, `airflow-development-laa-s3-ops`<br>Read/List: `alpha_user_jamess-moj` only (placeholder) | Airflow write-only. `alpha_user_jamess-moj` holds the read slot as a testing placeholder; DE/DS named roles to be added after testing. | [splink_s3_source_output.tf](https://github.com/ministryofjustice/analytical-platform/blob/f-ldb-314-splink-buckets/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_source_output.tf)<br>[kms_key_users.json](https://github.com/ministryofjustice/analytical-platform/blob/f-ldb-314-splink-buckets/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json) keys: `splink_s3_output_write_bucket`, `splink_s3_output_read_bucket` |
| **laa-search-audit** | Airflow: Write<br>DE/DS: Read | Write/Delete: `airflow-development-laa-search-index-source`, `airflow-development-laa-si-access-test`, `airflow-development-laa-s3-ops`, `alpha_user_jamess-moj`<br>Read/List: `alpha_user_jamess-moj` only (placeholder) | Airflow write-only. `alpha_user_jamess-moj` holds the read slot as a testing placeholder; DE/DS named roles to be added after testing. | [splink_s3_audit.tf](https://github.com/ministryofjustice/analytical-platform/blob/f-ldb-314-splink-buckets/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit.tf)<br>[kms_key_users.json](https://github.com/ministryofjustice/analytical-platform/blob/f-ldb-314-splink-buckets/terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json) keys: `splink_s3_audit_write_bucket`, `splink_s3_audit_read_bucket` |

---

A couple of qualifiers:

1. The  Required (spec)  column for input/output buckets should include "LAA user: Write" (input) and "LAA user: Read" (output). `alpha_user_jamess-moj`  is currently acting as placeholder for both DE/DS and LAA user roles.

2. For  laa-search-audit , the Notes say "Airflow write-only" but  `alpha_user_jamess-moj`  is also in  `audit_write_bucket`. This is intentional for DE/DS testing access, not an error.

Both these issues must be revisited after testing.

JamesS-MoJ:
I think James B has needed to step away for the rest of today. I will try to update the confluence page with the permissions. Essentially:

- there is a laa-splink-source-s3 bucket missing
- The airflow roles should be production not development
- We want a -test set of 6 buckets for testing both the IaC and for Dev testing (using the development airflow roles

JamesS-MoJ:
I have updated the bucket design: https://dsdmoj.atlassian.net/wiki/spaces/LSI/pages/6267667710/Release+1+-+Bucket+Security with what the permissions the buckets should have. This includes the new set of test buckets.

Could @dami-moj, @JamesABaker and @jh-justice-dev take a look at the above doc, just to double check, but this shouldn't stop preparation of this PR.

JamesS-MoJ:
There is a new bucket (test and prod) This is for the zip files. It's include in the table on Confluence

JamesS-MoJ:
Bucket | Airflow airflow-production-laa-search-index-s3-ops | Airflow airflow-development-laa-search-index-search | Airflow airflow-development-laa-search-index-source | Airflow airflow-production-laa-search-index-search | Airflow airflow-production-laa-search-index-source | DE/DS/DevOps Access (NAMED) | LAA User | Object Lock | Status
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
mojap-data-production-datasync-laa-ingress-production | Read | None | None | None | None | Read | None | Too Late | Exists. Open to DevOps. Not separated - contains splink model in parquet and source files
laa-splink-search-s3 | Read and write | None | None | Read | None | None | None | ? | Exists. Access WIP.
laa-splink-source-s3 | Read and write | None | None | None | Read | None | None | ? | Not created
alpha-search-legal-aid-test | None | None | None | None | None | Read | None | No | Exists.
laa-splink-search-s3-test | None | Read | None | None | None | Read | None | No | Exists.
laa-splink-source-s3-test | None | None | Read | None | None | Read | None | No | Not Created
laa-splink-search-input-test | None | Read | Read | None | None | Read and Write | None | No | Not Created
laa-splink-search-output-test | None | Write | None | None | None | Read | None | No | Not Created
laa-splink-source-file-input-test | None | None | Read | None | None | Read and Write | None | No | Not Created
laa-splink-source-file-output-test | None | None | Write | None | None | Read | None | No | Not Created
laa-splink-source-zip-test | None | None | Read and Write | None | None | Read | None | No | Not Created
laa-search-audit-test | None | Write | Write | None | None | Read | None | No | Not Created
laa-splink-search-input | None | None | None | Read | Read | None | Write | No | Not created
laa-splink-search-output | None | None | None | Write | None | None | Read | Yes | Not created
laa-splink-source-file-input | None | None | None | None | Read | None | Write | No | Not created
laa-splink-source-file-output | None | None | None | None | Write | None | Read | Yes | Not created
laa-splink-source-zip | None | None | None | None | Read and Write | None | None | No | Not Created
laa-search-audit | None | None | None | Write | Write | Read | None | Yes | Not created

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_search_input_test.tf
```diff
@@ -0,0 +1,114 @@
...
+        Resource  = "arn:aws:s3:::${local.splink_search_input_bucket_test_name}/*"
+        Condition = { StringNotEquals = { "s3:x-amz-server-side-encryption" = "aws:kms" } }
+      },
+      {
+        Sid       = "DenyWrongKMSKey"
+        Effect    = "Deny"
+        Principal = "*"
+        Action    = "s3:PutObject"
+        Resource  = "arn:aws:s3:::${local.splink_search_input_bucket_test_name}/*"
+        Condition = { StringNotEquals = { "s3:x-amz-server-side-encryption-aws-kms-key-id" = aws_kms_key.s3_kms_key.arn } }
```
JamesS-MoJ: Can the test and prod buckets use different, but similarly configured keys, please

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -13,5 +13,106 @@
       "alpha_user_jamess-moj",
       "alpha_user_dami-moj"
     ]
+  },
+  "splink_s3_source_read_bucket": {
+    "key_users": [
+      "airflow-production-laa-search-index-source",
+      "airflow-production-laa-si-access-test",
+      "airflow-production-laa-search-index-search",
```
JamesS-MoJ: I think these still don't match the table on confluence / in the PR comments

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit.tf
```diff
@@ -0,0 +1,115 @@
+
+module "s3_bucket_audit" {
```
JamesS-MoJ: The audit test bucket seems to be missing, so is the source-s3-test bucket

JamesS-MoJ:
@arun-moj-laa Updated bucket permissions spec - there was a mistake in the input buckets:

Bucket | Airflow airflow-production-laa-search-index-s3-ops | Airflow airflow-development-laa-search-index-search | Airflow airflow-development-laa-search-index-source | Airflow airflow-production-laa-search-index-search | Airflow airflow-production-laa-search-index-source | DE/DS/DevOps Access (NAMED) | LAA User | Object Lock | Status
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
mojap-data-production-datasync-laa-ingress-production | Read | None | None | None | None | Read | None | Too Late | Exists. Open to DevOps. Not separated - contains splink model in parquet and source files
laa-splink-search-s3 | Read and write | None | None | Read | None | None | None | ? | Exists. Access WIP.
laa-splink-source-s3 | Read and write | None | None | None | Read | None | None | ? | Not created
alpha-search-legal-aid-test | None | None | None | None | None | Read | None | No | Exists.
laa-splink-search-s3-test | None | Read | None | None | None | Read | None | No | Exists.
laa-splink-source-s3-test | None | None | Read | None | None | Read | None | No | Not Created
laa-splink-search-input-test | None | Read and Write | Read | None | None | Read and Write | None | No | Not Created
laa-splink-search-output-test | None | Write | None | None | None | Read | None | No | Not Created
laa-splink-source-file-input-test | None | None | Read and Write | None | None | Read and Write | None | No | Not Created
laa-splink-source-file-output-test | None | None | Write | None | None | Read | None | No | Not Created
laa-splink-source-zip-test | None | None | Read and Write | None | None | Read | None | No | Not Created
laa-search-audit-test | None | Write | Write | None | None | Read | None | No | Not Created
laa-splink-search-input | None | None | None | Read and Write | Read | None | Write | No | Not created
laa-splink-search-output | None | None | None | Write | None | None | Read | Yes | Not created
laa-splink-source-file-input | None | None | None | None | Read and Write | None | Write | No | Not created
laa-splink-source-file-output | None | None | None | None | Write | None | Read | Yes | Not created
laa-splink-source-zip | None | None | None | None | Read and Write | None | None | No | Not Created
laa-search-audit | None | None | None | Write | Write | Read | None | Yes | Not created

JamesS-MoJ (changes requested):
Couple of missing things.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit.tf
JamesS-MoJ: Also, can we toggle off the prod bucket creation for now - until after testing the test buckets, please.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
arun-moj-laa: discussed with Dami and removed this across prod buckets.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit.tf
arun-moj-laa: done

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -13,5 +13,70 @@
...
+  },
+  "splink_s3_source_read_bucket": {
+    "key_users": [
+      "airflow-production-laa-search-index-source",
+      "alpha_user_jamess-moj"
+    ]
+  },
+  "splink_s3_source_write_bucket": {
+    "key_users": [
+      "airflow-production-laa-search-index-source",
```
JamesS-MoJ: airflow-production-laa-search-index-source shouldn't have access, see the table, neither should the alpha user

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -13,5 +13,70 @@
       "alpha_user_jamess-moj",
       "alpha_user_dami-moj"
     ]
+  },
+  "splink_s3_source_read_bucket": {
+    "key_users": [
+      "airflow-production-laa-search-index-source",
+      "alpha_user_jamess-moj"
```
JamesS-MoJ: Alpha user shouldn't have access

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -13,5 +13,70 @@
...
+  "splink_s3_source_write_bucket": {
+    "key_users": [
+      "airflow-production-laa-search-index-source",
+      "airflow-production-laa-s3-ops",
+      "alpha_user_jamess-moj"
+    ]
+  },
+  "splink_s3_input_read_bucket": {
+    "key_users": [
+      "airflow-production-laa-search-index-source",
```
JamesS-MoJ: This should be the search not source role and the ops role and alpha user shouldn't have access

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -13,5 +13,70 @@
...
+      "airflow-production-laa-s3-ops",
+      "alpha_user_jamess-moj"
+    ]
+  },
+  "splink_s3_input_write_bucket": {
+    "key_users": ["alpha_user_jamess-moj"]
```
JamesS-MoJ: Need to add the search airflow role

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -13,5 +13,70 @@
...
+  },
+  "splink_s3_output_write_bucket": {
+    "key_users": [
+      "airflow-production-laa-search-index-source",
+      "airflow-production-laa-s3-ops"
```
JamesS-MoJ: Remove ops

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -13,5 +13,70 @@
...
+  },
+  "splink_s3_output_read_bucket": {
+    "key_users": ["alpha_user_jamess-moj"]
+  },
+  "splink_s3_audit_read_bucket": {
+    "key_users": ["alpha_user_jamess-moj"]
+  },
+  "splink_s3_audit_write_bucket": {
+    "key_users": [
+      "airflow-production-laa-s3-ops",
```
JamesS-MoJ: shuold not be ops - should be the source and search airflow roles

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -13,5 +13,70 @@
...
+      "alpha_user_jamess-moj"
+    ]
+  },
+  "splink_s3_source_read_bucket_test": {
```
JamesS-MoJ, line 41: All these test buckets need to follow the rules below:


Bucket | Airflow airflow-production-laa-search-index-s3-ops | Airflow airflow-test-laa-search-index-search | Airflow airflow-test-laa-search-index-source | DE/DS/DevOps Access (NAMED) | LAA User | Object Lock | Status
-- | -- | -- | -- | -- | -- | -- | --
laa-splink-search-s3-test | None | Read | None | Read and Write | None | No | Exists.
laa-splink-source-s3-test | None | None | Read | Read and Write | None | No | Not Created
laa-splink-search-input-test | None | Read and Write | Read | Read and Write | None | No | Not Created
laa-splink-search-output-test | None | Write | None | Read | None | No | Not Created
laa-splink-source-file-input-test | None | None | Read and Write | Read and Write | None | No | Not Created
laa-splink-source-file-output-test | None | None | Write | Read | None | No | Not Created
laa-splink-source-zip-test | None | None | Read and Write | Read | None | No | Not Created
laa-search-audit-test | None | Write | Write | Read | None | No | Not Created

JamesS-MoJ (changes requested):
Please check the roles follow the spec

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
arun-moj-laa: done

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
arun-moj-laa: done

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
arun-moj-laa: done

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
arun-moj-laa: done

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
arun-moj-laa: done

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
arun-moj-laa: done

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -1,9 +1,6 @@
 {
   "splink_s3_bucket": {
-    "key_users": [
-      "alpha_user_jamess-moj",
-      "airflow-production-laa-search-index-s3-ops"
-    ]
+    "key_users": ["airflow-production-laa-search-index-s3-ops"]
   },
   "splink_s3_test_bucket": {
```
JamesS-MoJ, line 5: There is still a bucket missing, I think: laa-splink-search-s3-test. Is that what this JSON is for? I can't find a corresponding S3 terraform resource. This bucket needs so exist with the perms on the confluence page.

JamesS-MoJ (changes requested):
Still a missing bucket, I think

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -13,5 +10,127 @@
...
+    "key_users": [
+      "alpha_user_jh-justice-dev",
+      "alpha_user_nina-knowles-justice",
+      "alpha_user_jfjb-justice"
+    ]
+  },
+  "splink_s3_source_output_write_bucket_test": {
+    "key_users": ["airflow-test-laa-search-index-source"]
+  },
+  "splink_s3_search_input_read_bucket_test": {
```
JamesS-MoJ, line 82: Sorry, my fault, there is a mistake here, airflow-test-laa-search-index-source should not have read access.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -13,5 +10,127 @@
...
+      "alpha_user_jh-justice-dev",
+      "alpha_user_nina-knowles-justice",
+      "alpha_user_jfjb-justice",
+      "airflow-test-laa-search-index-search"
+    ]
+  },
+  "splink_s3_search_output_write_bucket_test": {
+    "key_users": ["airflow-test-laa-search-index-search"]
+  },
+  "splink_s3_search_output_read_bucket_test": {
```
JamesS-MoJ, line 102: Sorry, my fault, there is a mistake here, airflow-test-laa-search-index-source should have read access.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
```diff
@@ -13,5 +10,127 @@
...
+    "key_users": [
+      "alpha_user_jh-justice-dev",
+      "alpha_user_nina-knowles-justice",
+      "alpha_user_jfjb-justice"
+    ]
+  },
+  "splink_s3_source_zip_read_bucket_test": {
+    "key_users": [
+      "alpha_user_jamess-moj",
```
JamesS-MoJ, line 111: This should be the list of three users that is elsewhere, not me.

JamesS-MoJ (changes requested):
Some more changes

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit_test.tf
```diff
@@ -0,0 +1,115 @@
...
+  server_side_encryption_configuration = {
+    rule = {
+      apply_server_side_encryption_by_default = {
+        kms_master_key_id = aws_kms_key.s3_kms_key.arn
+        sse_algorithm     = "aws:kms"
+      }
+      bucket_key_enabled = true
+    }
+  }
+  object_lock_configuration = {
```
JamesS-MoJ, line 96: I think this can be removed as we are not using object lock here?

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit_test.tf
```diff
@@ -0,0 +1,115 @@
...
+    target_prefix = "s3access/${local.splink_audit_bucket_test_name}/"
+  }
+  lifecycle_rule = [{
+    id                                     = "expire-noncurrent-versions"
+    enabled                                = true
+    noncurrent_version_expiration          = { days = 5110 }
+    abort_incomplete_multipart_upload_days = 7
+  }]
+  tags = merge(local.tags, { Name = local.splink_audit_bucket_test_name })
+}
```
JamesS-MoJ, line 115: Event rules are missing here.

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_search_output_test.tf
```diff
@@ -0,0 +1,160 @@
...
+  server_side_encryption_configuration = {
+    rule = {
+      apply_server_side_encryption_by_default = {
+        kms_master_key_id = aws_kms_key.s3_kms_key.arn
+        sse_algorithm     = "aws:kms"
+      }
+      bucket_key_enabled = true
+    }
+  }
+  object_lock_configuration = {
```
JamesS-MoJ, line 95: Can this be removed because we are not doing object lock?

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/splink_s3_audit_test.tf
```diff
@@ -0,0 +1,115 @@
...
+        Resource  = "arn:aws:s3:::${local.splink_audit_bucket_test_name}/*"
+        Condition = { StringNotEquals = { "s3:x-amz-server-side-encryption" = "aws:kms" } }
+      },
+      {
+        Sid       = "DenyWrongKMSKey"
+        Effect    = "Deny"
+        Principal = "*"
+        Action    = "s3:PutObject"
+        Resource  = "arn:aws:s3:::${local.splink_audit_bucket_test_name}/*"
+        Condition = { StringNotEquals = { "s3:x-amz-server-side-encryption-aws-kms-key-id" = aws_kms_key.s3_kms_key.arn } }
```
JamesS-MoJ: Shouldn't this be s3_kms_key_test the test key? Same in the other test buckets?

JamesS-MoJ (changes requested):
See comments

terraform/aws/analytical-platform-data-production/laa-search-legal-aid-data/kms_key_users.json
arun-moj-laa: done

JamesS-MoJ (approved):
Changes made as requested

laurentb4 (approved):
All changes requested by James have been made

PR #10602, opened 2026-08-27
Update bedrock Airflow iam policy to allow batch processing

guyvw1:
From @Gary-H9 - This PR provides a tactical work around which allows users and airflow to use batch inference. This violates the policy of using AI gateway for this kind of operation. However, we have been given little opportunity to work on this in the AI gateway and as a result we are adding this position.


Gary-H9 (approved):
LGTM

PR #10617, opened 2026-09-01
SP-4026 - OPG - Add PutObjectAcl for S3 Sync while setting ACL to bucket-owner-full-control #minor

sixdaysandy:


BrianEllwood (approved):
LGTM

PR #10620, opened 2026-09-01
Switch mojap-data-production-opg-forms bucket encryption to S3-managed

BrianEllwood:
This pull request removes the use of a custom AWS KMS key for server-side encryption in the OPG Forms S3 bucket, switching instead to the default S3-managed encryption (AES256). The KMS key module and its configuration are deleted, and the S3 bucket is updated to use AES256 encryption.

Key changes:

**Encryption configuration changes:**
* The `opg_forms_kms` module and all associated custom KMS key configuration have been removed from `kms-keys.tf`.
* The S3 bucket in `s3.tf` is now configured to use the default `AES256` server-side encryption instead of a custom KMS key.


Gary-H9 (approved):
LGTM

PR #10680, opened 2026-09-09
Update secrets scan workflow

muhammad9970:
Implement a for loop to loop through eu-west-1 and eu-west-2 to fetch the secrets from there.

Relates to the work being done [here](https://github.com/ministryofjustice/data-platform/issues/378)


BrianEllwood (approved):
LGTM

